PR #6, opened 2016-03-30
Automated PR for ADSL

myndzi:


src/ADSL.xml
```diff
@@ -0,0 +1,14 @@
...
+  <urls>
+    <url>https://fedoraproject.org/wiki/Licensing/AmazonDigitalServicesLicense</url>
+  </urls>
+  <license>
+    <body>
+      <p>This software code is made available "AS IS" without warranties of any kind. You may copy, display,
+         modify and redistribute the software code either by itself or as incorporated into your code; provided
+         that &gt; you do not remove any proprietary notices. Your use of this software code is at your own
+         risk and you waive any claim against Amazon Digital Services, Inc. or its affiliates with respect to
+         your use of this software code. (c) 2006 Amazon Digital Services, Inc. or its affiliates.</p>
```
sam-ellis: Is it permitted to use a <copyright> tag within <body>, so as to capture the copyright here?

myndzi: We're going to have to do that with "optional" anyway, so I don't see why we can't do it with any relevant tag. One purpose of the "copyright" tag is to eventually allow matching of any relevant copyright notice, not just the format used in the file as given.

jlovejoy:
discussed on legal call: further markup not needed due to vanity license

PR #8, opened 2016-03-30
Automated PR for AFL-1.2

myndzi:


src/AFL-1.2.xml
```diff
@@ -0,0 +1,78 @@
+<SPDX name="Academic Free License v1.2" identifier="AFL-1.2" osi-approved="true">
+  <urls>
+    <url>http://opensource.linux-mirror.org/licenses/afl-1.2.txt</url>
```
sam-ellis: URL appears broken. Try to find an alternative.

src/AFL-1.2.xml
```diff
@@ -0,0 +1,78 @@
...
+      <p>Licensed under the Academic Free License version 1.2</p>
+    </optional>
+    <body>
+      <p>Grant of License. Licensor hereby grants to any person obtaining a copy of the Original Work ("You") a
+         world-wide, royalty-free, non-exclusive, perpetual, non-sublicenseable license (1) to use, copy,
+         modify, merge, publish, perform, distribute and/or sell copies of the Original Work and derivative
+         works thereof, and (2) under patent claims owned or controlled by the Licensor that are embodied in
+         the Original Work as furnished by the Licensor, to make, use, sell and offer for sale the Original
+         Work and derivative works thereof, subject to the 
+        <br/>following conditions.
```
sam-ellis: Remove \<br/>?

src/AFL-1.2.xml
```diff
@@ -0,0 +1,78 @@
...
+         indirect, special, incidental, or consequential damages of any character arising as a result of this
+         License or the use of the Original Work including, without limitation, damages for loss of goodwill,
+         work stoppage, computer failure or malfunction, or any and all other commercial damages or losses.
+         This limitation of liability shall not apply to liability for death or personal injury resulting from
+         Licensor's negligence to the extent applicable law prohibits such limitation. Some jurisdictions do
+         not allow the exclusion or limitation of incidental or consequential damages, so this exclusion and
+         limitation may not apply to You.</p>
+      <p>License to Source Code. The term "Source Code" means the preferred form of the Original Work for making
+         modifications to it and all available 
+        <br/>documentation describing how to modify the Original Work. Licensor hereby agrees to provide a
```
sam-ellis: Remove \<br/>?

src/AFL-1.2.xml
bradleeedmondson: Yes, I believe we can & should remove be tags, replacing them with p tags
where necessary (though I wouldn't say it's necessary here).

src/AFL-1.2.xml
bradleeedmondson: I don't know that this is the right place to check for this, at least right now. @sam-ellis do you want to check with @jlovejoy ? In the interest of moving things along, I think I'll go ahead and merge this license, and assume that this URL issue will get worked out through another process.

PR #9, opened 2016-03-30
Automated PR for AFL-2.0

myndzi:


src/AFL-2.0.xml
```diff
@@ -0,0 +1,189 @@
+<SPDX name="Academic Free License v2.0" identifier="AFL-2.0" osi-approved="true">
+  <urls>
+    <url>http://opensource.linux-mirror.org/licenses/afl-2.0.txt</url>
```
sam-ellis: URL no longer works. Try to find an alternative.

PR #11, opened 2016-03-30
Automated PR for AFL-3.0

myndzi:


src/AFL-3.0.xml
```diff
@@ -0,0 +1,208 @@
+<spdx name="Academic Free License v3.0" identifier="AFL-3.0" osi-approved="true">
+  <urls>
+    <url>http://www.rosenlaw.com/AFL3.0.htm</url>
```
sam-ellis: rosenlaw link appears to no longer work. Remove or find an alternative. The OSI link below is fine.

wking: > rosenlaw link appears to no longer work

This issue still stands, and there's still a needs-edit label on this PR, so I'm not sure why it was merged.  Should I file a PR removing the broken link?

bradleeedmondson: Not 100% sure we have a policy that we will remove currently non-working URLs -- we might want to keep them for archival purposes, and just add new URLs as new authoritative(-ish) URLs come up. 

Since this affects a number of licenses, can you open an issue where we can decide how to handle linkrot (or document it, if there's already a policy), before any more PRs?

wking: > … we might want to keep them for archival purposes, and just add new URLs as new authoritative(-ish) URLs come up.

We have Git for archival ;).  If we want to keep known-dead URLs in the XML, I recommend we at least mark them up as such (e.g. `<url dead="true">…</url>`).

> Since this affects a number of licenses, can you open an issue where we can decide how to handle linkrot (or document it, if there's already a policy), before any more PRs?

I haven't been able to turn up a policy around this.  But it seems like an issue for the legal team, so I don't know if an issue in this repo makes sense.  Maybe I should mail spdx-legal?  Or file an issue in sdpx/spdx-spec?  Or maybe this repo is the right place?

bradleeedmondson:
I think an issue in this repo is the right place; we can then discuss on
the email list or, more likely, the next legal conference call.

PR #37, opened 2016-03-30
Automated PR for CDDL-1.0

myndzi:


src/CDDL-1.0.xml
```diff
@@ -0,0 +1,397 @@
...
+    </title>
+    <body>
+      <list>
+        <li>
+          <b>1.</b>
+          <p>Definitions.</p>
+        </li>
+        <li>
+          <b>1.1.</b>
+          <p>&#x201C;Contributor&#x201D; means each individual or entity that creates or contributes to the
```
pmadick: Number of areas in license where technical fix may be required.  Unusual characters appear in blue.  This license will need to be rechecked before approval.

sam-ellis: The &#x201C; and &#x201D; correspond to HTML hex encoded left and right double quote characters respectively (see http://www.fileformat.info/info/unicode/char/201C/index.htm), which is consistent with the original source file. Whether we strictly need them versus just directly using the equivalent UTF-8 characters I do not know.

sam-ellis: This (http://stackoverflow.com/questions/1091945/what-characters-do-i-need-to-escape-in-xml-documents) provides a list of XML characters that need to be escaped, and the left and right quote are not amongst them, so in that case it ought to be valid to replace these with their proper UTF-8 characters.

bradleeedmondson:
looks like these hex-encoded HTML characters need to be replaced by their UTF-8 equivalents, per Sam's comment on an [older diff](https://github.com/spdx/license-list-XML/pull/37/files/74c34dce795df79eb5033f6f4135885af8b4ae33#diff-63f22c8bb5e1181eb18a7475e79721e6)

myndzi:
This one looks to all be curly quotes, which should be safe to replace with `'` or `"`

jlovejoy:
@goneall - will these HTML characters get stripped out in the clean-up? I don't think we specifically need "curly" quotes here! (this license is otherwise ready to merge)

PR #56, opened 2016-03-30
Automated PR for CrystalStacker

myndzi:


src/CrystalStacker.xml
```diff
@@ -0,0 +1,25 @@
+<SPDX name="CrystalStacker License" identifier="CrystalStacker" osi-approved="false">
+  <urls>
+    <url>https://fedoraproject.org/wiki/Licensing:CrystalStacker?rd=Licensing/CrystalStacker</url>
+  </urls>
+  <license>
+    <title>
+      <p>Crystal Stacker is freeware. This means you can pass copies around freely provided you include this
```
sam-ellis: Lines 7-9 do not look like a title. Consider merging with optional or body.

jlovejoy: agreed that it is not a title. I'd also say that we can simply remove the optional tag altogether

pmadick: Agreed.  Suggest that this be considered part of the body of the license.

bradleeedmondson: agreed, make part of license and remove &lt;optional&gt; tag around disclaimer

jlovejoy: discussed on 8/16 legal call - remove title and optional tag

src/CrystalStacker.xml
```diff
@@ -0,0 +1,21 @@
+<SPDX name="CrystalStacker License" identifier="CrystalStacker" osi-approved="false">
+  <urls>
+    <url>https://fedoraproject.org/wiki/Licensing:CrystalStacker?rd=Licensing/CrystalStacker</url>
+  </urls>
+  <license>
+    <p>Crystal Stacker is freeware. This means you can pass copies around freely provided you include this
```
sam-ellis: Personally I would put all of the text within body tags, as none of it specifically looks like is it title or optional. Pull request submitted.

PR #62, opened 2016-03-30
Automated PR for ECL-1.0

myndzi:


src/ECL-1.0.xml
```diff
@@ -0,0 +1,44 @@
...
+  </header>
+  <license>
+    <title>
+      <p>The Educational Community License</p>
+    </title>
+      <p>This Educational Community License (the "License") applies to any original work of authorship (the
+         "Original Work") whose owner (the "Licensor") has placed the following notice immediately following
+         the copyright notice for the Original Work:</p>
+      <p>Copyright (c) 
+        <alt name="copyright" match=".+">&lt;year&gt; &lt;copyright holders&gt;</alt>
```
jlovejoy: does this need markup here?  doesn't seem like it'd change in license, but only in standard header

PR #68, opened 2016-03-30
Automated PR for EUPL-1.0

myndzi:


src/EUPL-1.0.xml
```diff
@@ -0,0 +1,249 @@
...
+          <p>Any litigation arising between Parties, other than the European Commission, and resulting from
+             the interpretation of this License, will be subject to the exclusive jurisdiction of the
+             competent court where the Licensor resides or conducts its primary business.</p>
+        </li>
+        <li>
+          <b>15.</b>
+          <p>Applicable Law</p>
+        </li>
+      </list>
+      <p>This Licence shall be governed by the law of the European Union country where the Licensor resides or has
```
sam-ellis: These paragraphs need to be made part of bullet 15 above. Pull request submitted.

src/EUPL-1.0.xml
```diff
@@ -0,0 +1,249 @@
...
+         his registered office.</p>
+      <p>This licence shall be governed by the Belgian law if:</p>
+      <p>− a litigation arises between the European Commission, as a Licensor, and any Licensee;</p>
+      <p>− the Licensor, other than the European Commission, has no residence or registered office inside a
+         European Union country.</p>
+    </body>
+    <optional>
```
sam-ellis: I'd suggest removing optional here as content is referred to elsewhere in license, suggesting it has important meaning.

bradleeedmondson: I think I would tend to make this optional, as @myndzi had suggested, but since it's not optional [here](http://git.spdx.org/?p=license-list.git;a=blob;f=EUPL-1.0.txt;h=5b1a6e6ad572c611cba0183aa2ccc4ee749837d4;hb=HEAD), I say we leave it as not optional for now and move on.

PR #70, opened 2016-03-30
Automated PR for Entessa

myndzi:


jlovejoy:
still shows an error in the XML edit preview ??  @myndzi can you have a look? (otherwise, looks good to me)

bradleeedmondson:
The XML editor error was caused by unbalanced &lt;br&gt; tags, which I've removed. (I think we are generally removing them anyway in favor of &lt;p&gt; tags.)

Structure looks OK and now has no XML errors, so I'll merge.

PR #76, opened 2016-03-30
Automated PR for Fair

myndzi:


src/Fair.xml
```diff
@@ -0,0 +1,26 @@
...
+    </title>
+    <copyright>
+      <p>&lt;Copyright Information&gt;</p>
+    </copyright>
+    <body>
+      <p>Usage of the works is permitted provided that this instrument is retained with the works, so that any
+         entity that uses the works is notified of this instrument.</p>
+      <p>DISCLAIMER: THE WORKS ARE WITHOUT WARRANTY.</p>
+    </body>
+    <optional>
```
sam-ellis: I do not believe this is part of the actual license, and I suspect this is just somebody's comment on it. See http://fairlicense.org. I suggest this text is entirely removed.

jlovejoy:
discussed on 5/12 legal call: remove last line of text as Sam suggested; add url to URL tag for fairlicense.org

PR #79, opened 2016-03-30
Automated PR for GFDL-1.1

myndzi:


jlovejoy:
reviewed, but has same question as to copyright notice as GFDL-12

PR #80, opened 2016-03-30
Automated PR for GFDL-1.2

myndzi:


jlovejoy:
reviewed and ready to merge, but have question on copyright notice, which maybe should not be in copyright tag for FSF licenses, because it is specific to license (also check Apache-2.0)

bradleeedmondson:
Agree that we can remove the copyright tag where the (C) notice applies to
the text of the license itself, rather than the software to which the
license applies.

PR #83, opened 2016-03-30
Automated PR for GPL-1.0

myndzi:


src/GPL-1.0.xml
```diff
@@ -0,0 +1,242 @@
+<spdx name="GNU General Public License v1.0 only" identifier="GPL-1.0" osi-approved="false">
+  <urls>
+    <url>http://www.gnu.org/licenses/old-licenses/gpl-1.0-standalone.html</url>
+  </urls>
+  <notes>This license was released: February 1989. This license has been deprecated. This refers to when this GPL 1.0
```
wking: Why is GPL-1.0 deprecated?  It [isn't deprecated in license-list][1] (although [GPL-1.0+ is deprecated][2]).

[1]: https://github.com/spdx/license-list/blob/788e54b566dc78358e3e1c3f9b7187ad4005114e/GPL-1.0.txt
[2]: https://github.com/spdx/license-list/blob/788e54b566dc78358e3e1c3f9b7187ad4005114e/deprecated_GPL-1.0%2B.txt

src/GPL-1.0.xml
```diff
@@ -0,0 +1,242 @@
+<spdx name="GNU General Public License v1.0 only" identifier="GPL-1.0" osi-approved="false">
```
wking: Why the “ only”?  Won't that be true of all the entries in this repository?  Before 2.0, there were explicit entries for [`GPL-1.0+`][1], etc., but now that is covered by the `+` [license expression syntax][2] (so `GPL-1.0+` is still a valid expression, but is no longer a license ID.

[1]: https://github.com/spdx/license-list/blob/788e54b566dc78358e3e1c3f9b7187ad4005114e/deprecated_GPL-1.0%2B.txt
[2]: https://spdx.org/spdx-specification-21-web-version#h.jxpfx0ykyb60

src/GPL-1.0.xml
jlovejoy: in this case "deprecated" means by the author in lieu of a more recent version.  See https://spdx.org/spdx-license-list/license-list-overview for full description of the Notes field. Now that we have "deprecated" licenses in the SPDX License List, we might want to use different language there... although the explanations should clarify?

src/GPL-1.0.xml
```diff
@@ -0,0 +1,240 @@
+<SPDX name="GNU General Public License v1.0 only" identifier="GPL-1.0" osi-approved="false">
```
wking: This [still][1] has ` only`, which I don't think we want.

[1]: https://github.com/spdx/license-list-XML/pull/83/#discussion_r102152377

src/GPL-1.0.xml
wking: > in this case "deprecated" means by the author in lieu of a more recent version

The [field docs][1] you reference just say:

> Only factual information should be included here (e.g. the license has been deprecated).

which doesn't say anything about “by the author”.  If we are tracking license-author advice, I recommend breaking it out into a new field:

    <authorSuggests>GPL-3.0</authorSuggests>

Although the rest of this ontology doesn't have much to say about who the license author is.  Maybe pull in some Dublin Core to define the [`creator`][2] whose advice we're passing along?

And if this is ongoing advice for every license, `creator` is probably too restrictive.  What if the original creator wants to hand off responsibility for providing advice to a third party?  Maybe that's not a use-case we care about covering?

[1]: https://spdx.org/spdx-license-list/license-list-overview#fields
[2]: http://dublincore.org/documents/2012/06/14/dcmi-terms/?v=terms#terms-creator

src/GPL-1.0.xml
bradleeedmondson: Opened #398 for this since it affects several license files.

PR #106, opened 2016-03-30
Automated PR for LGPL-2.1

myndzi:


jlovejoy:
copyright notice is for license, so should not be replaceable as per guidelines?

PR #127, opened 2016-03-30
Automated PR for MPL-2.0

myndzi:


src/MPL-2.0.xml
```diff
@@ -0,0 +1,436 @@
...
+    <title>
+      <p>Mozilla Public License Version 2.0</p>
+    </title>
+    <body>
+      <list>
+        <li>
+          <b>1.</b>
+          <p>Definitions</p>
+        </li>
+        <list>
```
sam-ellis: This list should be part of the previous list. Other lists elsewhere in the document need similar attention.

PR #132, opened 2016-03-30
Automated PR for MakeIndex

myndzi:


src/MakeIndex.xml
```diff
@@ -0,0 +1,43 @@
+<!-- REVIEW -->
```
sam-ellis: Remove this line.

PR #133, opened 2016-03-30
Automated PR for MirOS

myndzi:


src/MirOS.xml
```diff
@@ -0,0 +1,92 @@
+<SPDX name="MirOS Licence" identifier="MirOS" osi-approved="true">
```
sam-ellis: Suggest to change spelling of "Licence" to "License" in SPDX name for consistency with other license list names.

jlovejoy: agreed - that is how it appears on OSI site too, not sure how that 'c' got in there.

sam-ellis: Pull request submitted to change spelling.

PR #139, opened 2016-03-30
Automated PR for NCSA

myndzi:


src/NCSA.xml
```diff
@@ -0,0 +1,53 @@
...
+  </urls>
+  <license>
+    <title>
+      <p>University of Illinois/NCSA Open Source License</p>
+    </title>
+    <copyright>
+      <p>Copyright (c) 
+        <alt name="copyright" match=".+"> &lt;Year&gt; &lt;Owner Organization Name&gt;</alt>. All rights reserved. 
+      </p>
+      <p>Developed by:
```
jlovejoy: should this be part of copyright tag?  (also may want to check if LLVM uses this line - if not, may want to make this line optional)

bradleeedmondson: agree; i'd separate the developed by into a "note" tag

pmadick: Agree.

jlovejoy: move end copyright tag up, discussed on 8/16 call

PR #140, opened 2016-03-30
Automated PR for NGPL

myndzi:


src/NGPL.xml
```diff
@@ -0,0 +1,95 @@
...
+    <url>http://www.opensource.org/licenses/NGPL</url>
+  </urls>
+  <license>
+    <title>
+      <p>NETHACK GENERAL PUBLIC LICENSE</p>
+    </title>
+    <copyright>
+      <p>(Copyright 1989 M. Stephenson)</p>
+    </copyright>
+    <optional>
```
jlovejoy: I'm not sure this should be optional, wasn't before

src/NGPL.xml
```diff
@@ -0,0 +1,95 @@
...
+         For example, if you distribute copies of NetHack, you must give the recipients all the rights that you
+         have. You must make sure that they, too, receive or can get the source code. And you must tell them
+         their rights.</p>
+      <p>Also, for our own protection, we must make certain that everyone finds out that there is no warranty for
+         NetHack. If NetHack is modified by someone else and passed on, we want its recipients to know that
+         what they have is not what we distributed.</p>
+      <p>Therefore we (Mike Stephenson and other holders of NetHack copyrights) make the following terms which say
+         what you must do to be allowed to distribute or change NetHack.</p>
+      <p>COPYING POLICIES</p>
+      <p>You may copy and distribute verbatim copies of NetHack source code as you receive it, in any medium,
```
jlovejoy: see OSI site - has numbering for these paragraphs, add higher level list here?

src/NGPL.xml
```diff
@@ -0,0 +1,95 @@
...
+         provided that you keep intact the notices on all files that refer to copyrights, to this License
+         Agreement, and to the absence of any warranty; and give any other recipients of the NetHack program a
+         copy of this License Agreement along with the program.</p>
+      <p>You may modify your copy or copies of NetHack or any portion of it, and copy and distribute such
+         modifications under the terms of Paragraph 1 above (including distributing this License Agreement),
+         provided that you also do the following:</p>
+      <list>
```
jlovejoy: extra list tag, but see comment above - may want to just add higher level list above

src/NGPL.xml
bradleeedmondson: I'd support making this optional; it seems to be a note and doesn't seem to have any legal effect. Not sure if there is an official process for marking something optional in a template though.

pmadick: Not sure this should be optional.  Perhaps this one is worthy of discussion on a call.

jlovejoy: remove optional tag, discussed on 8/16 call

PR #141, opened 2016-03-30
Automated PR for NLPL

myndzi:


src/NLPL.xml
```diff
@@ -0,0 +1,35 @@
...
+  <urls>
+    <url>https://fedoraproject.org/wiki/Licensing/NLPL</url>
+  </urls>
+  <license>
+    <title>
+      <p>NO LIMIT PUBLIC LICENSE 
+        <br/>Version 0, June 2012 
+      </p>
+    </title>
+    <copyright>
```
jlovejoy: I don't think this should be labeled with copyright tag.  Not sure it should be tagged at all??

bradleeedmondson: I could go either way. It seems like it's intended either to be an author statement or a copyright statement.

We could take an empirical approach: see what kind of use this license has, and what percentages include/exclude this section.

bradleeedmondson:
discussed on 2016-08-18 SDPX con call
decided to mark the address paragraph as optional rather than copyright

PR #156, opened 2016-03-30
Automated PR for ODbL-1.0

myndzi:


pmadick:
Unusual characters in license.

bradleeedmondson:
This license has html-encoded and hex-encoded HTML characters that (I believe) should be converted to their UTF-8 equivalents. @myndzi can you help/confirm?

myndzi:
&#x2013; is an en-dash, which should be replaceable with a simple hyphen; &quot; and the rest are unnecessary but not-harmful encodings of `"` and `'` (&apos;)

(In short, they can all become valid low-ascii just fine)

PR #181, opened 2016-03-30
Automated PR for OSL-2.1

myndzi:


src/OSL-2.1.xml
```diff
@@ -0,0 +1,199 @@
...
+  <urls>
+    <url>http://opensource.org/licenses/OSL-2.1</url>
+    <url>http://web.archive.org/web/20050212003940/http://www.rosenlaw.com/osl21.htm</url>
+  </urls>
+  <notes>Same as version 2.0 of this license except with changes to section 10</notes>
+  <license>
+    <title>
+      <p>The Open Software Licensev. 2.1</p>
+    </title>
+    <optional>
```
jlovejoy: this wasn't optional before, not sure it should be now

src/OSL-2.1.xml
```diff
@@ -0,0 +1,199 @@
...
+      <p>This Open Software License (the &quot;License&quot;) applies to any original work of authorship (the
+         &quot;Original Work&quot;) whose owner (the &quot;Licensor&quot;) has placed the following notice
+         immediately following the copyright notice for the Original Work:</p>
+      <p>Licensed under the Open Software License version 2.1</p>
```
jlovejoy: this is a standard header, but we had missed it previously (or at least it's not showing up on website) - add as standard header

src/OSL-2.1.xml
```diff
@@ -0,0 +1,199 @@
...
+    </optional>
+    <body>
+      <list>
+        <li>
+          <b>1)</b>
+          <p>Grant of Copyright License. Licensor hereby grants You a world-wide, royalty-free, non-exclusive,
+             perpetual, sublicenseable license to do the following:</p>
+        </li>
+        <list>
```
jlovejoy: @myndzi - should new <list> be included within <li> tag of last item or like this?

src/OSL-2.1.xml
bradleeedmondson: not 100% sure about the substantive question on this one, but I agree we shouldn't make it optional right now. let's leave it required for now.

jlovejoy:
good to go, but needs the odd character in section 11 (section symbol) to be fixed @myndzi or @sam-ellis - can one of you fix this?

sam-ellis:
Pull request submitted to correct bad char in section 11.

PR #183, opened 2016-03-30
Automated PR for OpenSSL

myndzi:


src/OpenSSL.xml
```diff
@@ -0,0 +1,125 @@
+<!-- REVIEW -->
```
pmadick: Not sure this line is appropriate.

jlovejoy: agree, can remove

bradleeedmondson:
structure looks good

PR #184, opened 2016-03-30
Automated PR for PDDL-1.0

myndzi:


src/PDDL-1.0.xml
```diff
@@ -0,0 +1,348 @@
+<spdx name="ODC Public Domain Dedication &amp; License 1.0" identifier="PDDL-1.0" osi-approved="false">
+  <urls>
+    <url>http://opendatacommons.org/licenses/pddl/1.0/</url>
+  </urls>
+  <license>
+    <title>
+      <p>Open Data Commons - Public Domain Dedication &amp; License (PDDL)</p>
+    </title>
+    <optional>
```
jlovejoy: I'm not sure the preamble should be optional - it wasn't before... ??

jlovejoy:
I'd like to see how this will look in HTML page.  The formatting in the original license was pretty flat, might be nice to use more nesting for sub-sections for ease of reading.

src/PDDL-1.0.xml
bradleeedmondson: Agree; I think it should not be optional.

We could reconsider if we empirically find versions of this in the wild that both include and exclude the preamble.

pmadick:
Agree that the preamble should NOT be optional.

jlovejoy:
fix, to remove optional tag

PR #189, opened 2016-03-30
Automated PR for Python-2.0

myndzi:


bradleeedmondson:
not sure we're OK to have multiple title and body tags in our format; this may need further modifications.

note for posterity: this license stack is arguably both notice file and/or an intended license, depending on how it's being used. for now we've included the stack as a separate license.

PR #190, opened 2016-03-30
Automated PR for QPL-1.0

myndzi:


src/QPL-1.0.xml
```diff
@@ -0,0 +1,123 @@
...
+    <url>http://www.opensource.org/licenses/QPL-1.0</url>
+  </urls>
+  <license>
+    <title>
+      <p>THE Q PUBLIC LICENSE version 1.0</p>
+    </title>
+    <copyright>
+      <p>Copyright (C) 1999-2005 Trolltech AS, Norway.</p>
+    </copyright>
+    <optional>
```
jlovejoy: I don't think this should be optional, wasn't before

bradleeedmondson: agreed

PR #194, opened 2016-03-30
Automated PR for RPL-1.5

myndzi:


jlovejoy:
removed optional tag around preamble - wasn't there before. also removed body tag and fixed list tags

PR #203, opened 2016-03-30
Automated PR for SGI-B-1.1

myndzi:


jlovejoy:
check header - might need markup?

bradleeedmondson:
discussed on legal call today. I will add several alts for the replaceable texts in the header.

PR #215, opened 2016-03-30
Automated PR for Spencer-86

myndzi:


src/Spencer-86.xml
```diff
@@ -0,0 +1,32 @@
...
+          <b>2.</b>
+          <p>The origin of this software must not be misrepresented, either by explicit claim or by omission.</p>
+        </li>
+        <li>
+          <b>3.</b>
+          <p>Altered versions must be plainly marked as such, and must not be misrepresented as being the
+             original software.</p>
+        </li>
+      </list>
+      <p>Beware that some of this code is subtly aware of the way operator precedence is structured in regular
```
sam-ellis: This paragraph seems non-legal in nature and could be made optional.

jlovejoy: good point, it does not appear on the examples on Fedora either.  Should it be optional or simply removed?

bradleeedmondson: agree that it's non-legal.

can we run a google-fu search to see whether it's included in source in the wild? if not, we can remove; if so, we can make optional

pmadick: Agree with Brad's conclusions.

jlovejoy: to remove, discussed on 8/16 legal call

sam-ellis:
Pull request submitted.

PR #224, opened 2016-03-30
Automated PR for Unicode-TOU

myndzi:


src/Unicode-TOU.xml
```diff
@@ -0,0 +1,174 @@
+<SPDX name="Unicode Terms of Use" identifier="Unicode-TOU" osi-approved="false">
+  <urls>
+    <url>http://www.unicode.org/copyright.html</url>
+  </urls>
+  <license>
+    <title>
+      <p>Unicode Terms of Use</p>
+    </title>
+    <optional>
```
jlovejoy: I am not sure this should be optional??

jlovejoy:
(bug) lists not nesting
(legal) see comment re: first paragraph; also note that copyright in A1 is not marked up,  I think this is correct as it never was and is a specific Unicode license

src/Unicode-TOU.xml
bradleeedmondson: agree this should not be optional. seems intended to have the legal effect of incorporating the privacy policy and tm policy. Whether it's successful, I'm not sure, but I would say don't mark as optional unless we empirically find we miss matches for licenses in the wild based on this.

pmadick:
Agree with legal team.  Fix bug, take out optional tag and commit.

PR #228, opened 2016-03-30
Automated PR for Vim

myndzi:


src/Vim.xml
```diff
@@ -0,0 +1,120 @@
+<SPDX name="Vim License" identifier="Vim" osi-approved="false">
```
sam-ellis: Original license text contains texted bullets, where as the bullets here have been flattened to the top-level. Does the XML format permit nesting bullets, and if so, should we use that here?

myndzi: Seems to be indent misread based on the text file spacing, all we have to do is nest the actual <list> tags. Will do that and update.

PR #229, opened 2016-03-30
Automated PR for W3C

myndzi:


src/W3C.xml
```diff
@@ -0,0 +1,58 @@
...
+  </header>
+  <license>
+    <title>
+      <p>W3C SOFTWARE NOTICE AND LICENSE</p>
+    </title>
+    <optional>
+      <p>This work (and included software, documentation such as READMEs, or other related items) is being
+         provided by the copyright holders under the following license.</p>
+      <p>License</p>
+      <p>By obtaining, using and/or copying this work, you (the licensee) agree that you have read, understood,
```
pmadick: The body of this license appears to start at line 22 with the sentence "By obtaining..."

pmadick: Once the issue with line 22 is resolved this is approved.

jlovejoy: agreed. I think the end optional and body tag simply need to be moved up to line 22 - Kris, can you confirm?

bradleeedmondson: I mostly agree -- I think the license actually starts at line 19.

I think the confusion comes from the multiple meanings of the word "license," which can mean the set of rights granted to a licensee within a contract, or colloquially can refer to the whole contract itself. I think in this case we have the former. If you look at the [original page](https://www.w3.org/Consortium/Legal/2002/copyright-software-20021231.html), it looks like there is a license-granting section of the agreement with header "License" and a warranty-disclaiming section of the license with header "Disclaimers."

This is somewhat undercut by the "Notes" section at the end which seems like it is _not_ intended to be part of the agreement, but I think that's a separate question and that words to the effect of "this work is provided by the copyright holders under the following license" is a part of the agreement.

bradleeedmondson:
suggested minor change in new PR #331

PR #230, opened 2016-03-30
Automated PR for W3C-19980720

myndzi:


src/W3C-19980720.xml
```diff
@@ -0,0 +1,65 @@
...
+         modifications, that you make:</p>
+      <list></list>
+      <list>
+        <li>
+          <b>1.</b>
+          <p>The full text of this NOTICE in a location viewable to users of the redistributed or derivative work.</p>
+        </li>
+        <li>
+          <b>2.</b>
+          <p>Any pre-existing intellectual property disclaimers, notices, or terms and conditions. If none
```
pmadick: The body of this license starts at least at line 16 and probably at line 15.  Please check with J.

jlovejoy: agreed.  I think we can simply remove the optional tag and move the body tag up (check that is correct change as to body tag)

src/W3C-19980720.xml
```diff
@@ -0,0 +1,65 @@
...
+         FOR ANY PARTICULAR PURPOSE OR THAT THE USE OF THE SOFTWARE OR DOCUMENTATION WILL NOT INFRINGE ANY
+         THIRD PARTY PATENTS, COPYRIGHTS, TRADEMARKS OR OTHER RIGHTS.</p>
+      <p>COPYRIGHT HOLDERS WILL NOT BE LIABLE FOR ANY DIRECT, INDIRECT, SPECIAL OR CONSEQUENTIAL DAMAGES ARISING
+         OUT OF ANY USE OF THE SOFTWARE OR DOCUMENTATION.</p>
+      <p>The name and trademarks of copyright holders may NOT be used in advertising or publicity pertaining to
+         the software without specific, written prior permission. Title to copyright in this software and any
+         associated documentation will at all times remain with copyright holders.</p>
+    </body>
+    <optional>
+      <p>____________________________________</p>
```
jlovejoy: we did not have this marked as optional before - should we do so now or leave it as part of the license?

src/W3C-19980720.xml
bradleeedmondson: agree w/ @jlovejoy

src/W3C-19980720.xml
bradleeedmondson: Coming at this fresh, I would say this following should be optional. But I would also say that it's fine to leave it as-is for now and review whether to apply the &lt;optional&gt; tag more aggressively the in future.

jlovejoy: yes, ok as optional, discussed on Aug 18 legal call

PR #239, opened 2016-03-30
Automated PR for YPL-1.0

myndzi:


src/YPL-1.0.xml
```diff
@@ -0,0 +1,193 @@
...
+         agree as follows:</p>
+    </optional>
+    <body>
+      <list>
+        <li>
+          <b>1.</b>
+          <p>Grant of Copyright License</p>
+        </li>
+      </list>
+      <list>
```
jlovejoy: not nested, should be

jlovejoy:
legal review: I don't think the first parapraph should be optional. can I get another lawyer to agree?
for @myndzi - list tags seem to be really messed up

bradleeedmondson:
agree first two paragraphs should not be optional. they include definitions, words of agreement, and a halfhearted recitation of adequate consideration.

The heavy reference to Yahoo! seems like it might be the reason why this had been marked optional in the first place, but I'm guessing this was not and need not be templatized b/c it's a vanity license just for Yahoo! If we do find others using it I suppose we could eventually mark it up and mark out the company name and address as replaceable.

pmadick:
Agreed.  Remove optional reference and commit.

PR #240, opened 2016-03-30
Automated PR for YPL-1.1

myndzi:


bradleeedmondson:
What's the question for legal discussion here?

jlovejoy:
I don't think we need the optional tag at the top

bradleeedmondson:
agree, since this is a vanity license there's no reason to mark that paragraph as optional

pmadick:
Agreed.  Remove optional mark and commit.

PR #248, opened 2016-03-30
Automated PR for Zlib

myndzi:


src/Zlib.xml
```diff
@@ -0,0 +1,35 @@
...
+         be held liable for any damages arising from the use of this software.</p>
+      <p>Permission is granted to anyone to use this software for any purpose, including commercial applications,
+         and to alter it and redistribute it freely, subject to the following restrictions:</p>
+      <list></list>
+      <list>
+        <li>
+          <b>1.</b>
+          <p>The origin of this software must not be misrepresented; you must not claim that you wrote the
+             original software. If you use this software in a product, an acknowledgment in the product
+             documentation would be appreciated but is not required.</p>
```
pmadick: Do we want to mark for acknowledgment when acknowledgment is optional?

jlovejoy: I don't think so, plus there is no specific acknowledgment text provided in this license.

bradleeedmondson: Agreed; I think we should think of the has-acknowledgement label as "requires acknowledgement," so we wouldn't attach it where acknowledgement is not required.

I would actually propose changing the label to "requires acknowledgment" for clarity; if people agree I think we would have to select all the PRs with the old label, add the new label, then remove the old.

PR #257, opened 2016-03-30
Automated PR for iMatix

myndzi:


src/iMatix.xml
```diff
@@ -0,0 +1,68 @@
...
+         documentation, and executable files, hereinafter referred to as "the Product".</p>
+      <p>The Product is Copyright (c) 1991-2000 iMatix Corporation. You may use it and distribute it according to
+         this following License Agreement. If you do not agree with these terms, please remove the Product from
+         your system. By incorporating the Product in your work or distributing the Product to others you
+         implicitly agree to these license terms.</p>
+      <p>Statement Of Copyright</p>
+      <p>The Product is, and remains, Copyright © 1991-2000 iMatix Corporation, with exception of specific
+         copyrights as noted in the individual source files.</p>
+      <p>Conditions Of Use</p>
+      <p>You do not need to provide the source code for the Product as part of your product. However, you must do
```
bradleeedmondson: should we mark this PR "requires acknowledgement" where it doesn't _always_ require acknowledgement, but does require one of the following?
- source code for modules used, OR
- open-source derivative works, OR
- acknowledgement

src/iMatix.xml
```diff
@@ -0,0 +1,68 @@
...
+         so long as you respect the License Agreement. The License Agreement does not affect any software
+         except the Product. In particular, any application that uses the Product does not itself fall under
+         the License Agreement.</p>
+      <p>You may modify any part of the Product, including sources and documentation, except this License
+         Agreement, which you may not modify.</p>
+      <p>You must clearly indicate any modifications at the start of each source file. The user of any modified
+         Product code must know that the source file is not original.</p>
+      <p>At your discretion, you may rewrite or reuse any part of the Product so that your derived code is not
+         obviously part of the Product. This derived code does not fall under the Product License Agreement
+         directly, but you must include a credit at the start of each source file indicating the original
```
bradleeedmondson: There is also a conditional acknowledgement requirement here (or would we regard this as an absolute acknowledgement requirement?)

bradleeedmondson:
approved for structure; only question is whether to apply "requires acknowledgement" tag

src/iMatix.xml
bradleeedmondson: discussed on legal call; will add acknowledgement label to this PR so that we can make these conditional acknowledgement requirements extractable later

src/iMatix.xml
bradleeedmondson: discussed on legal call; will add acknowledgement label to this PR so that we can make these conditional acknowledgement requirements extractable later

PR #262, opened 2016-03-30
Automated PR for xinetd

myndzi:


src/xinetd.xml
```diff
@@ -0,0 +1,58 @@
...
+          <b>2.</b>
+          <p>The author of the modifications will include his/her name by appending it along with the new
+             version number to this file and will be responsible for any wrong behavior of the modified
+             software.</p>
+        </li>
+      </list>
+      <p>The author makes no representations about the suitability of this software for any purpose. It is
+         provided "as is" without any express or implied warranty.</p>
+    </body>
+    <optional>
```
bradleeedmondson: this optional block looks like a note rather than an optional piece of text. but it's included at http://spdx.org/licenses/xinetd.html and at https://fedoraproject.org/wiki/Licensing/Xinetd_License 

should we remove, or at least move to notes?

it seems like matching should still work without including this, even if users included this extraneous text in their source code, no?

src/xinetd.xml
```diff
@@ -0,0 +1,58 @@
...
```
bradleeedmondson: I think this should be deleted (lines 46-56). It appears to be extraneous text copied & pasted from the license as found at https://fedoraproject.org/wiki/Licensing/Xinetd_License

(I had previously noted that it wasn't optional in SPDX git, but on review I think its inclusion at all looks like a mistake.)

bradleeedmondson:
Need legal discussion to address whether to completely remove lines 46-56.

src/xinetd.xml
bradleeedmondson: discussed on legal call today. leave text in and mark not optional, since it wasn't before. 

also note that the license point 2 above requires adherence to version-bumping guidelines which the text explains.

bradleeedmondson:
discussed on legal call today. leave text in and mark not optional, since it wasn't before.

also note that the license point 2 above requires adherence to version-bumping guidelines which the text explains.

PR #265, opened 2016-03-30
Automated PR for Adobe-Glyph

myndzi:


src/Adobe-Glyph.xml
```diff
@@ -0,0 +1,37 @@
...
+        </li>
+      </list>
+      <p>Permission is hereby granted, free of charge, to any person obtaining a copy of this documentation file,
+         to create their own derivative works from the content of this document to use, copy, publish,
+         distribute, sublicense, and/or sell the derivative works, and to permit others to do the same,
+         provided that the derived work is not represented as being a copy or version of this document.</p>
+      <p>Adobe shall not be liable to any party for any loss of revenue or profit or for indirect, incidental,
+         special, consequential, or other similar damages, whether based on tort (including without limitation
+         negligence or strict liability), contract or other legal or equitable grounds even if Adobe has been
+         advised or had reason to know of the possibility of such damages.Ê The Adobe materials are provided on
```
bradleeedmondson: Extra character "Ê" appears here and on the next line as well. This seems to come from the source/template file

Also appears in source: http://spdx.org/licenses/Adobe-Glyph.html and https://fedoraproject.org/wiki/Licensing:MIT?rd=Licensing/MIT#AdobeGlyph

Maybe this is a known issue and we just leave it?

myndzi: This is almost certainly a UTF-8 encoding artifact, but I'm not sure what character it was intended to be. Normally another character would follow it.

If we remove it, this license will probably not match other people who have copied the same text, but if we leave it it won't match people who fixed it or got the "correct" version from somewhere. It might deserve consideration for a new matching guideline dealing with ignoring high ascii characters, but that has potential problems with languages that need them. Definitely worth discussion?

myndzi:
The Adobe Glyph List seems to have been released under BSD anyway, so this license might be somewhat defunct

myndzi:
...there are also licenses in the wild that have been 'cleaned up', e.g. https://svn.apache.org/repos/asf/pdfbox/trunk/preflight-app/src/main/appended-resources/META-INF/LICENSE

jlovejoy:
discussed on 5/12 legal call:  add optional tag around the offending character

wking:
> discussed on 5/12 legal call:  add optional tag around the offending character

I don't see details for that discussion in [the minutes][1].  But if folks are interested in justifying `<optional>Ê</optional>`, they're [not in the agl-specification repo][2] which is presumably canonical because [Adobe links to it][3] and [an Adobe employee is the only contributor][4].

However, I think `<alt>` would be a better approach to keep the odd characters out of the canonical form.  I can work up a PR for this if there are no objections.

[1]: https://wiki.spdx.org/view/Legal_Team/Minutes/2016-05-12
[2]: https://github.com/adobe-type-tools/agl-specification/blame/b3e00bd2b2e65c70be034043a5fe7942a3f2b126/README.md#L13
[3]: https://www.adobe.com/devnet/font.html
[4]: https://github.com/adobe-type-tools/agl-specification/graphs/contributors

bradleeedmondson:
I agree that that's a better approach to keep the canonical cleaner. If you do a PR and @goneall reviews, I'd be in favor of merging.

PR #268, opened 2016-03-30
Automated PR for Apache-1.0

myndzi:


myndzi:
I assume the question was about your PR since there's no comment, so I removed it after I merged your changes. Good catch!

PR #274, opened 2016-03-30
Automated PR for Artistic-2.0

myndzi:


src/Artistic-2.0.xml
```diff
@@ -0,0 +1,196 @@
...
+         distributed, and/or redistributed. The intent is that the Copyright Holder maintains some artistic
+         control over the development of that Package while still keeping the Package available as open source
+         and free software.</p>
+      <p>You are always permitted to make arrangements wholly outside of this license directly with the Copyright
+         Holder of a given Package. If the terms of this license do not permit the full use that you propose to
+         make of the Package, you should contact the Copyright Holder and seek a different licensing
+         arrangement.</p>
+    </optional>
+    <body>
+      <p>Definitions</p>
```
bradleeedmondson: any reason we do not wrap definitions sections with <list>?

myndzi: I think that would be perfectly valid. The automated process based the list tags on the presence of a bullet, which is why it was not automatically converted here.

PR #276, opened 2016-03-30
Automated PR for BSD-2-Clause-FreeBSD

myndzi:


jlovejoy:
would we want to make the last bit optional?  I think we decided this made it a different license than BSD-2-Clause, but... might want to have another look?

jlovejoy:
discussed on Aug 4 legal call: leave as is, but this may be future feature request to be able to collapse, e.g., NetBSD, FreeBSD, BSD-2-Clause

PR #277, opened 2016-03-30
Automated PR for BSD-2-Clause-NetBSD

myndzi:


jlovejoy:
would we want to make the first line optional? I think we decided this made it a different license b/c of this, but... might want to have another look?

bradleeedmondson:
Sorry, I don't follow. Which line?

src/BSD-2-Clause-NetBSD.xml
```diff
@@ -0,0 +1,42 @@
...
+  <urls>
+    <url>http://www.netbsd.org/about/redistribution.html#default</url>
+  </urls>
+  <notes>NetBSD adopted this 2-clause license in 2008 for code contributed to NetBSD Foundation.</notes>
+  <license>
+    <copyright>
+      <p>Copyright (c) 2008 The NetBSD Foundation, Inc. All rights reserved.</p>
+    </copyright>
+    <body>
+      <p>This code is derived from software contributed to The NetBSD Foundation by
```
jlovejoy: should this line be optional?  I think we decided this made it a different license b/c of this, but... might want to have another look?

jlovejoy:
discussed on Aug 4 legal call: leave as is, but this may be future feature request to be able to collapse, e.g., NetBSD, FreeBSD, BSD-2-Clause

PR #284, opened 2016-03-30
Automated PR for BSD-Protection

myndzi:


myndzi:
There's something wrong here, do not merge -- investigate later

PR #286, opened 2016-03-30
Automated PR for Bahyph

myndzi:


src/Bahyph.xml
```diff
@@ -0,0 +1,29 @@
+<SPDX name="Bahyph License" identifier="Bahyph" osi-approved="false">
+  <urls>
+    <url>https://fedoraproject.org/wiki/Licensing/Bahyph</url>
+  </urls>
+  <license>
+    <optional>
+      <p>COPYRIGHT NOTICE</p>
+    </optional>
+    <body>
+      <p>These patterns and the generating sh script are Copyright (c) GMV 1991</p>
```
jlovejoy: I think this should be noted with the <copyright> tag??

src/Bahyph.xml
```diff
@@ -0,0 +1,29 @@
+<SPDX name="Bahyph License" identifier="Bahyph" osi-approved="false">
+  <urls>
+    <url>https://fedoraproject.org/wiki/Licensing/Bahyph</url>
+  </urls>
+  <license>
+    <optional>
+      <p>COPYRIGHT NOTICE</p>
```
jlovejoy: should this be optional or part of copyright tag (see other comment)??

myndzi: The copyright tag is intended to surround an actual copyright notice, so this falls more under the matching rules about omittable text

src/Bahyph.xml
jlovejoy: discussed on call; decided not to add copyright tag and never had matching markup, so not adding it now as more of a one-off license

PR #288, opened 2016-03-30
Automated PR for Beerware

myndzi:


src/Beerware.xml
```diff
@@ -0,0 +1,12 @@
+<SPDX name="Beerware License" identifier="Beerware" osi-approved="false">
+  <urls>
+    <url>https://fedoraproject.org/wiki/Licensing/Beerware</url>
+  </urls>
+  <license>
+    <body>
+      <p>"THE BEER-WARE LICENSE" (Revision 42): &lt;phk@FreeBSD.ORG&gt; wrote this file. As long as you retain
```
jlovejoy: Shouldn't "THE BEER-WARE LICENSE"  (Revision 42):" be marked as <title>?

src/Beerware.xml
```diff
@@ -0,0 +1,12 @@
+<SPDX name="Beerware License" identifier="Beerware" osi-approved="false">
+  <urls>
+    <url>https://fedoraproject.org/wiki/Licensing/Beerware</url>
```
jlovejoy: I'd like to add another url here... (but that would be a fork of the current list...)
https://people.freebsd.org/~phk/

src/Beerware.xml
myndzi: The </p> is on line 9 -- the text can span multiple lines, which is how we keep it from scrolling way off to the side :)

PR #290, opened 2016-03-30
Automated PR for BitTorrent-1.1

myndzi:


src/BitTorrent-1.1.xml
```diff
@@ -0,0 +1,509 @@
...
+      <p>New versions of this License may be published from time to time in connection with new versions of a
+         Licensed Product or otherwise. You may choose to continue to use the license terms in this version of
+         the License for the Licensed Product that was originally licensed hereunder, however, the new versions
+         of this License will at all times apply to new versions of the Licensed Product released by Licensor
+         after the release of the new version of this License. Only the Licensor has the right to change the
+         License terms as they apply to the Licensed Product.</p>
+      <p>This License relies on precise definitions for certain terms. Those terms are defined when they are first
+         used, and the definitions are repeated for your convenience in a Glossary at the end of the
+         License.</p>
+    </optional>
```
jlovejoy: should we have everything above this as optional?  (we did not have this marked as so before)

src/BitTorrent-1.1.xml
```diff
@@ -0,0 +1,509 @@
...
+         Code of the Licensed Product. (See Section 1(a))</p>
+      <p>You: This term is defined in Section 14 of this License.</p>
+    </body>
+    <optional>
+      <p>EXHIBIT A</p>
+      <p>The Notice below must appear in each file of the Source Code of any copy you distribute of the Licensed
+         Product or any hereto. Contributors to any Modifications may add their own copyright notices to
+         identify their own contributions.</p>
+      <p>License: 
+        <br/>The contents of this file are subject to the BitTorrent Open Source License Version 1.0 (the
```
jlovejoy: this is a standard header, but I don't see the header tag included anywhere - need to add, I think

jlovejoy:
I have a couple questions inline.  
Also, this is a license that has self-referring numbering - should we NOT use the bullet tag in this case?

myndzi:
I think it's pretty clear from our discussions at Collab and elsewhere that this is the direction we want to go, so the only question is whether we want to take on the extra load of looking for and modifying licenses as appropriate now or later. Since you did spot this one, let me know if you'd like me to go ahead and clean it out.

src/BitTorrent-1.1.xml
bradleeedmondson: no, definitely not. lines 12-96 should not be optional

pmadick: Agreed.  Take out optional and submit.

PR #292, opened 2016-03-30
Automated PR for CATOSL-1.1

myndzi:


src/CATOSL-1.1.xml
```diff
@@ -0,0 +1,420 @@
...
+         growth, and innovation. To demonstrate our continuing commitment to open source, we are releasing the
+         Program (as defined below) under this License.</p>
+      <p>This License is intended to permit contributors and recipients of the Program to use the Program,
+         including its source code, freely and without many of the concerns of some other open source licenses.
+         Although we expect the underlying Program, and Contributions (as defined below) made to such Program,
+         to remain open, this License is designed to permit you to maintain your own software programs free of
+         this License unless you choose to do so. Thus, only your Contributions to the Program must be
+         distributed under the terms of this License.</p>
+      <p>The provisions that follow set forth the terms and conditions under which you may use the Program.</p>
+    </optional>
```
jlovejoy: we didn't have this as optional before - not sure it should be now ??

src/CATOSL-1.1.xml
```diff
@@ -0,0 +1,420 @@
...
+        <list>
+          <li>
+            <b>*</b>
+            <p>it complies with the terms and conditions of this License; and</p>
+          </li>
+          <li>
+            <b>*</b>
+            <p>its license agreement:</p>
+          </li>
+          <list>
```
jlovejoy: extra list tag here?

src/CATOSL-1.1.xml
```diff
@@ -0,0 +1,420 @@
...
+              <b>*</b>
+              <p>states that any provisions which are inconsistent with this License are offered by that
+                 Contributor alone and not by any other party; and</p>
+            </li>
+            <li>
+              <b>*</b>
+              <p>states that source code for the Program is available from such Contributor at the cost of
+                 distribution, and informs licensees how to obtain it in a reasonable manner.</p>
+            </li>
+          </list>
```
jlovejoy: extra end list tag

jlovejoy:
might be nice to add indented sub-lists for subsections (e.g., 1.2, 1.3, etc.) - this was not done in the original, but would look nicer...

src/CATOSL-1.1.xml
bradleeedmondson: agree; this (lines 11-28) should not be optional

jlovejoy:
I updated list nesting... @myndzi - can you check it?

myndzi:
Was a missing </li>, fixed it and something else, no warnings for xml structure now

src/CATOSL-1.1.xml
pmadick: agreed.  Fix and submit.

PR #294, opened 2016-03-30
Automated PR for CC-BY-2.0

myndzi:


jlovejoy:
note: has self-referring numbering

jlovejoy:
This is ready to go, just noticed that the paragraph at the end of section 3 which is not part of the list item, doens't show in the xml edit tool - @myndzi can you check that this is otherwise okay (and just a peculiarity of the tool)?

PR #302, opened 2016-03-30
Automated PR for CC-BY-NC-4.0

myndzi:


jlovejoy:
I think I fixed this, but would be helpful if someone else did a sanity check against text/numbering of actual license @bradleeedmondson or @THV - can you have a look?

THV:
I will check in the morning. Well, later in the morning.

bradleeedmondson:
checked this today and made a few small modifications to make it better reflect the formatting and grouping from https://creativecommons.org/licenses/by-nc/4.0/legalcode

bradleeedmondson:
marked needs legal review because of my change to mark the final paragraph &lt;optional&gt;

bradleeedmondson:
discussed on SPDX legal call 2016-08-18 and decided not to mark final paragraph optional b/c:
* in practice no one removes this when using this license in the wild
* we haven't marked this paragraph (or its analogues) optional in other CC licenses, and if we are going to change that we should do a more comprehensive review of all CC licenses

PR #315, opened 2016-05-13
Pr/76 sellis

sam-ellis:
Added new URL and removed optional section as discussed in SPDX legal
call on 2016-05-13.


sam-ellis:
Incorrectly submitted as a patch against spdx master. Will re-issue as a patch against existing patch.

PR #334, opened 2017-02-01
Deprecated gpl with autoconf

goneall:
Adding the deprecated license GPL-2.0-with-autoconf-exception


goneall:
Resubmitting PR without the unnecessary commits

PR #335, opened 2017-02-01
Add gpl2 with bison exception

goneall:
Adding the deprecated license GPL 2 with Bison Exception


goneall:
Resubmitting PR without the unnecessary commits

PR #336, opened 2017-02-01
Add deprecated gpl2 with classpath

goneall:
Adding the deprecated license GPL 2 with Bison Exception


goneall:
Removing PR - found an error in the title

PR #337, opened 2017-02-01
Add deprecated gpl2 with font

goneall:
Add the deprecated license GPL2.0-with-font-exception


goneall:
Resubmitting PR without the unnecessary extra commits

PR #338, opened 2017-02-01
Add deprecated gpl2 with classpath

goneall:
Adding deprecated license GPL 2.0 with classpath exception


goneall:
Noticed some extra commits - going to resubmit to clean this up a bit

PR #349, opened 2017-02-05
Adding exception: WxWindows-exception-3.1

myndzi:


jlovejoy:
is title tag in right place?

bradleeedmondson:
No, I agree that it's not. Title shouldn't be "EXCEPTION NOTICE." Modified to move that text down into body, and added title "WxWindows Library Exception 3.1." Not sure if this will cause matching problems? Or if we should remove the title tag from this exception altogether (since there appears to be no inline title)? @goneall

jlovejoy:
I removed the title as it wasn't part of the exception on the list and in the original.

PR #356, opened 2017-02-05
Adding exception: LZMA-exception

myndzi:


jlovejoy:
does title look ok  / correct?

bradleeedmondson:
The [sourceforge URL referenced](http://nsis.sourceforge.net/Docs/AppendixI.html#lzma-special-exception) incorporates this exception into a larger agreement under the heading currently listed in the title field. The sections in that agreement are:
- I.1 Copyright
- I.2 Applicable licenses
- I.3 zlib/libpng license
- I.4 bzip2 license
- I.5 Common Public License version 1.0
- I.6 Special exception for LZMA compression module

Note that I.6 is the same text as the current contents of the title tag. I agree with Jilayne's critique of this, in that I think we shouldn't name our exception according to where it falls in a particular license-manifest document. I think we should change the title to something more generic like "LZMA exception." In fact I am not 100% clear on why we have both a "name" attribute and a "title" tag -- seems like they should pretty much always be identical? (Or if not, please let me know what I'm missing.)

Anyway, for this exception, I'll change the title, review the XML structure, and merge.

PR #361, opened 2017-02-05
Adding exception: GCC-exception-2.0

myndzi:


jlovejoy:
if no urls should empty tag be removed?
formatting in Notes field would be nice here... also updated this note similar to Autoconf-exception-2.0

PR #369, opened 2017-02-05
Adding exception: Classpath-exception-2.0

myndzi:


jlovejoy:
see added wording on Note tag about use of exception and not using deprecated license - ok?

PR #370, opened 2017-02-05
Adding exception: Bison-exception-2.2

myndzi:


jlovejoy:
@goneall we have a special tag for exceptions - I added it to Google doc, but noting here as an example. Also, can we have formatting in Notes tag/field?

goneall:
@jlovejoy Thanks for pointing out the exception tag.  I missed that when we did the original XML tags and attributes list in google docs.  Would it be OK to change the name from "exception" to "licenseException" in the final version?  This would make it more consistent with the RDF class names.  We could continue using "exception" and convert it with the other terms that have changed.  BTW - I'm OK with "exception" if anyone feels strongly we should keep it shorter.

goneall:
@jlovejoy Currently, the HTML generator for the spdx.org/licenses page treats notes as a plain text field and doesn't support formatting.

bradleeedmondson:
I think it makes sense to merge all exceptions with "exception" now and convert to "licenseException" when we do the other tag conversions.

PR #371, opened 2017-02-05
Adding exception: Autoconf-exception-3.0

myndzi:


jlovejoy:
same comment re: notes as others

PR #378, opened 2017-02-05
Adding license: OSET-PL-2.1

myndzi:


jlovejoy:
this license is really messed up. it's missing a bunch of text.

bradleeedmondson:
Noting here the issues reported by Steve Snow/Black Duck back in Nov:

--------------------

Team,

I noted some errors in your "OSET Public License version 2.1" while prepping it for inclusion in our KnowledgeBase here at Black Duck.

See: Paragraphs 6, 7 & 10 where the title of the paragraph blends abruptly into the text of the graph with no spaces.

Also note that other sub-paragraphs are not as clear as they should be (see at least "2.1 Grants" as an example.)  In the originals the sub-paragraph title is separated by at least a <br> character on the OSI page while it is a full </p><p> separator on OSET's version.  The breaks in the SPDX version are only one space visually though in the source it is five spaces.

Thanks!

Steve Snow

zvr:
Wow, that was a mess!
Missing lots of text, XML tags completely wrong, ...
I think it's all fixed now, but someone should review it before merging.
Open issues: whether the text in the beginning (before "license text starts here") or the Exhibits at the end should be marked **optional**.

jlovejoy:
thanks @zvr !  that one was killing me!  I'll have a look after the weekend.

bradleeedmondson:
I haven't done a line-by-line against the original, but the structure looks good and a cursory comparison checks out. merging.

PR #380, opened 2017-02-05
Adding license: Net-SNMP

myndzi:


jlovejoy:
@goneall - I added the new listVersionAdded attribute for version of license list, optional attribute. Can you check that I did this correctly before I do any more?

goneall:
@jlovejoy Reviewed and it looks good to me

PR #381, opened 2017-02-05
Adding license: LiLiQ-Rplus-1.1

myndzi:


src/LiLiQ-Rplus-1.1.xml
```diff
@@ -0,0 +1,258 @@
...
+          <p>Le concédant peut aussi choisir de concéder la licence
+             sous la version actuelle ou toute version ultérieure,
+             auquel cas le licencié peut choisir sous quelle
+             version la licence lui est accordée.</p>
+        </li>
+        <li>
+          <b>11.</b>
+          <p>Divers</p>
+        </li>
+      </list>
```
steparce: Does list end after first paragraph of item 11 or second paragraph?

jlovejoy: hard to say from looking at the original license. can you understand enough to make a call?  if it's not clear, it's probably fine to put it after the second paragraph

src/LiLiQ-Rplus-1.1.xml
```diff
@@ -0,0 +1,258 @@
...
+                 droits d'auteur, des marques de commerce, des
+                 garanties ou de la paternité concernant le
+                 logiciel ne doivent pas être modifiées ou
+                 supprimées, à moins que ces étiquettes ou
+                 mentions ne soient inapplicables à un logiciel
+                 modifié ou dérivé donné.
+          </p>
+        </li>
+        <li>
+          <b>4.1.</b>
```
jlovejoy: I'd probably put a nested list for 4.1 and 4.2

src/LiLiQ-Rplus-1.1.xml
bradleeedmondson: There's a full (if unofficial) English translation at the OSI URL: https://opensource.org/licenses/LiLiQ-Rplus-1.1

I think the list ends after the second paragraph in 11. The license starts with a list item and looks to be one of those where every paragraph is within the numbering scheme.

FWIW, note that Kris' original parsing code very frequently made this error, ending the list entity before the start of the last &lt;li&gt; list item instead of after the &lt;/li&gt; was closed.

src/LiLiQ-Rplus-1.1.xml
bradleeedmondson: Yep we've been trying to do this for all licenses. Anything with 1.1. and 1.2. gets two &lt;list&gt; levels, and could be more if it's 1.a.i.A., 1.a.i.B., 1.a.ii.A., etc.

jlovejoy:
@steparce - I did some more fixing on this one as the formatting was not great to begin with and the conversion missed some of the bullets, etc.  It is ready to be merged, but leaving it here if you want to copy format for other ones in this license family (or feel free to re-assign to me and pick up others that aren't so messed up!)

steparce:
Thanks @jlovejoy this is exactly what I needed! I will copy and use for the other ones in same license family. 
Merging per approval. 
--Stephanie Arce

PR #392, opened 2017-02-21
Add <obsoletedBy> to licenses and exceptions

wking:
From [here][1] (whose source I could not find):

> As a result, a number of licenses formerly included in the SPDX License List have been deprecated as licenses, and correct usage employs the License Expression Syntax as of v2.0.

So the sole reason for the deprecations seems to be “there's a better way to say that now”.

Telling people “use an expression instead” is less useful than saying “use this expression instead: `$SOME_EXPRESSION`”.  This commit updates the eCos license (the only deprecated license currently in this repository) to do that, using the [`WITH` exception syntax][2] and the [eCos exception label][3] defined in SPDX 2.1 (and possibly in earlier versions).

This is related to spdx/tools#73, although discussion there seemed to be doubling down on a deprecated boolean.

[1]: https://spdx.org/licenses/
[2]: https://spdx.org/spdx-specification-21-web-version#h.jxpfx0ykyb60
[3]: https://spdx.org/spdx-specification-21-web-version#h.ruv3yl8g6czd

### DIFF
--- a/src/GFDL-1.1.xml
+++ b/src/GFDL-1.1.xml
@@ -3,13 +3,11 @@
   <license licenseId="GFDL-1.1" isOsiApproved="false"
   name="GNU Free Documentation License v1.1"
   isDeprecated="true" deprecatedVersion="3.0">
+    <obsoletedBy>GFDL-1.1-only</obsoletedBy>
+    <obsoletedBy expression="GFDL-1.1+">GFDL-1.1-or-later</obsoletedBy>
     <crossRefs>
       <crossRef>http://www.gnu.org/licenses/old-licenses/fdl-1.1.txt</crossRef>
     </crossRefs>
-    <notes>
-      DEPRECATED: Deprecated in lieu of more
-      explicit license identifier of GFDL-1.1-only
-    </notes>
     <standardLicenseHeader>
       Copyright (c) <alt name="copyright" match=".+">YEAR YOUR NAME</alt>
       . Permission is granted to copy, distribute and/or modify this
--- a/src/GFDL-1.1.xml
+++ b/src/GFDL-1.1.xml
@@ -3,8 +3,10 @@
   <license licenseId="GFDL-1.1" isOsiApproved="false"
   name="GNU Free Documentation License v1.1"
   isDeprecated="true" deprecatedVersion="3.0">
-    <obsoletedBy>GFDL-1.1-only</obsoletedBy>
-    <obsoletedBy expression="GFDL-1.1+">GFDL-1.1-or-later</obsoletedBy>
+    <obsoletedBys>
+      <obsoletedBy>GFDL-1.1-only</obsoletedBy>
+      <obsoletedBy expression="GFDL-1.1+">GFDL-1.1-or-later</obsoletedBy>
+    </obsoletedBys>
     <crossRefs>
       <crossRef>http://www.gnu.org/licenses/old-licenses/fdl-1.1.txt</crossRef>
     </crossRefs>
--- a/schema/ListedLicense.xsd
+++ b/schema/ListedLicense.xsd
@@ -42,6 +42,7 @@
 			</documentation>
 		</annotation>
 		<all>
+			<element name="obsoletedBy" type="tns:obsoletedByType" minOccurs="0" maxOccurs="unbounded" />
 			<element name="crossRefs" type="tns:crossRefsType" minOccurs="0" maxOccurs="1"/>
 			<element name="notes" type="tns:formattedFixedTextType" minOccurs="0" maxOccurs="1"/>
 			<element name="text" type="tns:formattedAltTextType" minOccurs="0" maxOccurs="1"/>
@@ -61,6 +62,7 @@
 			</documentation>
 		</annotation>
 		<all>
+			<element name="obsoletedBy" type="tns:obsoletedByType" minOccurs="0" maxOccurs="unbounded" />
 			<element name="crossRefs" type="tns:crossRefsType" minOccurs="0" maxOccurs="1"/>
 			<element name="notes" type="tns:formattedFixedTextType" minOccurs="0" maxOccurs="1">
 				<annotation>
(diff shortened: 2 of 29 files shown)


goneall:
I like the <obsoleted-by> concept - it would convey important information.  I do have a couple of implementation concerns.  If we propogate this out to the JSON and RDF representations of the licenses, it would break current implementations which look for the deprecated flag.  There may also be cases in the future where we deprecate licenses without a replacement.  I would propose that we make this an optional attribute of licenses and exceptions and retain the notes and deprecated fields.

wking:
On Tue, May 30, 2017 at 11:00:48AM -0700, goneall wrote:
> There may also be cases in the future where we deprecate licenses
> without a replacement.  I would propose that we make this an
> optional attribute of licenses and exceptions and retain the notes
> and deprecated fields.

‘notes’ is so generic.  If we want to encourage notes about
deprecation-without-replacement, I'd recommend an ‘obsolete’ tag with
human-oriented content that explained why SPDX decided to drop the
license/exception in question.

Then when we build JSON / RDF / whatever from the XML, you can set a
‘deprecated’ boolean when either ‘obsoleted-by’ or ‘obsolete’ are set,
populate ‘notes’ appropriately, or whatever else we need to do for
backwards compat.  But I'd rather avoid repetition and unstructured
data in the canonical source where we can find DRY, structured way to
represent the information we're trying to pass along.

wking:
On Tue, May 30, 2017 at 09:50:42PM -0700, goneall wrote:
> Disagree that isDeprecated is duplicated.  You can have a deprecated
> license without it being obsoleted by another license.  obsoleted-by
> is just providing additional useful data.

But you can't have obsoleted-by *without* having deprecated.  Not
having a deprecated attribute means you don't have to worry about
validating the “obsoleted-by set but deprecated not set” case.  If in
the future we add other ways for deprecating a license/exception
(e.g. by removing it from the SPDX list without a replacement License
Expression), we can add a new tag to cover that case (e.g. obsolete
[1]).  So rather than requiring folks to set deprecated *and* some
more granular field, they'd just set the more granular field.  And
then when any of the granular fields were set you'd also set
deprecated in the compiled JSON.

[1]: https://github.com/spdx/license-list-XML/pull/392#issuecomment-304999964

goneall:
In thinking about this a bit more, I wonder if we should take a different approach.  obsoleted-by would be used to reference a preferred method of expressing the deprecated license.  Could we generalize this to a field to express an alternative license expression?  This has been proposed in past threads on the legal email list.  Something like "preferredAlternative" or "preferredExpression".  Then we would continue to use the deprecated flag to indicate a license has been deprecated and the alternative expression could be used with our without the deprecated licenses.

wking:
On Wed, May 31, 2017 at 10:11:09AM -0700, goneall wrote:
> Could we generalize this to a field to express an alternative
> license expression?  This has been proposed in past threads on the
> legal email list.

Can you link to that discussion?  I'm having trouble imagining a
situation where we have a different preferred License Expression but
do not deprecate the short identifier.

goneall:
I wasn't able to locate a link to the discussion (it was a while back).  I believe it was related to licenses which were a concatonation of other licenses.  We may have a similar scenario once we resolve the GPL-2.0+/GPL-2.0-only discussion.

My primary concern with removing the deprecated flag is that the semantics are clear and the information is very useful.  For example, I filter out all deprecated licenses when I show a selection of license choices.  In this scenario I don't need to refer to the obsoleted by expression.

wking:
On Wed, May 31, 2017 at 12:33:07PM -0700, goneall wrote:
> I believe it was related to licenses which were a concatonation of
> other licenses.

In this case, I'd prefer deprecating the short identifier for the
concatenated license and using obsoleted-by.

> We may have a similar scenario once we resolve the
> GPL-2.0+/GPL-2.0-only discussion.

As I said in [1], I'd prefer we stuck with GPL-2.0, etc. as the
license short name, since that's the license.  The + and -only
suffixes would be License Expression syntax for the license *grants*.
If that distinction holds water, there would never be short names for
+ or -only forms, so no need for special handling in the XML.

> My primary concern with removing the deprecated flag is that the
> semantics are clear and the information is very useful.

You can define clear semantics when you specify the obsoleted-by tag.
And you can recover the ease-of-use in compiled output without adding
redundant information to the XML source (because every entry that sets
obsoleted-by will be deprecated).

> For example, I filter out all deprecated licenses when I show a
> selection of license choices.  In this scenario I don't need to
> refer to the obsoleted by expression.

But you can check the obsoleted-by presence.  I'd rather not make
consumer-convenience a criterion for the source format.
Consumer-convenience is important, but you can add it during
compilation.  Your license-choice widget would be just as easy to
write when it used compiled output that included a deprecated boolean,
regardless of whether that boolean came from an explicit deprecated
attribute or from the presence of obsoleted-by in the source XML.

[1]: https://lists.spdx.org/pipermail/spdx-tech/2017-May/003355.html

wking:
I've just filed #583, which drops `isDeprecated` (like this PR), but it's leaning on the existing `deprecatedVersion` attribute instead of this PR's new `<obsoleted-by>`.  If/when #583 lands, I'll rebase this PR to use:

```xml
<obsoleted-by deprecatedVersion="3.0">AGPL-3.0-only</obsoleted-by>
```

where `version` would be a required attribute of `<obsoleted-by>`.  That would require further spdx/tools changes (probably any schema change will need spdx/tools changes), but it would protect us from providing a deprecation version without a migration plan or a migration plan without a deprecation version.

wking:
This is currently tagged “later release”.  If the holdup is concern over dropping `deprecated`, can we punt that particular issue to #583 and focus this more narrowly on the `<obsoleted-by>` addition?  I think having structured information for the recommended replacement would help as tools start migrating to the license-list 3.0 identifiers.  For example, being able to automatically generate warnings that suggest updating `GPL-3.0` to `GPL-3.0-only` would be nice.  Issues which discuss warning about deprecated identifiers include rust-lang/cargo#2039 and npm/npm#19558.

goneall:
>  If the holdup is concern over dropping deprecated, can we punt that particular issue to #583 and focus this more narrowly on the <obsoleted-by> addition?

I think this is a good approach.  If we separate out the issue on deprecated, we can probably close on the addition of the `<obsoleted-by>` rather quickly.  I agree this would be quite useful.

Since this is a broader change, we should get comments/feedback on the legal and tech mailing lists before closing.   At some point, this should be documented in the spec (e.g. in a [SPDX specification appendix](https://github.com/spdx/spdx-spec/issues/46), but that can be a separate effort.

@wking Do you want to update the title of this issue to reflect this change?

goneall:
Suggest changing the element name to `<obsoletedBy>` to be consistent with the naming conventions of other element names.  

Note - we have tried to have the same element names as the license RDFa and license list JSON representations which led to this particular convention.

wking:
I've pushed 7b56ba9 → 42b06b5, which:

* Renames the element from `<obsoleted-by>` → `<obsoletedBy>` following @goneall's [suggestion][1].
* Rebases onto master, so I had a number of newly-deprecated licenses to add `<obsoletedBy>` to.
* Added an XSD entry for the new element (now that we have an XSD :).
* Added an optional `expression` attribute, so `GFDL-1.1` can point to `GFDL-1.1-only` or `GFDL-1.1-or-later` depending on whether it was originally `GFDL-1.1+` or not (and similarly for other licenses affected by the `-only` / `-or-later` split).

[1]: https://github.com/spdx/license-list-XML/pull/392#issuecomment-357314992

src/GFDL-1.1.xml
```diff
@@ -3,13 +3,11 @@
   <license licenseId="GFDL-1.1" isOsiApproved="false"
   name="GNU Free Documentation License v1.1"
   isDeprecated="true" deprecatedVersion="3.0">
+    <obsoletedBy>GFDL-1.1-only</obsoletedBy>
+    <obsoletedBy expression="GFDL-1.1+">GFDL-1.1-or-later</obsoletedBy>
```
wking: The or-later-ness of this identifier is not clear to me, because (as of list 3.0) the grant for `GFDL-1.1` includes “[or any later version][1]”.  Because there is no `ONLY` operator, I'm guessing that `GFDL-1.1` means `GFDL-1.1-only` and that you'd need `GFDL-1.1+` (pre-3.0) to mean `GFDL-1.1-or-later`, despite the wording we claimed in `<standardLicenseHeader>`.  But we may want to update the standard license header and/or file an erratum (#598) for clarity going forward.

This issue also affects version 1.2 and 1.3 of the GFDL.

[1]: https://github.com/spdx/license-list-XML/blob/v3.0/src/GFDL-1.1.xml#L17

wking:
Rebased around #586 and #599 with 390fc63 → 9ba7946.  The Node tests are currently choking with:

```
Entity: line 65: element element: Schemas parser error : Element '{http://www.w3.org/2001/XMLSchema}element': Invalid value for maxOccurs (must be 0 or 1).
```

which is mentioned in albanm/node-libxml-xsd#12.  But `maxOccurs="unbounded"` [is in XSD 1.0 too][1], so this may just be a libxml bug.

[1]: https://www.w3.org/TR/xmlschema-1/#declare-element

wking:
Maybe we need a wrapper to attach a `<sequence>` too (like we have for `<crossRefs>`).  I'd rather avoid that if possible, but I'll fall back to it if we can't find a way to allow multiple `<obsoletedBy>` entries within an `<all>`.

wking:
Ah, despite `unbounded` being in the 1.0 spec, `maxOccurs` was restricted to 1 inside `<all>` in XSD 1.0.  The restriction was [lifted in 1.1][1].  The backing libxml issue is [here][2].

[1]: https://www.w3.org/TR/xmlschema11-1/#ch_models
[2]: https://bugzilla.gnome.org/show_bug.cgi?id=765936

wking:
@goneall , thoughts about [`maxOccurs="unbounded"` support][1]?  Should I duck by [following `<crossRefs>` and adding a wrapper][2]?  Maybe:

```xml
<obsoletedBy>
  <alternative>…</alternative>
  <alternative>…</alternative>
</obsoletedBy>
```

[1]: https://github.com/spdx/license-list-XML/pull/392#issuecomment-361440374
[2]: https://github.com/spdx/license-list-XML/pull/392#issuecomment-361439223

goneall:
> Should I duck by following <crossRefs> and adding a wrapper? 

I kind of like the structure of the wrapper.  This could allow us to add more meaning to the nested element names.  For example, we could have ```<alternativeExpression>...</alternativeExpression>``` for a license expression and ```<alternativeId>...</alternativeId>``` for alternative license ID's.  It would also allow us to expand to additional reasons for obsoleting a license that we have not yet thought of.

wking:
On Fri, Mar 23, 2018 at 09:27:27PM +0000, goneall wrote:
> For example, we could have
> ```<alternativeExpression>...</alternativeExpression>``` for a
> license expression and ```<alternativeId>...</alternativeId>``` for
> alternative license ID's.

I don't think we need that, because you can distinguish IDs from
expressions by looking for an internal space.  But I'll add the
wrapper to work around the XSD issues, and get this moving forward
again.

wking:
I've pushed 9ba7946 → 7a56340, rebasing onto master again and adding an `<obsoletedByes>` wrapper.  That lets me stick with `<obsoletedBy>` for the individual elements, which will make it easier for consumers to migrate when we eventually get an XSD 1.1 validator and can revert 7a56340.

wking:
@goneall, Travis is happy here :).  Want to take a final look when you get time?

wking:
> ...could we change this to "obsoletedBys"...

Done.

PR #394, opened 2017-03-04
update location of license-list repo

mlinksva:
Existing location no longer exists, spdx/license-list probably now current per https://spdx.org/news/news/2017/01/move-github-complete-tools (which could use a link perhaps to https://github.com/spdx but I'm not sure how the website is maintained).


jlovejoy:
Thanks @mlinksva ! good catch.

Issue #396, opened 2017-03-27
Error in SPDX version of Apache-1.1


bradleeedmondson:
Reported by Richard Fontana:

---------------------------------------------------

Hi,

The license text for Apache-1.1 
has for clause 4: "4. The "Apache" and "Apache Software Foundation" must not be used to endorse or promote products derived from this software without prior written permission. For written permission, please contact apache@apache.org" 

Assuming the following text at apache.org is or should be treated as authoritative:
https://www.apache.org/licenses/LICENSE-1.1

the SPDX text is in error, as it is missing the word "names" (and the closing period).

See also the template version at https://github.com/spdx/license-list/blob/master/Apache-1.1.txt#L15.

 Richard

bradleeedmondson:
Also reported by Richard: 

-----------------------------------

The same error exists in the SPDX text for Apache-1.0:
https://spdx.org/licenses/Apache-1.0.html
"4. The "Apache" and "Apache Software Foundation" must not be used to endorse or promote products derived from this software without prior written permission. For written permission, please contact apache@apache.org"

Compare to http://www.apache.org/licenses/LICENSE-1.0 :
*  4. The names "Apache Server" and "Apache Group" must not be used to
 *    endorse or promote products derived from this software without
 *    prior written permission. For written permission, please contact
 *    apache@apache.org.

jlovejoy:
have created PRs for both of these issues. to review on April 13 legal call, then merge 
issue from @richardfontana

bradleeedmondson:
fixed in #400 and #401 

resolved

PR #400, opened 2017-04-12
update clause 3 to original text

jlovejoy:
update clause 3 to match original text at http://www.apache.org/licenses/LICENSE-1.0 and add period at end of clause as per https://github.com/spdx/license-list-XML/issues/396

### DIFF
--- a/src/Apache-1.0.xml
+++ b/src/Apache-1.0.xml
@@ -32,10 +32,10 @@
         <li>
           <b>4.</b>
           <p>The 
-            <alt name="organizationClause4" match=".+">"Apache" and "Apache Software Foundation"</alt> must
+            <alt name="organizationClause4" match=".+">names "Apache Server" and "Apache Group"</alt> must
             not be used to endorse or promote products derived from this software without prior written
             permission. For written permission, please contact 
-            <alt name="contactClause4" match=".+">apache@apache.org</alt>
+            <alt name="contactClause4" match=".+">apache@apache.org</alt>.
           </p>
         </li>
         <li>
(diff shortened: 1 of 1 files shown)


goneall:
I would remove the spaced in name="nameClause4" match="name | names | name(s)" -> name="nameClause4" match="name|names|name(s)"

Other than that it looks fine

goneall:
@bradleeedmondson @jlovejoy I just realized there is one more fix needed - the parenthesis are special characters and need to be escaped with a "\".  match="name | names | name(s)" -> match="name|names|name\(s\)"  I looked at the most recent patch and the name pattern is no longer in the file - so I don't think anything further needs to be done.

bradleeedmondson:
No it's still there on line 41:
https://github.com/spdx/license-list-XML/blob/master/src/Apache-1.1.xml#L41

I'm not sure the escape character you intended came through in the comment (it may need to be escaped as well?). I'm assuming it's a backslash \ but let me know if that's wrong.

bradleeedmondson:
escaped parens with backslash in f1636ca7e60dbc4c79ba34e8d4014c97185f48df

goneall:
@bradleeedmondson The backslashes were removed in the comment - I did check your commit f1636ca and it looks correct

wking:
The branch from this PR (`jlovejoy-patch-1`) is still in this repo.  Can someone with write access remove it now that it's no longer needed?  And `jlovejoy-patch-2` from #401 is in the same boat.

bradleeedmondson:
removed

PR #401, opened 2017-04-12
update clause 4 to original text

jlovejoy:
updated clause 4 to match original text and added missing period at end of clause

### DIFF
--- a/src/Apache-1.1.xml
+++ b/src/Apache-1.1.xml
@@ -11,66 +11,64 @@
     <copyright>
       <p>Copyright (c) 2000 The Apache Software Foundation. All rights reserved.</p>
     </copyright>
-    <body>
-      <p>Redistribution and use in source and binary forms, with or without modification, are permitted provided
-         that the following conditions are met:</p>
-      <list>
-        <li>
-          <b>1.</b>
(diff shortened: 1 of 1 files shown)


goneall:
I believe we should remove the spaces in the match match="name | names | name(s)" -> match="name|names|name(s)"

Other than that, it looks fine to me

goneall:
@bradleeedmondson @jlovejoy I just realized there is one more fix needed - the parenthesis are special characters and need to be escaped with a "\".  match="name | names | name(s)" -> match="name|names|name\(s\)"  I looked at the most recent patch and the name pattern is no longer in the file - so I don't think anything further needs to be done.

bradleeedmondson:
No it's still there on line 41:
https://github.com/spdx/license-list-XML/blob/master/src/Apache-1.1.xml#L41

I'm not sure the escape character you intended came through in the comment (it may need to be escaped as well?). I'm assuming it's a backslash \ per regex guidelines but let me know if that's wrong.

bradleeedmondson:
escaped parens with backslash in 20a4fbd

goneall:
Looks like you got it fixed with 20a4fbd.  The github comments did indeed escape out my escape characters.

PR #406, opened 2017-05-23
Fix BSD-4-Clause.xml

zvr:
There was a double "the"


src/BSD-4-Clause.xml
```diff
@@ -28,7 +28,7 @@
           <b>3.</b>
           <p>All advertising materials mentioning features or use of this software must display the following
              acknowledgement: 
-            <br/>This product includes software developed by the 
+            <br/>This product includes software developed by
```
wking: I think you want to remove the trailing whitespace too, although I'm happy to rebase #410 to remove your trailing whitespace if this PR lands first.

Issue #404, opened 2017-04-25
ISC copyright and URL


bradleeedmondson:
See https://github.com/spdx/license-list/issues/7

jlovejoy:
updated URL here

jlovejoy:
note: need to add markup in copyright notice, also in the disclaimer
@goneall or @sam-ellis - could one of you do this?

goneall:
@jlovejoy is the request  to allow <optional> and <alt> within the <copyrightText> elements?  If so, this is non-trivial to update the XML schema and tooling.  Can we postpone this enhancement until after we release the current license list?

wking:
On Fri, Jul 28, 2017 at 11:51:03PM +0000, goneall wrote:
> @jlovejoy is the request to allow <optional> and <alt> within the
> <copyrightText> elements?

Currently *all* of copyright(Text) is optional [1].  But the matching
guidelines are a bit too ambiguous for machine detection.  For
example, it's clear (to a human reader) that the copyright notice
should be ignored (guideline 10), and that extraneous text at the end
of the license should be ignored (guideline 12), but other additional
text should result in a non-match (guideline 2.1.2).  However, there
are no rules about deciding if leading content is a copyright notice
(which should be ignored) or other substantive text (which should
result in a non-match).  To cover *that*, we'd want to require
copyright(Text) content to match, and, as you suggest, use <optional>
and <alt> within it to mark the replacable parts.

[1]: https://spdx.org/spdx-license-list/matching-guidelines
     entry 10 is about copyright notices

wking:
Interestingly, BSD-2-Clause already has an `<alt>` entry inside it's `<copyright(Text)>`, in both the [master branch][1] and [schemadev][2].  Does that need to be temporarily removed until the tooling catches up with it, or can we go ahead and add similar markup to other licenses where the project's copyright statement (vs. the copyright statement for the license text itself) is often included in the provided text?

[1]: https://github.com/spdx/license-list-XML/blob/42b7ea5bba502f5f835a73dbdd96d5aa8b11ce1d/src/BSD-2-Clause.xml#L8
[2]: https://github.com/spdx/license-list-XML/blob/2b7dd5d067561b8e27c1efbc35fc88851356edd7/src/BSD-2-Clause.xml#L10

goneall:
@wking Thanks for pointing out the BSD-2-Clause.  I checked and it does pass the XML validation and renders correctly.  We can embed the alt and var tags inside copyright text.

[...]

PR #423, opened 2017-07-29
ISC: Add <alt> tags

wking:
Following the [example][1] [set][2] by BSD-2-Clause, because users who are not the ISC are going to want to put their own name (or something generic like the BSD's “the copyright holders and contributors”) in those slots.  The OSI has [replaced the original “ISC” occurrences with “THE AUTHOR”][2.1] and not marked them for replacement.

I'm not sure about the naming.  I expect the values to be the same or similar for the two `copyrightHolder*` entries.  But the BSD's seem to use [different][2] [names][3] while other licenses like the MPL-1.1 seem to [reuse][4] [the][5] [same][6] [name][7].

Addresses part of spdx/license-list#7 (see also #404).

Also interesting, but probably unrelated, is that the ISC has [bumped their terminal copyright year from 2010 to 2013][8].  Presumably they're showing how the replacable text would change and are not claiming an extended copyright on the license content, because I see no changes in the license text.  I've left the terminal year alone for this commit.

[1]: https://github.com/spdx/license-list-XML/blob/2b7dd5d067561b8e27c1efbc35fc88851356edd7/src/BSD-2-Clause.xml#L10
[2]: https://github.com/spdx/license-list-XML/blob/2b7dd5d067561b8e27c1efbc35fc88851356edd7/src/BSD-2-Clause.xml#L30
[2.1]: https://opensource.org/licenses/ISC
[3]: https://github.com/spdx/license-list-XML/blob/2b7dd5d067561b8e27c1efbc35fc88851356edd7/src/BSD-2-Clause.xml#L33
[4]: https://github.com/spdx/license-list-XML/blob/2b7dd5d067561b8e27c1efbc35fc88851356edd7/src/MPL-1.1.xml#L20
[5]: https://github.com/spdx/license-list-XML/blob/2b7dd5d067561b8e27c1efbc35fc88851356edd7/src/MPL-1.1.xml#L21
[6]: https://github.com/spdx/license-list-XML/blob/2b7dd5d067561b8e27c1efbc35fc88851356edd7/src/MPL-1.1.xml#L23
[7]: https://github.com/spdx/license-list-XML/blob/2b7dd5d067561b8e27c1efbc35fc88851356edd7/src/MPL-1.1.xml#L26
[8]: https://www.isc.org/downloads/software-support-policy/isc-license/

### DIFF
--- a/src/ISC.xml
+++ b/src/ISC.xml
@@ -8,16 +8,17 @@
       <p>ISC License:</p>
     </title>
     <copyright>
-      <p>Copyright (c) 2004-2010 by Internet Systems Consortium, Inc. ("ISC") 
-        <br/>Copyright (c) 1995-2003 by Internet Software Consortium 
+      <p>Copyright (c) <alt name="copyright" match=".+">2004-2010 by Internet Systems Consortium, Inc. ("ISC")
+        <br/>Copyright (c) 1995-2003 by Internet Software Consortium</alt>
       </p>
     </copyright>
     <body>
       <p>Permission to use, copy, modify, and/or distribute this software for any purpose with or without fee is
          hereby granted, provided that the above copyright notice and this permission notice appear in all
          copies.</p>
-      <p>THE SOFTWARE IS PROVIDED "AS IS" AND ISC DISCLAIMS ALL WARRANTIES WITH REGARD TO THIS SOFTWARE INCLUDING
-         ALL IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL ISC BE LIABLE FOR ANY
+      <p>THE SOFTWARE IS PROVIDED "AS IS" AND <alt name="copyrightHolder" match=".+">ISC</alt> DISCLAIMS ALL
+         WARRANTIES WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF MERCHANTABILITY AND
+         FITNESS. IN NO EVENT SHALL <alt name="copyrightHolderLiability" match=".+">ISC</alt> BE LIABLE FOR ANY
          SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES WHATSOEVER RESULTING FROM LOSS OF
          USE, DATA OR PROFITS, WHETHER IN AN ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING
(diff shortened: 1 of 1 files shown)


wking:
Also odd about the ISC license is that [the title ends in a colon][1].  We only have three licenses like that in the current master (unless some slipped through my grep):

```
$ git grep -A1 '<title>' | grep ':</p>'
src/ISC.xml-      <p>ISC License:</p>
src/exceptions/openvpn-openssl-exception.xml-      <p>Special exception for linking OpenVPN with OpenSSL:</p>
src/exceptions/u-boot-exception-2.0.xml-      <p>The U-Boot License Exception:</p>
```

The [OSI does not include the trailing colon][2] and the ISC has (added?) [“Text of the” to their title][3].  I haven't touched any of that in this PR yet either, but we may want:

```xml
<title>
  <p><optional>Text of the </optional>ISC License<optional>:</optional></p>
</title>
```

[1]: https://github.com/spdx/license-list-XML/blob/2b7dd5d067561b8e27c1efbc35fc88851356edd7/src/ISC.xml#L9
[2]: https://opensource.org/licenses/ISC
[3]: https://www.isc.org/downloads/software-support-policy/isc-license/

wking:
And we may also want:

```xml
<p>Permission to use, copy, modify, and<optional>/or</optional> distribute…
```

to encode [the 2015-04-30 decision that the proposed ISC-OpenBSD wording was functionally equivalent to the stock ISC][1].

[1]: https://lists.spdx.org/pipermail/spdx-legal/2015-April/001398.html

jlovejoy:
re: colon in the title - it doesn't matter, as matching guidelines don't count the title for matching purposes, so I don't think there's any need for putting an <optional> tag inside the <title> tag.

re: adding the <optional> tag to the /or text - yes, that would be accurate -  good catch!

wking:
On Thu, Aug 03, 2017 at 08:36:46PM -0700, Jilayne Lovejoy wrote:
> re: colon in the title - it doesn't matter, as matching guidelines
> don't count the title for matching purposes, so I don't think
> there's any need for putting an <optional> tag inside the <title>
> tag.

That makes sense for human readers and matching, although see [1]
about “how do machines decide if something is allowed alternative
copyright (or title) text, and not forbidden leading substantive
text?”.  And even if we don't want to help machines by tightening up
title/copyright matching, it seems strange to not match the current
upstream's “Text of the ISC License:” [2].

> re: adding the <optional> tag to the /or text - yes, that would be
> accurate…

Pushed with 233ae57.

[1]: https://github.com/spdx/license-list-XML/issues/404#issuecomment-318792367
[2]: https://www.isc.org/downloads/software-support-policy/isc-license/

PR #425, opened 2017-07-31
CC-BY-SA-3.0: Add <alt> for “Ssection” typo

wking:
The typo exists in [the original][1], so instead of just fixing it, I'm using an `<alt>` entry.  That allows us to match both the original and the fixed version.

[1]: https://creativecommons.org/licenses/by-sa/3.0/legalcode


bradleeedmondson:
Good catch. Confirmed in original; regex match looks good. Approved & merged.

PR #78, opened 2016-03-30
Automated PR for FreeImage

myndzi:

PR #427, opened 2017-07-31
FTL: Replace dead freetype.fis.uniroma2.it and remove <copyright>

wking:
Cleanups for the just-landed #78.

That host no longer exists:

    $ ping freetype.fis.uniroma2.it
    ping: unknown host freetype.fis.uniroma2.it

The Savannah link I'm using comes from [here][1], so it's probably a good choice for a canonical URL, despite being a floating link into a Git archive.

I've removed the `<copyright>` tag, shifting the copyright line into the `<optional>` section, because the line seems to [apply to the license itself][1.5].  For example, see [here][2], where the bumped `FTL.TXT` copyright does not include 2004 but the bumped `README` copyright does.  Regardless of the author intentions, the remainder of the FTL "license" is clearly FreeType-specific, including a contact section.  So it seems very unlikely that another project would use the same license and only replace the line formerly contained within `<copyright>`.

[1]: https://www.freetype.org/license.html
[1.5]: https://wiki.spdx.org/view/Legal_Team/Templatizing/tags-matching
[2]: http://git.savannah.gnu.org/cgit/freetype/freetype2.git/commit/?id=68a15ebbe4f61999368a72a2953e257513a3cbba

### DIFF
--- a/src/FTL.xml
+++ b/src/FTL.xml
@@ -1,6 +1,7 @@
 <SPDX name="Freetype Project License" identifier="FTL" osi-approved="false">
   <urls>
     <url>http://freetype.fis.uniroma2.it/FTL.TXT</url>
+    <url>http://git.savannah.gnu.org/cgit/freetype/freetype2.git/tree/docs/FTL.TXT</url>
   </urls>
   <notes>This license was released 27 Jan 2006</notes>
   <license>
--- a/src/FTL.xml
+++ b/src/FTL.xml
@@ -9,10 +9,8 @@
       <p>The FreeType Project LICENSE</p>
       <p>2006-Jan-27</p>
     </title>
-    <copyright>
-      <p>Copyright 1996-2002, 2006 by David Turner, Robert Wilhelm, and Werner Lemberg</p>
-    </copyright>
     <optional>
+      <p>Copyright 1996-2002, 2006 by David Turner, Robert Wilhelm, and Werner Lemberg</p>
       <p>Introduction</p>
       <p>The FreeType Project is distributed in several archive packages; some of them may contain, in addition to
          the FreeType font engine, various tools and contributions which rely on, or relate to, the FreeType


jlovejoy:
I think the copyright tag should stay as it was. I'm not convinced this is the copyright for the license itself, but also it looks like it's varied over time. 
I think the optional tag probably shouldn't be there at all - some of these were added in the first-run automated conversion, but this text was not considered optional before

wking:
On Fri, Aug 04, 2017 at 04:55:40AM +0000, Jilayne Lovejoy wrote:
> I think the copyright tag should stay as it was. I'm not convinced
> this is the copyright for the license itself, but also it looks like
> it's varied over time.

Are you aware of any non-FreeType projects using this license?  As I
point out in the topic post for this PR, there's a lot of
FreeType-specific stuff in the license itself.  It's not clear to me
that we gain anything by allowing other projects to only adjust the
copyright line.

And the project is still active.  I'll look through their mail
archives and mail their list asking for clarification if I can't find
any archived comments.

> I think the optional tag probably shouldn't be there at all - some
> of these were added in the first-run automated conversion, but this
> text was not considered optional before

After all that <optional> stuff is a “Legal Terms” line and a §0 [1].
To me, that means that all the material before “Legal Terms” is
optional, non-binding front-matter.

[1]: https://github.com/spdx/license-list-XML/blob/7ecb7363bc82aedd0e293ca8825e348181619e6a/src/FTL.xml#L52-L55

bradleeedmondson:
per legal call today, will add new URL instead of replacing old one, and we agree w/ @wking that the copyright notice at the top is intended to apply to the license and should not have the &lt;copyright&gt; tag.

wking:
On Thu, Sep 14, 2017 at 05:39:37PM +0000, Bradlee H. Edmndson wrote:
> per legal call today, will add new URL instead of replacing old one…

Done with ce37a9b → 9935ac9.

Issue #414, opened 2017-06-13
Formatting Lists creates extra paragraphs


goneall:
In translating the XML into HTML, list items need a linefeed to be formatted properly as well as an indent.

To accomplish this, I am creating a paragraph with an indent attribute at the start of every list item.

Most of the license XML's include a paragraph tag after the bullet tag (see Apache-1.0.xml for an example).  This second paragraph tag creates a separate line feed. (e.g. the HTML ends up being <p style="margin-left: 40px;"> 1. <p>..." for the first bullet in Apache 1.0).

It seems like the best way to fix this is to remove the first paragraph tags that are inside the list items (if they exist).

I can add this to the utility which translates the old XML format to the new XML format, but it may remove more paragraphs than desired.

@bradleeedmondson Let me know what you think.

wking:
On Mon, Jun 12, 2017 at 06:50:42PM -0700, goneall wrote:
> Most of the license XML's include a paragraph tag after the bullet
> tag (see Apache-1.0.xml for an example).  This second paragraph tag
> creates a separate line feed. (e.g. the HTML ends up being <p
> style="margin-left: 40px;"> 1. <p>..." for the first bullet in
> Apache 1.0).

I think we should fix this by keeping the paragraph tags and updating
the styling.  Adding a left-margin to p tags feels a bit strange
anyway.  Perhaps you just want padding on the wrapping div (or
whatever tag holds the full license)?

> It seems like the best way to fix this is to remove the first
> paragraph tags that are inside the list items (if they exist).

This will make things awkward when you need multiple paragraphs within
a single list entry.  I'd rather keep them, and just fix the styling.

goneall:
I believe we want to do some formatting to the lists, otherwise there is no way to indicate whether text should be indented etc.

@wking Can you provide a proposed translation to HTML which would format the list properly?

Here is a snippet from the current Apache-1.0:

     <list>
        <item>
            <bullet>1.</bullet>
            <p>Redistributions of source code must retain the above copyright notice, this list of conditions
             and the following disclaimer.</p>
        </item>
        <item>
            <bullet>2.</bullet>
            <p>Redistributions in binary form must reproduce the above copyright notice, this list of conditions
             and the following disclaimer in the documentation and/or other materials provided with the
             distribution.</p>
        </item>
        <item>
            <bullet>3.</bullet>
            <p>All advertising materials mentioning features or use of this software must display the following
             acknowledgment: "This product includes software developed by the 
[...]

PR #430, opened 2017-08-03
Condor-1.1:  Add <alt> for "royalty- free" and use wayback URL

wking:
The extra space is part of the upstream HTML:

```
$ curl -s http://web.archive.org/web/20121029030238/http://research.cs.wisc.edu/condor/license.html | grep -A1 'royalty-$'
   the Software, you are granted a non-exclusive, worldwide, royalty-
   free perpetual license under such patent claims, with the rights
```

I've used an `<alt>` entry to allow both the appropriately space-less version (which I've set as canonical) and the upstream typo'ed version.

The new URL is because the old link redirects to the HTCondor license page:

```
$ curl -sI http://research.cs.wisc.edu/condor/license.html
HTTP/1.1 301 Moved Permanently
Date: Thu, 03 Aug 2017 17:54:54 GMT
Server: Apache
Location: http://research.cs.wisc.edu/htcondor/license.html
Content-Type: text/html; charset=iso-8859-1
```

which does not contain the old Condor 1.1 license.  Walking the archives for the non-HT page, the Condor 1.1 license is [still listed in 2012-10-29][1] but the [next archived result for that page on 2016-05-14 is the 301 redirect][2].

[1]: http://web.archive.org/web/20121029030238/http://research.cs.wisc.edu/condor/license.html
[2]: http://web.archive.org/web/20160514123949/http://research.cs.wisc.edu/condor/license.html

PR #410, opened 2017-06-01
src: Whitespace cleanup

wking:
Details in the commit messages, but this PR is about removing unnecessary whitespace from the source.  It keeps helpful whitespace (like indentation) and whitespace that is semantically important (e.g. between words ;).

PR #432, opened 2017-08-03
format-xml: Add a script for pretty-printing our XML source

wking:
Using [Python][1], because I'm familiar with it, and I've floated a Python script for spdx-spec in spdx/spdx-spec#10.  I'm not sure which tools/languages other license-list maintainers are familiar with.

Also using [LXML][2], since it's popular for XML editing within Python.

If you run this script on our current master, you get a lot of changes:

```
$ ./bin/format-xml.py src
$ git --no-pager diff --shortstat
 270 files changed, 16593 insertions(+), 16945 deletions(-)
```

I've left those changes off this PR for now while we discuss the feasability of this approach, but I can push them up publically somewhere else if folks want to browse them without running the script locally.

This is laying the ground-work for automatically manipulating our XML, e.g. converting to `<bullet>` (#414).

[1]: https://www.python.org/
[2]: http://lxml.de/

### DIFF
--- a/bin/format-xml.py
+++ b/bin/format-xml.py
@@ -1,0 +1,53 @@
+#!/usr/bin/env python3
+#
+# Pretty-print JSON, ordering attributes, stripping trailing
+# whitespace, nesting tags, etc.  This does not adjust the indenting
+# *within* text sections; you get to do that on your own.
+#
+# usage: ./bin/format-xml.py PATH...
+#
+# You can pass multiple PATHs to a single invocation, in which case
+# the script will format each PATH.  When PATH is a directory, the
+# script will recurse through its contents, formating every file which
+# ends in '.xml'.
+#
+# This script depends on LXML (http://lxml.de/)
+#
+# Script licensed under SPDX-License-Identifier: MIT
+
+import os
+import re
+import sys
+from lxml import etree
+
+
+_PARSER = etree.XMLParser(
+    ns_clean=True,
+    remove_blank_text=True,
+    remove_comments=True,
+    remove_pis=True,
+)
+
+
+def format_xml(path):
+    if os.path.isdir(path):
+        for dirpath, _, filenames in os.walk(path):
+            for filename in filenames:
+                if filename.endswith('.xml'):
+                    format_xml(path=os.path.join(dirpath, filename))
+        return
+    with open(path, 'r') as f:
+        original_content = f.read()
+    content = original_content.replace('\t', ' ')
+    tree = etree.fromstring(content, parser=_PARSER)
+    content = etree.tostring(tree, encoding='unicode', pretty_print=True)
+    content = re.sub(r'\s+$', '', content, flags=re.MULTILINE)
+    if content != original_content:
+        with open(path, 'wb') as f:
+            f.write(content.encode('UTF-8'))
+            f.write('\n'.encode('UTF-8'))
+
+
+if __name__ == '__main__':
+    for path in sys.argv[1:]:
+        format_xml(path=path)


zvr:
:-)
I've written a similar one myself.

But:
- let's separate the "inline" from the "block" tags
- you should get rid of all whitespace formatting on input
- maximum line length is nice

wking:
On Thu, Aug 03, 2017 at 02:39:18PM -0700, Alexios Zavras (zvr) wrote:
> - let's separate the "inline" from the "block" tags

That requires a mapping from the tags to inline/block.  And I'm not
sure if LXML provides a way to override their inline classification
(which may be stored in [1], I'm not sure).  If not, you'd need
something more sophisticated than the current ‘pretty_print=True’.

I think this would be nice, but not strongly enough to want to go
through that work ;).  Feel free to PR support for it, or to just PR
your own script.

> - you should get rid of all whitespace formatting on input

Do you mean my current tab-space conversion [2]?  I don't think we
want tabs in our output XML, and there are only a few licenses where
our current XML contains tabs:

  $ git grep -c "$(echo -en '\t')" origin/master
  origin/master:src/APL-1.0.xml:11
  origin/master:src/GPL-2.0-with-GCC-exception.xml:4
  origin/master:src/GPL-2.0-with-autoconf-exception.xml:6
  origin/master:src/GPL-2.0-with-bison-exception.xml:4
  origin/master:src/GPL-2.0-with-classpath-exception.xml:5
  origin/master:src/GPL-2.0-with-font-exception.xml:4
  origin/master:src/GPL-3.0-with-autoconf-exception.xml:26
  origin/master:src/OSET-PL-2.1.xml:193
  origin/master:src/eCos-2.0.xml:8

I see my tab→space conversion as part of a guard against future
accidental tab use, with something like Travis CI (or other CI
tooling) making sure that whatever pretty-printer we settle on finds
nothing to change in proposed PRs.

> - maximum line length is nice

I guess, although then you have to decide what it should be ;).  If we
do have an opinion there, I'd be happy to accept tooling to automate
it (note that we don't want to inject bugs like #430 and #431).

I'm also fine handling this by hand, since that seems like less work
to me then writing a reliable implementation.

In any case, I think the output from this script is a lot better than
what we have now ;).  I'm in favor of landing it or something similar,
and then iterating on those tools as we have time to improve things.

[1]: http://lxml.de/api/lxml.html.defs-module.html#special_inline_tags
[2]: https://github.com/spdx/license-list-XML/pull/432/files#diff-da7eafb36baacd32384e657599594a11R41

zvr:
very quickly, because I have to get ready for a flight in a few hours.
- yes, manual mapping of tags to the inline/block set; the built-in pretty_print was not enough, so I re-wrote it
- you actually want something like `re.sub(r'\s+', ' ', txt)`
- once the indent is calculated, the text is formatted to maxlength and lines filled, like fmt(1)

of course, as you say, none of this should be a blocking point, and I'd be happy to do another iteration later, when I get back

wking:
On Thu, Aug 03, 2017 at 10:12:32PM +0000, Alexios Zavras (zvr) wrote:
> - yes, manual mapping of tags to the inline/block set; the built-in
>   pretty_print was not enough, so I re-wrote it

Sounds good to me, as long as it's short and/or you're willing to
maintain it ;).

> - you actually want something like `re.sub(r'\s+', ' ', txt)`

I'm fine leaving things at this scale up to manual edits.  But if
we're comfortable not allowing doubled spaces, etc., then it's good to
have tooling that enforces it.  Are we comfortable forbidding doubled
spaces?

> - once the indent is calculated, the text is formatted to maxlength
>   and lines filled, like fmt(1)

Do you shell out to fmt?  It cites Knuth for detecting line breaks
[1], so I doubt it is worth rolling our own version of that.  Python
also has a library for this sort of thing [2], although you'd want to
set break_long_words [3] and break_on_hyphens [4] false to avoid
issues like #430 and #431.

> … of course, as you say, none of this should be a blocking point,
> and I'd be happy to do another iteration later, when I get back

And there's no rush either, so I'm happy to wait.

[1]: https://www.gnu.org/software/coreutils/manual/html_node/fmt-invocation.html
[2]: https://docs.python.org/3/library/textwrap.html
[3]: https://docs.python.org/3/library/textwrap.html#textwrap.TextWrapper.break_long_words
[4]: https://docs.python.org/3/library/textwrap.html#textwrap.TextWrapper.break_on_hyphens

jlovejoy:
we'll address this in a later release, leaving issue open for discussion then.

wking:
@zvr, have you had time to look over this or push forward with your alternative?  The current trailing whitespace issues (#410) are leading to conflicts between in-flight fixups like #496 (which removed trailing whitespace [here][1]) and #514 (which makes a semantic change to that line).  Basically, folks are contributing who seem to have editors which cleanup whitespace issues, and unless they go through the extra trouble to *preserve* those existing issues, there are likely to be more of these unnecessary merge conflicts.

[1]: https://github.com/spdx/license-list-XML/pull/496/files#diff-80c837db3b5a1f62ac112e832ad69929L19

zvr:
Yes, the tool is ok and I've been using it for the licenses I am adding.
For typical output, take a loot at #472 or #470 ...

wking:
> Yes, the tool is ok and I've been using it...

Can you PR it so we can fix all the other licenses?  I'd rather get that churn out of the way, and automated formatting would mean one less thing to manually review.

goneall:
> Can you PR it so we can fix all the other licenses? I'd rather get that churn out of the way, and automated formatting would mean one less thing to manually review.

Are we going to require all contributors to use the same pretty printing tool?  Seems like there is going to be a lot of churn if we use this for all licenses.  Not sure this is something we should be doing right before the release.

wking:
On Fri, Dec 15, 2017 at 05:59:21PM +0000, goneall wrote:
> Are we going to require all contributors to use the same pretty printing tool?

I'd like to, with a Makefile rule to make it easy to apply and a
Travis rule to enforce it.

> Seems like there is going to be a lot of churn if we use this for
> all licenses.

Yes, there will be a single-commit bump for all the existing licenses.
But the bump should be whitespace-only, so it should be easy to
review.  It will require rebasing for any open PRs that edit an
existing license, but I'm ok with that because we're seeing similar
conflicts already, even without shared format tooling [1].

> Not sure this is something we should be doing right before the
> release.

There are only 14 open pull requests, and with 5 each by @mlinksva and
me, that leaves 4 by other people which might need rebasing.  I don't
think that's too high a cost.  And if we want to hustle the 9 that
aren't mine in before running a reformat, I'm fine with that.

I would like to get it done before the release.  As I said above, I
expect it to be easy to audit.  And churning before the release means
that diffing between this release and the next will be much easier.

[1]: https://github.com/spdx/license-list-XML/pull/432#issuecomment-351805315

goneall:
>  And churning before the release means that diffing between this release and the next will be much easier.

Good point.  If we're going to pretty print all the XML, this would be the time to do it.  Suggest after we merge most of the other PR's to avoid re-basing.  

With all of the recent PR's, I'm expecting some changes once we test the XML files using the license generator tool.  I'll want to coordinate my fixes either before or after pretty printing.

wking:
On Fri, Dec 15, 2017 at 08:02:54PM +0000, goneall wrote:
> I'll want to coordinate my fixes either before or after pretty
> printing.

If we have the tool complete (and it sounds like we do), I'd recommend
folks who want to avoid rebasing hold off until we get the tool
landed.  So the order would be:

1. Any acceptable current PRs.
2. The formatting tool, with a mass reformat.
3. New PRs.

I'm fine with all of my existing PRs landing in phase 3 after I rebase
them, so don't hold up 2 on my account.

wking:
On Fri, Dec 15, 2017 at 08:02:54PM +0000, goneall wrote:
> > And churning before the release means that diffing between this
> > release and the next will be much easier.
>
> Good point.  If we're going to pretty print all the XML, this would
> be the time to do it.

With 3.0 cut, this particular ship has sailed.  But at the moment, the
*only* open PR (besides mine, which I'm happy rebasing) that is
touching existing license content is @mlinksva's #489.  The other
non-me PRs are adding new from-scratch content.  So if we can agree on
a pretty-printer (@zvr, still no PR for yours?), now seems like a
great time to start using it.

wking:
@zvr, still no PR for your pretty-printer?  In the absence of that PR, can we land this on so we can get mostly auto-formatted XML?  There are still very few PRs in flight that touch existing licenses, and if we are really concerned about merge conflicts I'm fine rolling out the auto-formatting in stages to work around any existing PRs.

zvr:
I have published the pretty printer in https://github.com/zvr/xmlindent -- last touched when the schema changed 5 months ago.
I admit I don't remember the current state, but I used it on the massive update we did back then.

goneall:
Discussed on the legal call 13 Dec 2016.  The online tools will pretty print edited XML files prior to the pull request, so this should not be needed in the normal workflow.  For other workflows, we would encourage submitter to pretty print prior to submit.  Pretty printing after submitting may cause issues identifying change history in git.  Decision made to close / reject the PR.

Issue #433, opened 2017-08-03
Handling stale URLs?


wking:
There are a number of licenses with stale URLs (e.g. the upstream URLs for the [AFL-3.0][1] and [Condor-1.1][2]).  @bradleeedmondson [pointed out][3] that there is no policy (or at least, no easily discoverable policy) around how those links are maintained.  Should stale URLs be:

* Removed from the XML source (we'd still have them in the Git archives).
* Kept in the XML source…
    * … with known-dead markup (e.g. `<url dead="true">…</url>`).
    * … without special markup.
* Replaced by Internet Archive links to old copies of the original URL (like [this][4]).

The presence of Internet Archive links in the current master (e.g. [here][5]) suggests that as the preferred approach, but it would be good to have an official legal-team policy on this.

If we decide to go with archive.org replacements, is archive.org the only acceptable archival source?  Or are other links (e.g. into mailing list archives, see my comments in #431) also acceptable?

[1]: https://github.com/spdx/license-list-XML/pull/11#r63288907
[2]: https://github.com/spdx/license-list-XML/pull/430
[3]: https://github.com/spdx/license-list-XML/pull/11#r131222292
[4]: https://github.com/spdx/license-list-XML/pull/430/commits/6a6384bd99d178f69e6b035f7ba676a1ad9e6772
[5]: https://github.com/spdx/license-list-XML/blob/b42efd5a6339ef7248d79d53a830b7d2caa07bf8/src/OSL-2.0.xml#L3

bradleeedmondson:
I agree that this should be made explicit, but on reflection I don't think this needs to hold up the next release. Removing immediate-release milestone.

wking:
While there is, as far as I know, still no official policy on this, recent legal-team decisions ([here][1], [here][2], and [here][3]) have supported a “keep dead links without special markup, and add working Internet Archive links where they can be found” approach.  Unmarked, known-dead links in the XML is my least favorite approach to handling known-dead links, but that's still probably the precedent to follow until we get an official policy.

[1]: https://github.com/spdx/license-list-XML/pull/430#issuecomment-329553704
[2]: https://github.com/spdx/license-list-XML/pull/427#issuecomment-329555901
[3]: https://github.com/spdx/license-list-XML/pull/419#issuecomment-329563093

wking:
[...]

PR #436, opened 2017-10-01
Sendmail: Add an Internet Archive link

wking:
The old URL currently redirects to a page that no longer contains the license text:

    $ curl -sIL http://www.sendmail.com/pdfs/open_source/sendmail_license.pdf | grep -i '^HTTP\|^location'
    HTTP/1.1 301 Moved Permanently
    Location: https://www.sendmail.com/pdfs/open_source/sendmail_license.pdf
    HTTP/2 302
    location: https://www.proofpoint.com/us/products/sendmail-sentrion/
    HTTP/2 301
    location: https://www.proofpoint.com/us/products/sendmail-sentrion
    HTTP/2 301
    location: https://www.proofpoint.com/us/products/mail-routing-agent
    HTTP/2 200

It's possible that the license text is somewhere on www.proofpoint.com, but if so neither Google nor I could find it.  The link I've used is the most recent to include the license text.  The Internet Archive's next access [on 2017-07-07 20:23:32 UTC resulted in a 302 redirect to the sendmail-sentrion/ URL][1].

I'd prefer to remove the old URL, but while we still don't have an official policy on that (#433), my previous attempts to replace dead links were [met with suggestions to add Internet Archive links while keeping the the known-dead links without special markup][3].

[1]: https://web.archive.org/web/20170707202332/https://www.sendmail.com/pdfs/open_source/sendmail_license.pdf
[3]: https://github.com/spdx/license-list-XML/issues/433#issuecomment-329582571

### DIFF
--- a/src/Sendmail.xml
+++ b/src/Sendmail.xml
@@ -1,6 +1,7 @@
 <SPDX name="Sendmail License" identifier="Sendmail" osi-approved="false">
   <urls>
     <url>http://www.sendmail.com/pdfs/open_source/sendmail_license.pdf</url>
+    <url>https://web.archive.org/web/20160322142305/https://www.sendmail.com/pdfs/open_source/sendmail_license.pdf</url>
   </urls>
   <license>
     <title>


bradleeedmondson:
Agree that we should consist-ify this in #433 but merging this for now.

PR #437, opened 2017-10-15
all CECILL licenses in English! 

zvr:
Since the authors of the licenses publish official versions in English (http://www.cecill.info/licences.en.html), it makes little sense for us to have the French versions.

I've created the English versions for CECILL-2.0, CECILL-2.1, CECILL-B and CECILL-C.
I have _not_ touched -1.0 or -1.1, since the license itself does not specify equivalence ("the French version shall be deemed authentic").


zvr:
No, we should delete the -fr one; I didn't want to do it before I got an OK for somebody else ;)

bradleeedmondson:
I agree we should delete the French ones, for now, then address translations all together after next release.

Issue #445, opened 2017-10-18
Remove CeCILL-C fr (for now)


bradleeedmondson:
We've said the canonical CeCILL C license file will, for now, be the English one. @goneall to remove CeCILL-C fr file after conversion work is done. (Or during would probably be fine too.)

bradleeedmondson:
Just didn't want to forget. 

cc: @zvr

goneall:
Added pull request #447 to remove

wking:
With #447 merged, can this be closed?

bradleeedmondson:
Yep

PR #446, opened 2017-10-18
Fix parse errors

goneall:
Fix parsing errors identified when validating against the XML schema.

There are individual commits for each file.

### DIFF
--- a/src/CC-BY-SA-2.0.xml
+++ b/src/CC-BY-SA-2.0.xml
@@ -227,7 +227,7 @@
         <item>
             <bullet>7.</bullet>
           Termination
-			<list>
+		<list>
                <item>
                   <bullet>a.</bullet>
 					This License and the rights granted hereunder will terminate automatically upon any breach by
--- a/src/CC-BY-SA-2.0.xml
+++ b/src/CC-BY-SA-2.0.xml
@@ -227,30 +227,31 @@
         <item>
             <bullet>7.</bullet>
           Termination
-        <list>
+			<list>
                <item>
                   <bullet>a.</bullet>
-            This License and the rights granted hereunder will terminate automatically upon any breach by
-               You of the terms of this License. Individuals or entities who have received Derivative
-               Works or Collective Works from You under this License, however, will not have their
-               licenses terminated provided such individuals or entities remain in full compliance with
-               those licenses. Sections 1, 2, 5, 6, 7, and 8 will survive any termination of this
-               License.
-          </item>
+					This License and the rights granted hereunder will terminate automatically upon any breach by
+					You of the terms of this License. Individuals or entities who have received Derivative
+					Works or Collective Works from You under this License, however, will not have their
+					licenses terminated provided such individuals or entities remain in full compliance with
+					those licenses. Sections 1, 2, 5, 6, 7, and 8 will survive any termination of this
+					License.
+				</item>
                <item>
-                  <bullet>b.</bullet>
-            Subject to the above terms and conditions, the license granted here is perpetual (for the
-               duration of the applicable copyright in the Work). Notwithstanding the above, Licensor
-               reserves the right to release the Work under different license terms or to stop
-               distributing the Work at any time; provided, however that any such election will not serve
-               to withdraw this License (or any other license that has been, or is required to be,
-               granted under the terms of this License), and this License will continue in full force and
-               effect unless terminated as stated above.
-          </item>
+					<bullet>b.</bullet>
+					Subject to the above terms and conditions, the license granted here is perpetual (for the
+					duration of the applicable copyright in the Work). Notwithstanding the above, Licensor
+					reserves the right to release the Work under different license terms or to stop
+					distributing the Work at any time; provided, however that any such election will not serve
+					to withdraw this License (or any other license that has been, or is required to be,
+					granted under the terms of this License), and this License will continue in full force and
+					effect unless terminated as stated above.
+				</item>
             </list>
-            <item>
-               <bullet>8.</bullet>
-          Miscellaneous
+		</item>
+        <item>
+            <bullet>8.</bullet>
+			Miscellaneous
         </item>
             <list>
                <item>
@@ -287,7 +288,6 @@
                mutual written agreement of the Licensor and You.
           </item>
             </list>
-        </item>
       </list>
       <p>Creative Commons is not a party to this License, and makes no warranty whatsoever in connection with the
          Work. Creative Commons will not be liable to You or any party on any legal theory for any damages
--- a/src/CECILL-C-fr.xml
+++ b/src/CECILL-C-fr.xml
@@ -291,7 +291,6 @@
                       <bullet>6.4</bullet>
                       MENTIONS DES DROITS
                       <p>Le Licencié s'engage expressément:</p>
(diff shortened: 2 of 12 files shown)


src/CC-BY-SA-2.0.xml
```diff
@@ -227,30 +227,31 @@
         <item>
             <bullet>7.</bullet>
           Termination
-        <list>
+			<list>
```
wking: The cange to this line looks like it's just injecting whitespace.  I'd rather leave that alone (or at least not indent beyond the child tag as you're doing here) until we get #432 or similar.

src/CECILL-C-fr.xml
```diff
@@ -291,15 +291,13 @@
...
                       <p>Le Licencié s'engage expressément:</p>
-                      <list>
                             <p>à ne pas supprimer ou modifier de quelque manière que ce soit les mentions de propriété
                              intellectuelle apposées sur le Logiciel;</p>
                             <p>à reproduire à l'identique lesdites mentions de propriété intellectuelle sur les copies du
                              Logiciel modifié ou non;</p>
                             <p>à faire en sorte que l'utilisation du Logiciel, ses mentions de propriété intellectuelle et le
                              fait qu'il est régi par le Contrat soient indiqués dans un texte facilement accessible
                              notamment depuis l'interface de tout Logiciel Dérivé.</p> 
-                      </list>
```
wking: This looks like it was meant to be a list.  Maybe keep the `<list>` and replace the internal `<p>` with `<item>`?

src/HPND.xml
```diff
@@ -29,20 +29,20 @@
          <optional>and</optional> that both <optional>that</optional> 
          <optional>the</optional> copyright notice 
          and this permission notice appear in supporting documentation<optional>, and that the name <optional>of</optional> 
-            <alt match=".*">&lt;copyright holder&gt;</alt> 
+            <alt match=".*" name="copyrightholder0">&lt;copyright holder&gt;</alt>
```
wking: Can we camelCase there names?  Or hyphenate them, or whatever, as long as we're consistent?  severalwordswithoutseparators is hard to read ;).

src/Nokia.xml
```diff
@@ -9,9 +9,7 @@
          <p>Version 1.0a</p>
       </titleText>
     
-      <list>
-        <item>
-            <bullet>1.</bullet>
+        <bullet>1.</bullet>
```
wking: `<bullet>` outside of an `<entry>` seems odd to me.  I thing we want to summport paragraphs in items (as HTML does) for cases like this where it's warranted.

An alternative here would be to use headers (and maybe allow `<bullet>` in them?) for these numbered sections.

src/OSET-PL-2.1.xml
```diff
@@ -7,7 +7,7 @@
          <crossRef>http://www.osetfoundation.org/public-license</crossRef>
       </crossRefs>
       <titleText>
-         <br>OSET Public License</br>
+         <br />OSET Public License
```
wking: Can we drop `<br/>` entirely?  I'd expect `<titleText>` to already mean "on its own line" (among other things).

src/CECILL-C-fr.xml
goneall: This license is deleted on a subsequent PR #445

src/Nokia.xml
goneall: We do support paragraphs inside items - reference [XML schema] (https://github.com/spdx/license-list-XML/blob/19dcfd1b31c7e81582e5f23b91c0667544d4a678/schema/ListedLicense.xsd#L83)

We do not support anything other than items as children of a list tag - this is what caused the original parsing errors.

In looking at the source license text formatting, I decided that removing the top level list would be closer to the formatting than adding items around the bullet tags.

The semantics on the bullet tag is information that can be used for matching where an item is formatting hints used to indent properly (translates to li in HTML).  I think it should be fine to allow bullets outside of items.

We don't have any concept of headers in the current XML spec - perhaps something to be considered for a future release.

src/OSET-PL-2.1.xml
goneall: Hmmm.  Not how the tool currently works.  titleText currently is only a hint for matching and does not carry any formatting significance.

A general question is - should XML properties have both matching and formatting semantics or should we keep them as separate tags?  I have not thought through this - we are likely inconsistent on this today.  I would lean to keeping the matching and formatting tags separate for maximum flexibility.

src/CECILL-C-fr.xml
goneall: Fixed in a subsequent commit

wking: > This license is deleted on a subsequent PR #445

So land that PR and rebase this one on top?

src/Nokia.xml
wking: > We don't have any concept of headers in the current XML spec - perhaps something to be considered for a future release.

I'd like to have a consistent way to address the numbered-(sub-)section use case, since many licenses do that.  Headers would be one way.  Using lists would be another, although it feels like you'd need a way to distinguish header-like list entries.  For example, currently [the GPL-2.0 does not mark up the “Preamble” header][1], although it's a header upstream:

```
$ curl -s https://www.gnu.org/licenses/old-licenses/gpl-2.0.html | grep Preamble
      <li><a id="TOC2" href="#SEC2">Preamble</a></li>
<h3 id="preamble"><a id="SEC2">Preamble</a></h3>
```

Our GPL-2.0 [does use lists to mark up the numbered sections][2], although [the “NO WARRANTY” header gets stuffed into the end of the previous `<item>`][3].

So I'm fine with whatever temporary workarounds you want to use, but I think we need to explicitly address the representation of upstream headers in the near future.

[1]: https://github.com/spdx/license-list-XML/blob/19dcfd1b31c7e81582e5f23b91c0667544d4a678/src/GPL-2.0.xml#L33
[2]: https://github.com/spdx/license-list-XML/blob/19dcfd1b31c7e81582e5f23b91c0667544d4a678/src/GPL-2.0.xml#L63-L65
[3]: https://github.com/spdx/license-list-XML/blob/19dcfd1b31c7e81582e5f23b91c0667544d4a678/src/GPL-2.0.xml#L251

src/OSET-PL-2.1.xml
wking: > `titleText` currently is only a hint for matching and does not carry any formatting significance.

Then maybe use `<p>` like most of our existing licenses?

```
$ git grep -A1 '<titleText>'
src/AAL.xml:      <titleText>
src/AAL.xml-         <p>Attribution Assurance License</p>
--
src/AAL.xml:      <titleText>
src/AAL.xml-         <p>ATTRIBUTION ASSURANCE LICENSE (adapted from the original BSD license)</p>
--
…
```

> A general question is - should XML properties have both matching and formatting semantics or should we keep them as separate tags?

Do `<titleText>` and `<optional>` have distinct matching semantics?  I'd expect them to be interchangeable.  If we get [explicit header markup][1], we can separate concerns by dropping `<titleText>` in favor of `<optional><h1>`, etc.

[1]: https://github.com/spdx/license-list-XML/pull/446#discussion_r145519436

PR #450, opened 2017-10-19
*GPL-3.0: Optional HTTPS for outgoing link

wking:
The FSF seems to have bumped their outgoing links from HTTP to HTTPS between 2017-09-28 and 2017-09-30 (detail for each license in the commit messages).  With this PR, I use `<alt>` tags to prefer the new canonical `https://`, but this will continue to match the old `http://`.

Spun off from github/choosealicense.com#543.

### DIFF
--- a/src/AGPL-3.0.xml
+++ b/src/AGPL-3.0.xml
@@ -20,7 +20,7 @@
         <br/>Version 3, 19 November 2007 
       </p>
       </titleText>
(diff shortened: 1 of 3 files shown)


wking:
I've pushed a84bcf4 → 268c89d changing from `<alt>` to the simpler `<optional>`, since `<optional>` is sufficient for these cases.

goneall:
@wking Never mind - found a somewhat reasonable fix for the tooling - we can leave this as is.

PR #452, opened 2017-10-19
schema/ListedLicense: Punt license text into <text>

wking:
As it stands, there's not an easy way to extract the license text from the XML format without removing `<crossRefs>`, `<notes>`, `<standardLicenseHeader>`, and other siblings.  With this change (which, if accepted, I still have to propagate into `src/`) we collect it in `<text>`, just like we already collect the header text in `<standardLicenseHeader>`.

This is a WIP, because of the pending `src/` migration.  If this PR gets a green light, I'll go ahead and migrate `src/` before we merge it.

### DIFF
--- a/schema/ListedLicense.xsd
+++ b/schema/ListedLicense.xsd
@@ -17,18 +17,11 @@
 		</choice>
 	</complexType>
 	<complexType name="ExceptionType" mixed="true">
-		<choice minOccurs="1" maxOccurs="unbounded">
+		<all>
 			<element name="crossRefs" type="tns:crossRefsType" minOccurs="0" maxOccurs="1"/>
 			<element name="notes" type="tns:formattedFixedTextType" minOccurs="0" maxOccurs="1"/>
-			<element name="titleText" type="tns:formattedAltTextType" minOccurs="0" maxOccurs="1"/>
-			<element name="copyrightText" type="tns:formattedAltTextType" minOccurs="0" maxOccurs="1"/>
-			<element name="bullet" type="string"/>
-			<element name="list" type="tns:listType"/>
-			<element name="p" type="tns:formattedAltTextType"/>
-			<element name="optional" type="tns:optionalType"/>
-			<element name="alt" type="tns:altType"/>
-			<element name="br" type="tns:emptyType"/>
-		</choice>
+			<element name="text" type="tns:formattedAltTextType" minOccurs="0" maxOccurs="1"/>
+		</all>
 		<attribute name="licenseId" type="string" use="required" />
 		<attribute name="isDeprecated" type="boolean"/>
 		<attribute name="name" type="string" use="required" />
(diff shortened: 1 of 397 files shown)


wking:
I've added a WIP commit showing the changes to Apache-2.0 and u-boot-exception-2.0.  If you remove all the other files you can use `gulp validate` to confirm the changes pass the new schema.

wking:
Rebased around #475 with 619e2f7 → 3eee08f.

goneall:
I think this is a good improvement (it will simplify some of the code in the tools).

One consideration is the name of the field.  In the JSON and RDF/XML formats of the license data, the license text uses a property name of ```licenseText``` (see the [Accessing SPDX License Document for reference](https://github.com/spdx/license-list-data/blob/master/accessingLicenses.md)).  The exceptions use ```licenseExceptionText```.

For consistency, we could use the same terms.

wking:
> One consideration is the name of the field. In the JSON and RDF/XML formats of the license data, the license text uses a property name of `licenseText` (see the Accessing SPDX License Document for reference). The exceptions use `licenseExceptionText`.

I'm generally in favor of simple names (like `text`) where the full context is available via the element context.  So `<license><text>…</text></license>` would be license text and `<exception><text>…</text></exception>` would be exception text.  But I'm not going to block on that preference; if you don't find my reasoning convincing and really prefer `licenseText` and `licenseExceptionText`, let me know, and I'll update this PR.

goneall:
> I'm generally in favor of simple names (like text) 

I was also thinking of the advantages of a simpler name.  It would actually be easier to implement with a simple name since there is common code for parsing the license and exception XML.

The history on the longer names comes from the use in tag/value which doesn't always have a hierarchical structure to pull context from.  Since we don't really have that issue here, I'm leaning toward ```text``` myself.  Let's just confirm on the legal call on 21 Nov. to make sure no one has any heartburn over this approach.  We can also close on adding this in before the release.

bradleeedmondson:
> I'm generally in favor of simple names (like text) where the full context is available via the element context. So <license><text>…</text></license> would be license text and <exception><text>…</text></exception> would be exception text.

+1 for this approach, as long as we're sure this doesn't introduce ambiguity with other formats downstream from the XML. In other words, the RDF/JSON/etc. still has the complete context.

wking:
> ...as long as we're sure this doesn't introduce ambiguity with other formats downstream from the XML.

Whatever tool is converting from XML can translate tag names too, so I don't think this will be a problem.

goneall:
>  as long as we're sure this doesn't introduce ambiguity with other formats downstream from the XML. 

I'll make sure the tool retains the same format when translating to the JSON and other formats.

wking:
Rebased around #519's `required` addition with 3eee08f → a31d982.

bradleeedmondson:
Marking for 3.1 per @wking and @goneall request

wking:
I can rebase this and inject the remaining `<text>` tags whenever.  Let me know when would be easiest for review.

goneall:
> Let me know when would be easiest for review.

Two considerations on timing I can think of - possible merge conflicts and impact on the tooling.

I'm thinking we start rebasing and reviewing soon.

For the possible merge conflicts with existing PR's:  Most of the PR's are @wking  and I can take care of any conflict with #570 so those won't be an issue.  The other PR's are #587 , #551 and #489.  This seems a small enough set to be manageable. 

For the tooling: The tool which tests the licenses texts will need to be updated with this XML schema change.  I suggest we review and merge these changes before we enable the license text testing in PR #593  This way, I can update the tool before we start using it.

wking:
Rebased onto master and added `<text>` to all the other licenses and exceptions with a31d982 → c478ba1.  I've removed the “WIP” from the PR subject, and this is ready for review and merging.

goneall:
Thanks @wking !

PR #473, opened 2017-11-01
Work around license matching tools issue

goneall:
The license matching tools currently does not support optional characters within tokens.  This PR will provide the same matching functionality and will allow the tools to work.  See https://github.com/spdx/tools/issues/114


zvr:
I'm not sure I understood, but wouldn't this also be needed for #450 ?

goneall:
@zvr Yes - this would be a problem for #450 as well.  It didn't get caught in the testing since the optional text was not present.

goneall:
Found a solution for the tooling, closing the PR

Issue #462, opened 2017-10-27
Decide if nested optional and alt tags should be allowed and update schema and/or tools appropriately


goneall:
As noted by @wking  in issue https://github.com/spdx/license-list-XML/issues/457#issuecomment-340051518, the schema currently allows for nested optional tags.  The SPDX tools currently disallows any nesting of alt or optional tags.

Either the schema or the tools should be updated once we decide if this should be allowed.

wking:
On Fri, Oct 27, 2017 at 07:24:41PM +0000, goneall wrote:
> This should be disallowed.

Previous discussion in #393.  To take HPND (#457) as an explicit
example, the current [1]:

   <optional>, and that the name <optional>of</optional> 
     <alt match=".*" name="copyrightHolder0">&lt;copyright holder&gt;</alt> 
     <optional>
       <alt match=".*" name="orRelated">&lt;or related entities&gt;</alt>
     </optional> 
     not be used in advertising or publicity pertaining to distribution of the software without specific, written
     prior permission</optional>.

would need to be rephrased to be a series of <alt> blocks (or a single
<alt> with a fairly complicated regex, or something).  If we decide to
go with the no-nesting approach anyway, I'd rather see #457 addressed
before adjusting the schema, to make sure we know what we're getting
into ;).

[1]: https://github.com/spdx/license-list-XML/blob/e5da40e25becb0aa7626d3f62649d2387284a623/src/HPND.xml

goneall:
Although the nested alt and optional tags was part of the discussion for issue #393 , I would like to separate out the issue of allowing formatting and allowing nested optional and alt tags.

I do not understand the expected behavior of the above example.  What is the expected behavior of a nested alt text inside an optional block?  How is it to be used in a matching scenario.?

wking:
On Sat, Oct 28, 2017 at 12:29:20AM +0000, goneall wrote:
> What is the expected behavior of a nested alt text inside an
> optional block?  How is it to be used in a matching scenario.?

Lets simplify the HPND case to:

  Permission to use this software is hereby granted, provided that the
  above copyright notice appear<optional> and that the name
  <optional>of</optional> <alt match=".*"
  name="copyrightHolder0">&lt;copyright holder&gt;</alt> not be
  used</optional>.

That will match:

  Permission to use this software is hereby granted, provided that the
  above copyright notice appear.

if the outer optional block is skipped entirely.  It will also match:

  Permission to use this software is hereby granted, provided that the
[...]

PR #474, opened 2017-11-01
Add optional tags

goneall:
so that the OSI referenced license at https://opensource.org/licenses/Multics will match

### DIFF
--- a/src/Multics.xml
+++ b/src/Multics.xml
@@ -30,9 +30,9 @@
 	  <optional>
 	  .
 	  </optional>
-	  <optional>
+	  <alt match="-*" name="divider">
 	  -----------------------------------------------------------
-	  </optional>
+	  </alt>
       <p>Permission to use, copy, modify, and distribute these programs and their documentation for any purpose
          and without fee is hereby granted, provided that the below copyright notice and historical background
          appear in all copies and that both the copyright notice and historical background and this permission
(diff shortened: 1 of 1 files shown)


src/Multics.xml
```diff
@@ -27,6 +27,12 @@
          successor in interest by change in name only to Honeywell Bull Inc. and Honeywell Information Systems
          Inc.</p>
       </optional>
+	  <optional>
+	  .
```
wking: Looking at [the OSI text][1], I read this extra period as part of the previous paragraph.  That would make this:

```xml
<p>This edition… Inc.<optional> .</optional></p>
```

[1]: https://opensource.org/licenses/Multics

src/Multics.xml
```diff
@@ -27,6 +27,12 @@
...
+	  </optional>
+	  <optional>
+	  -----------------------------------------------------------
```
wking: I'd rather not be picky about the number of dashes in this ASCII horizontal rule.  Can we use:

```xml
<alt match="-*"><p>-----------------------------------------------------------</p></alt>
```

or similar?

src/Multics.xml
goneall: This would create an optional text inside another optional text, which is not currently supported by the tools.

wking: > This would create an optional text inside another optional text, which is not currently supported by the tools.

Ah.  I''m fine leaving this alone until #462 is settled then.

src/Multics.xml
goneall: I added an alt tag similar, but without the paragraph tag in an updated PR

Issue #393, opened 2017-02-25
Should we allow formatting in the alt and optional elements


goneall:
Question: Should we allow paragraphs, bullets, lists inside the optional and alt elements?  I would assume we should disallow nested optional and alt elements (e.g. "<optional> this is an <optional> really </optional> optional text </optional>")

Note that the current schema definition only uses simple text for optional and alt.

wking:
On Fri, Feb 24, 2017 at 07:32:33PM -0800, goneall wrote:
> I would assume we should disallow nested optional and alt elements
> (e.g. "<optional> this is an <optional> really </optional> optional
> text </optional>")

As I understand it, the purpose of <optional> is to allow a license to
match whether or not it contains the optional content.  If there are
multiple flavors of the optional content, you'd need either nested
<optional> elements or a series of <optional> elements:

  <optional>flavor A</optional>
  <optional>flavor B</optional>
  <optional>flavor</optional>

Personally I'd rather allow nesting:

  <optional>flavor<optional> <alt name="flavor" match="A|B">A</alt></optional></optional>

because with longer optional/alternative text it allows us to avoid
repeating shared content.

bradleeedmondson:
@goneall I can say that we have a fair number of licenses with significant portions of text marked optional, so I'd think that we'd need to be able to include formatting tags within &lt;optional&gt; blocks to be able to display that in a readable way.

I have no opinion on nesting &lt;optional&gt; or &lt;alt&gt; tags within each other.

goneall:
Agree on the nesting of the formatting.  We'll need to update the schema for this.

I'm thinking that we do not nest <optional> and <alt> - I haven't seen this in practice and it would really complicate the schema.

@wking Your proposal is quite reasonable, however, we have already standardized on the current approach for the text files.  I would like to keep the basic format between the template text and HTML structurally similar.  If we were to change the structure of both, it would involve a revision of the spec.  I would suggest we go with the current approach for now and consider alternative formats in future revisions of the spec.  We would also want to take into account backwards compatibility since these formats have been encoded in license text outside of the XML.

wking:
On Wed, Mar 22, 2017 at 10:35:30AM -0700, goneall wrote:
> I would suggest we go with the current approach for now and consider
> alternative formats in future revisions of the spec.

I was just trying to demonstrate where nesting alt/optional would be
useful, and not trying to propose an alternate format.  I agree that
nesting alt/optional makes the schema and license definitions more
[...]

Issue #451, opened 2017-10-19
OFL-1.1 standard license header


wking:
We [currently don't declare `standardLicenseHeader` for the OFL-1.1][1], but they define one in their license document (above the title):

```
$ curl -sL 'http://scripts.sil.org/cms/scripts/page.php?item_id=OFL_web' |
>   w3m -T text/html -dump | grep -A7 'Copyright (c)' | cut -b 15-
Copyright (c) <dates>, <Copyright Holder> (<URL|email>),
with Reserved Font Name <Reserved Font Name>.

Copyright (c) <dates>, <additional Copyright Holder> (<URL
|email>),
with Reserved Font Name <additional Reserved Font Name>.

Copyright (c) <dates>, <additional Copyright Holder> (<URL|
email>).

This Font Software is licensed under the SIL Open Font License,
Version 1.1.

This license is copied below, and is also available with a FAQ
at: http://scripts.sil.org/OFL
```

And they *also* [discuss their recommendations for using it in their FAQ][2]:

> * Put your copyright and Reserved Font Names information at the beginning of the main OFL.txt file in place of the dedicated placeholders (marked with the `<>` characters). Include this file in your release package.
> * Put your copyright and the OFL text with your chosen Reserved Font Name(s) into your font files (the copyright and license fields). A link to the OFL text on the OFL web site is an acceptable (but not recommended) alternative. Also add this information to any other components (build scripts, glyph databases, documentation, test files, etc). Accurate metadata in your font files is beneficial to you as an increasing number of applications are exposing this information to the user. For example, clickable links can bring users back to your website and let them know about other work you have done or services you provide. Depending on the format of your fonts and sources, you can use template human-readable headers or machine-readable metadata. You should also double-check that there is no conflicting metadata in the font itself contradicting the license, such as the fstype bits in the os2 table or fields in the name table.

The first point suggests they intend the full copyright/license to be used as you'd use the GPL-3.0 license text (for example), by dropping it in a project level file.

The “Also add this information to any other components (build scripts…” in the second point suggests they intend the full copyright/license to also be their standard license header.  Our current XML schema doesn't provide an easy way to declare that the `standardLicenseHeader` is the same as the license text, but maybe it should?  Perhaps via some generic XML reference mechanism?  Otherwise we're going to have to duplicate this information in both places.

[1]: https://github.com/spdx/license-list-XML/blob/9984ebd50cf621010d9b24b95fa7dd47e42ba79a/src/OFL-1.1.xml
[...]

PR #475, opened 2017-11-01
schema/ListedLicense: Do not allow <alt> or <optional> inside <alt>

wking:
By differentiating between `formattedAltText…`, which may contain `<alt>` and `<optional>`, and `formattedFixedText…`, which may not.  The `optionalType` is using `formattedAltTextGroup`, because nesting variable content inside an optional element can be useful (#393, #462).  The `<alt>` element, on the other hand, specifies all the variable options in its `match` attribute, so there's no need for nesting variable elements inside the `<alt>` body.

This nibbles off another non-contentious point from #462, and may allow implementation simplification without loss of generality.  Depending on how #462 shakes out, we may transition `optionalType` from `formattedAltTextGroup` to `formattedFixedTextGroup` at a later date.

And there may be a DRYer XSD phrasing for these rules, but I'm not familiar enough with the syntax to work one up on short notice ;).

### DIFF
--- a/schema/ListedLicense.xsd
+++ b/schema/ListedLicense.xsd
@@ -19,12 +19,12 @@
 	<complexType name="ExceptionType" mixed="true">
 		<choice minOccurs="1" maxOccurs="unbounded">
 			<element name="crossRefs" type="tns:crossRefsType" minOccurs="0" maxOccurs="1"/>
-			<element name="notes" type="string" minOccurs="0" maxOccurs="1"/>
-			<element name="titleText" type="tns:formattedTextType" minOccurs="0" maxOccurs="1"/>
-			<element name="copyrightText" type="tns:formattedTextType" minOccurs="0" maxOccurs="1"/>
+			<element name="notes" type="tns:formattedFixedTextType" minOccurs="0" maxOccurs="1"/>
+			<element name="titleText" type="tns:formattedAltTextType" minOccurs="0" maxOccurs="1"/>
+			<element name="copyrightText" type="tns:formattedAltTextType" minOccurs="0" maxOccurs="1"/>
 			<element name="bullet" type="string"/>
 			<element name="list" type="tns:listType"/>
-			<element name="p" type="tns:formattedTextType"/>
+			<element name="p" type="tns:formattedAltTextType"/>
 			<element name="optional" type="tns:optionalType"/>
 			<element name="alt" type="tns:altType"/>
 			<element name="br" type="tns:emptyType"/>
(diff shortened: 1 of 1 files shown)


schema/ListedLicense.xsd
```diff
@@ -59,7 +59,7 @@
 	</complexType>
 	<complexType name="notesType" mixed="true">
 		<choice minOccurs="1" maxOccurs="unbounded">
-			<element name="p" type="tns:formattedTextType"/>
+			<element name="p" type="tns:formattedAltTextType"/>
```
goneall: Notes does not support alternative or optional text - should this be formattedFixTextType?  If we change it, will it cause a problem with the other elements named "p" in other elements?

schema/ListedLicense.xsd
```diff
@@ -39,12 +39,12 @@
 		<choice minOccurs="1" maxOccurs="unbounded">
 			<element name="crossRefs" type="tns:crossRefsType" minOccurs="0" maxOccurs="1"/>
 			<element name="notes" type="string" minOccurs="0" maxOccurs="1"/>
-			<element name="standardLicenseHeader" type="tns:formattedTextType" minOccurs="0" maxOccurs="1"/>
-			<element name="titleText" type="tns:formattedTextType" minOccurs="0" maxOccurs="1"/>
-			<element name="copyrightText" type="tns:formattedTextType" minOccurs="0" maxOccurs="1"/>
+			<element name="standardLicenseHeader" type="tns:formattedAltTextType" minOccurs="0" maxOccurs="1"/>
```
goneall: I believe we decided not to support optional and alt tags in the license header for this release.  That being said, I could not find any record of that decision (gave up after a 10 minute search).

schema/ListedLicense.xsd
wking: > If we change it, will it cause a problem with the other elements named "p" in other elements?

I din't think so.  Will rebase tonight.

schema/ListedLicense.xsd
wking: > I believe we decided not to support optional and alt tags in the license header for this release.

But thing like `ISC` have `<alt>` and `<optional>` and are the standard headers (although we don't say so yet, see #451).  Is there a techical problem with them?  Once we have alt/optional matching for the licenses, couldn't we use it for headers too?

goneall:
> Is there a techical problem with them?

No technical issue - just a SMOP (simple matter of programming).

I just added support for templates, it took about a day (assuming I didn't introduce any bugs ;)

schema/ListedLicense.xsd
goneall: @wking Are there any existing licenses using <alt> and/or <optional> in the standard headers?  If not, we can implement this in the next release of the license list.

wking: > Are there any existing licenses using and/or in the standard headers?

Yup:

```
$ grep -rlz '<standardLicenseHeader>.*\(<alt \|<optional>\).*</standardLicenseHeader>' src | sort
src/AGPL-3.0.xml
src/Apache-2.0.xml
src/CPAL-1.0.xml
src/CUA-OPL-1.0.xml
src/ECL-1.0.xml
src/ECL-2.0.xml
src/GFDL-1.1.xml
src/GFDL-1.2.xml
src/GFDL-1.3.xml
src/GPL-1.0.xml
src/GPL-2.0.xml
src/GPL-3.0.xml
src/Interbase-1.0.xml
src/LGPL-2.0.xml
src/LGPL-2.1.xml
src/LPPL-1.0.xml
src/LPPL-1.1.xml
src/LPPL-1.2.xml
src/LPPL-1.3a.xml
src/LPPL-1.3c.xml
src/MPL-1.0.xml
src/MPL-1.1.xml
src/OCLC-2.0.xml
src/RPSL-1.0.xml
src/SGI-B-1.0.xml
src/SGI-B-1.1.xml
src/SISSL-1.2.xml
src/SISSL.xml
src/SPL-1.0.xml
src/W3C-20150513.xml
src/W3C.xml
```

goneall: OK - sounds like we should support the markup in this release.   I logged 2 separate issues for the SPDX tools to support templates in the standard license header.  I do not think either of these issues are critical to releasing this version of the license list, so we can proceed with keeping this in the schema.

Issues logged: 

* Support templates for the standard license header in the library model: https://github.com/spdx/tools/issues/116
* Highlight the alt and optional text in the standard license headers: https://github.com/spdx/tools/issues/115

wking:
So is there anything else that needs to happen in this PR, or is it good to go?

goneall:
@wking Have you addressed comment on the notes above (not supporting alt types for notes)?

wking:
> Have you addressed comment on the notes above (not supporting alt types for notes)?

Ah, right.  I still need to do that.

wking:
Ok, I've made `<notes>` fixed with a1f48eb → 1825366.  This also drops the unused `notesType`.

I'm pretty sure we don't want `titleText` or `copyrightText` in the choices for either `formatted*TextGroup`, but I'm leaving that for #452.

goneall:
Changes look good to me.  I'll test out the schema using the license generator tool before merging.  It may take a couple of days since I am in the middle of updating the tools to support template functionality in the standard license header.

PR #480, opened 2017-11-12
Change alt text match to use whole tokens 

goneall:
so that the matching tools can more efficiently match the pattern in license OSL 2.1


wking:
Won't this be unnecessary based on [this comment][1]?

[1]: https://github.com/spdx/license-list-XML/pull/450#issuecomment-341213759

goneall:
@wking The fix referenced in the comment only works for single characters at the beginning and ending of the tokens.  I could fix characters in the middle, but performance would suffer.  I'm already worried about the performance of the matching code.

PR #482, opened 2017-11-15
Zlib: add copyrightText as often seen in wild, including crossrefs

mlinksva:
Just as easy to open a throwaway PR as describing in an issue, so that's what I've done. Please throwaway if useless. :smile:

### DIFF
--- a/src/Zlib.xml
+++ b/src/Zlib.xml
@@ -8,7 +8,9 @@
       <titleText>
          <p>zlib License</p>
       </titleText>
-    
+      <copyrightText>
+         <p>Copyright (c) &lt;year&gt; &lt;copyright holders&gt;</p>
+      </copyrightText>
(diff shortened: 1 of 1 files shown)


src/Zlib.xml
```diff
@@ -8,7 +8,9 @@
       <titleText>
          <p>zlib License</p>
       </titleText>
-    
+      <copyrightText>
+         <p>Copyright (c) &lt;year&gt; &lt;copyright holders&gt;</p>
```
wking: This occurs in the [upstream text][1] and in the [OSI page][2], so I think this is something we certainly want.  The “as often seen in the wild” caveat seems like it's closer to “as seen everywhere outside of SPDX, until it was fixed by this commit” ;).

[1]: http://zlib.net/zlib_license.html
[2]: https://opensource.org/licenses/Zlib

PR #486, opened 2017-11-23
Fix CECILL license so they will match the OSI license text

goneall:
Add optional footnote for license name description and fix title for CECILL-B - allows the license to match the text on the OSI website

### DIFF
--- a/src/CECILL-2.0.xml
+++ b/src/CECILL-2.0.xml
@@ -35,7 +35,7 @@ secondly, the election of a governing law, French law, with which it is conforma
         </item>
          </list>
          <p>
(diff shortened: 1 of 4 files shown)


zvr:
No issue from me.
I had used the text versions (e.g. http://www.cecill.info/licences/Licence_CeCILL-B_V1-en.txt) which do not have footnote explaining the acronym but rather inline text.

goneall:
@zvr Thanks for the pointer to the text - I made a few more changes to the templates so they match the fully copy/pasted text (including fixing a bug in my regular expression on the previous commit).

PR #488, opened 2017-11-24
Accommodate canonical CC-BY-4.0 text

mlinksva:
I'm not sure whether the accommodations I've made here are possible to support, just making them as it's easier than separately describing how https://creativecommons.org/licenses/by/4.0/legalcode.txt differs from the current XML representation.


src/CC-BY-4.0.xml
```diff
@@ -26,7 +26,7 @@
          license they choose before applying it. Licensors should also secure all rights necessary before
          applying our licenses so that the public can reuse the material as expected. Licensors should clearly
          mark any material not subject to the license. This includes other CC-licensed material, or material
-         used under an exception or limitation to copyright. More considerations for licensors.</p>
+         used under an exception or limitation to copyright. More considerations for licensors<alt match=": wiki.creativecommons.org/Considerations_for_licensors|\.">: wiki.creativecommons.org/Considerations_for_licensors</alt></p>
```
mlinksva: Note the the current text doesn't make much sense; the sentence only works with the link. It was probably copied from the CC HTML version and the link lost. I think it would be fine to drop the current text entirely and thus avoid alt match, that is unless it is important to match texts previously published by SPDX (I imagine it might be).

mlinksva:
Similar probably applies to other CC-*-4.0 licenses, could be applied after working out these changes.

src/CC-BY-4.0.xml
```diff
@@ -5,10 +5,10 @@
       <crossRefs>
          <crossRef>http://creativecommons.org/licenses/by/4.0/legalcode</crossRef>
       </crossRefs>
-      <titleText>
-         <p>Creative Commons Attribution 4.0 International</p>
-      </titleText>
       <optional>
+         <titleText>
+            <p><optional>Creative Commons </optional>Attribution 4.0 International</p>
+         </titleText>
```
mlinksva: I moved this inside the optional block because it doesn't make sense to have

```
title
optional
title
body
```

meaning that

```
title
title
body
```

is a reasonable rendering.

And note that "Creative Commons " is not included in the version provided by CC, thus made optional.

src/CC-BY-4.0.xml
goneall: I'm not certain, but it probably is important for matching previous texts.  I would leave the alt text in.

src/CC-BY-4.0.xml
goneall: I'm actually treating all titleText as optional when creating the license matching templates - so we don't really need to wrap the titleText with optional.

src/CC-BY-4.0.xml
```diff
@@ -35,7 +35,7 @@
          the licensed material may still be restricted for other reasons, including because others have
          copyright or other rights in the material. A licensor may make special requests, such as asking that
          all changes be marked or described. Although not required by our licenses, you are encouraged to
-         respect those requests where reasonable. More considerations for the public.</p>
+         respect those requests where reasonable. More considerations for the public<alt match=": wiki.creativecommons.org/Considerations_for_licensees|\.">: wiki.creativecommons.org/Considerations_for_licensees</alt></p>
```
wking: Upstream has an unfortunate typo underscore [here][1] (since [2014-01-10][2]), so I think we want something like:

```xml
… More<alt match=" |_"> </alt>considerations…
```

as well.

[1]: https://creativecommons.org/licenses/by/4.0/legalcode.txt
[2]: https://web.archive.org/web/20140110003133/https://creativecommons.org/licenses/by/4.0/legalcode.txt

mlinksva: Pointed this out to them at https://github.com/creativecommons/creativecommons.org/pull/714 ... maybe they'll fix going forward.

I'll go ahead and add your suggestion in any case.

src/CC-BY-4.0.xml
```diff
@@ -35,7 +35,7 @@
          the licensed material may still be restricted for other reasons, including because others have
          copyright or other rights in the material. A licensor may make special requests, such as asking that
          all changes be marked or described. Although not required by our licenses, you are encouraged to
-         respect those requests where reasonable. More considerations for the public.</p>
+         respect those requests where reasonable. More<alt match=" |_"> </alt>considerations for the public<alt match=": wiki.creativecommons.org/Considerations_for_licensees|\.">: wiki.creativecommons.org/Considerations_for_licensees</alt></p>
```
wking: This and other `<alt>` elements in this PR will need a `name` attribute to comply with 0d0e50c (#519) (see also #518, which has downstream tooling complaining about a missing `name`).

PR #487, opened 2017-11-24
Add optional "Creative Commons Legal Code" to title

mlinksva:
In order to match canonical text which includes the line https://creativecommons.org/publicdomain/zero/1.0/legalcode.txt

ref #434

PR #489, opened 2017-11-25
CC0-1.0 accommodation for text used in ~30k repos on github.com

mlinksva:
Evidence: https://github.com/search?utf8=%E2%9C%93&q=filename%3ALICENSE+%22For+more+information%2C+please+see+%3Chttp%3A%2F%2Fcreativecommons.org%2Fpublicdomain%2Fzero%2F1.0%2F%3E%22&type=Code

The changes stem from the CC0-1.0 text at https://choosealicense.com/licenses/cc0-1.0/

The text at choosealicense.com should be replaced with the text from [CC](https://github.com/github/choosealicense.com/pull/488) or [SPDX](https://github.com/github/choosealicense.com/pull/489), but it _might_ be worth matching texts with the trivial changes so widely deployed.

(This edit is independent of/should not conflict with #487, which addresses a different line present in the text from CC, but not in the text in this repo or at choosealicense.com.)


src/CC0-1.0.xml
```diff
@@ -6,7 +6,7 @@
          <crossRef>http://creativecommons.org/publicdomain/zero/1.0/legalcode</crossRef>
       </crossRefs>
       <titleText>
-         <p>Creative Commons CC0 1.0 Universal</p>
+         <p><optional>Creative Commons </optional>CC0 1.0 Universal</p>
```
wking: +1 for making “Creative Commons” optional.  Following [their `legalcode.txt`][1], we may want:

```xml
<titleText>
  <optional><p>Creative Commons<optional> Legal Code</optional></p><optional>
  <p>CC0 1.0 Universal</p>
</titleText>
```

Since [at least 2009-03-14][2], their HTML legal code has used `<h1>` and `<h2>` for those titles, but we don't have sized headers in our XML format (yet?).

[1]: https://creativecommons.org/publicdomain/zero/1.0/legalcode.txt
[2]: https://web.archive.org/web/20090314042016/https://creativecommons.org/publicdomain/zero/1.0/legalcode

src/CC0-1.0.xml
```diff
@@ -143,6 +143,6 @@
             </list>
         </item>
       </list>
-    
+      <optional><p>For more information, please see &lt;http://creativecommons.org/publicdomain/zero/1.0/&gt;</p></optional>
```
wking: Since we don't know where this line comes from (I can't find it anywhere on creativecommons.org, anyway), I'd rather use an `<alt>` to keep it out of the canonical text.  Something like:

```xml
<alt match="For more information, please see <http://creativecommons.org/publicdomain/zero/1.0/>" name="upstreamLink"></alt>

File: src/CC0-1.0.xml (hunk 1 above)
Comment (mlinksva) [reply src/CC0-1.0.xml]: https://spdx.org/licenses/CC0-1.0.html has "Creative Commons CC0 1.0 Universal" all on one line though.

File: src/CC0-1.0.xml (hunk 2 above)
Comment (mlinksva) [reply src/CC0-1.0.xml]: It was introduced at https://github.com/github/choosealicense.com/pull/126 with no explanation of that line but I guess it was an attempt to provide something equivalent to the "Back to Commons Deed" link present [at the time](http://web.archive.org/web/20131030222244/https://creativecommons.org/publicdomain/zero/1.0/legalcode). Anyway, you're right, I'll use your suggestion!

Comment (mlinksva) [reply src/CC0-1.0.xml]: Hmm, it seems `<p>` still warranted . One way to get it would be to wrap in `<optional>`, but this probably isn't ideal:


```html
      <optional><p><alt match="For more information, please see <http://creativecommons.org/publicdomain/zero/1.0/>" name="upstreamLink"></alt></p></optional>
```

It looks to me as if `<alt match>` hasn't been used with mixed content so far. Can it?

goneall: Yes - text between ```<alt ...``` and ```</alt>``` will be considered the "original text" and rendered in the web page and any generated text.  You can include formatting for that text.

wking: > Hmm, it seems `<p>` still warranted .

Why?  The canonical form has no content there; certainly not an empty paragraph.  And the `match` content need only match the whitespace-normalized license content (per the matching guidelines).  So I don't think we need `<p>` at all.

src/CC0-1.0.xml
```diff
@@ -143,6 +143,6 @@
             </list>
         </item>
       </list>
-    
+      <alt><p>For more information, please see &lt;http://creativecommons.org/publicdomain/zero/1.0/&gt;</p></alt>
```
wking: Does `<alt>` work without a `name` and `match`?  And while I think we want to match files which include this line, I don't think we want to include it in our canonical text.  So I still prefer the [phrasing I'd suggested here][1].

[1]: https://github.com/spdx/license-list-XML/pull/489#discussion_r153060299

mlinksva: OK, I didn't fully understand -- it's including text in a `match` that keeps it out of canonical text, not wrapping it in `<alt>` -- I'll go with your suggestion, thanks for explaining.

src/CC0-1.0.xml
mlinksva: You're right, it didn't occur to me that valid HTML output is irrelevant for text not included in the output. :smile:

wking: > … it didn't occur to me that valid HTML output is irrelevant for text not included in the output.

I think there's a larger issue here about “how do we inform folks reading HTML renderings of the set of allowed alternatives?”.  The [current previews][1] just use:

```html
<span class="replacable-license-text">&lt;copyright holder&gt;</span>
```

and similar.  I think we should at least add a `title` attribute with the regexp, although I'm not very excited about exposing the regexps directly to end-users, many of whom may not be familiar with regexps.

[1]: https://spdx.org/licenses/preview/HPND.html

jlovejoy:
I think this should use the optional tag as the text in question could be there or missing and still be a match. This is how we've defined/used the optional tag otherwise, so we should be consistent

jlovejoy:
this also will allow the text to be displayed on the website as optional, like this: https://spdx.org/licenses/Apache-1.1.html (in blue text)

mlinksva:
@jlovejoy I started with optional and switched to alt at @wking's suggestion in order to keep the text out of the text displayed on the website, see https://github.com/spdx/license-list-XML/pull/489#discussion_r153060299 ... the text should be matched as there are lots of legitimate copies of the license with it, but there's no reason for a licensor to include the nonstandard text. Alt hides it while enabling matching, IIUC. Happy to be wrong!

jlovejoy:
as this ties into a bigger conversation as to how/when we use these tags, we are going to boot this for further discussion as it could have a knock-on effect with other licenses/markup

wking:
On Thu, Dec 21, 2017 at 07:03:33PM +0000, Jilayne Lovejoy wrote:
> as this ties into a bigger conversation as to how/when we use these
> tags…

I think we've *been* using alt vs. optional based on which form we
consider canonical (e.g. see #429, which very narrowly bears on this
issue).  There is work on display for empty-canonical-alts in
spdx/tools#120, but I think that's a tooling issue and not a “what
should we put in the template XML?” issue.

> … we are going to boot this for further discussion as it could have
> a knock-on effect with other licenses/markup

What sort of knock-on effect do you foresee?  I don't see a reason to
punt on this at the moment.

src/CC0-1.0.xml
```diff
@@ -147,6 +147,7 @@
             </list>
         </item>
       </list>
+      <alt match="For more information, please see <http://creativecommons.org/publicdomain/zero/1.0/>" name="upstreamLink"></alt>
```
wking: You may need `&lt;`, etc. escapes in the `match` value to get valid XML.

mlinksva:
The remaining failure https://travis-ci.org/spdx/license-list-XML/jobs/334218553#L471-L472 looks to be related to #487 rather than this change:

> 	Test for license CC0-1.0 failed: Test for license ID CC0-1.0 failed due to difference found Variable text rule upstreamLink did not match the compare text at end of text while processing rule var: upstreamLink.  Last optional text was not found due to the optional difference: 
> 	Normal text of license does not match starting at line 1 column 21 "1" when comparing to template text " Legal Code"

src/CC0-1.0.xml
```diff
@@ -147,6 +147,7 @@
             </list>
         </item>
       </list>
+      <alt match="For more information, please see &lt;http://creativecommons.org/publicdomain/zero/1.0/>" name="upstreamLink"></alt>
```
wking: I think you need `>` -> `&gt;` in the `match` value too.

mlinksva: It's not [necessary](https://www.w3.org/TR/xml/#syntax), and the validate-schema test passes. Happy to escape it though if that's stylistically preferred. The remaining failing test isn't contingent on the `>` being escaped or not (tried locally).

mlinksva:
That's actually two distinct errors:

> 	Test for license CC0-1.0 failed: Test for license ID CC0-1.0 failed due to difference found Variable text rule upstreamLink did not match the compare text at end of text while processing rule var: upstreamLink

This can be "fixed" by replacing the `<alt match...` with `<optional...`. But that then fails to distinguish between canonical and non-canonical text.

> Last optional text was not found due to the optional difference: 
	Normal text of license does not match starting at line #1 column #21 "1" when comparing to template text " Creative Commons Legal Code"

This can be "fixed" with something like

```xml
      <titleText>
         <optional><p>Creative Commons Legal Code</p></optional>
         <p><optional>Creative Commons </optional>CC0 1.0 Universal</p>
      </titleText>
```

but that allows for more variations than [seen in the wild](https://github.com/spdx/license-list-XML/pull/487#issuecomment-352914558).

I suspect both of these have to do with the matcher implementation being slightly less flexible than the schema would imply, but I haven't tried to confirm this by looking at the code. I'll play with it some more, just leaving this as a note to self, but hints welcome. :smile:

mlinksva:
I don't understand the knock-on causal relation, but you're right. Fixed the second by wrapping the `alt` with empty default in `optional`. Not certain that's the right thing, but the tests pass now...

goneall:
Just catching up on the conversation.  It looks like we ended with a reasonable solution IMHO.

BTW - the error reporting for parsing errors is not very satisfactory for the tool that compare the code to the canonical text.  I've been working on trying to improve it, but it can get quite complicated when there are nested alt's and optionals and it is not clear where the error occurs.  Suggestions and tool improvements are welcome.

wking:
On Sat, Jan 27, 2018 at 09:37:32PM -0800, goneall wrote:
> It looks like we ended with a reasonable solution IMHO.

If the matching implementation has trouble with empty alt elements,
perhaps it could wrap them in optional tags as a pre-processing step
so we didn't have to remember to do that here?  On the other hand,
test failures will remind us, so not a big deal if we leave things as
they stand.

bradleeedmondson:
Putting a ping on @goneall @mlinksva @wking -- did we reach a decision on this? Current result of this PR seems to substantively add only [one line](https://github.com/mlinksva/license-list-XML/blob/f18f377b92872151863117995addde01be8e3535/src/CC0-1.0.xml#L150), and not address the title question.

wking:
> Current result of this PR seems … not address the title question.

Doesn't it?  I thought the issue was that choosealicense had been including the line that this PR is adding.  It's not part of the upstream text, and with the empty `<alt>` here the new line is not included in our canonical wording.  But the new empty `<alt>` does allow for matching choosealicense's previous wording, which is what the title is claiming.

bradleeedmondson:
> Doesn't it? I thought the issue was that choosealicense had been including the line that this PR is adding. It's not part of the upstream text, and with the empty <alt> here the new line is not included in our canonical wording. But the new empty <alt> does allow for matching choosealicense's previous wording, which is what the title is claiming.

I didn't grok this well enough to resolve that this was the outstanding issue. Basically if the previous commenters you, @mlinksva, and @goneall are happy (that that's the question, and it's adequately resolved), then I'm happy.

mlinksva:
The title question was addressed in #487, which has been merged. This PR is also good to go.

bradleeedmondson:
Based on today's discussion here and on @goneall review on Jan 28, I understand this is ready to merge. Glad we'll be able to match on a commonly used variant; thanks all.

PR #491, opened 2017-11-25
OFL-1.1 optionally templated

### LINKED ISSUE #451: OFL-1.1 standard license header

mlinksva:
See previous discussion at #451

See https://opensource.org/licenses/OFL-1.1 and instructions at http://scripts.sil.org/cms/scripts/page.php?site_id=nrsi&id=OFL#83afb016 and many in the wild at https://github.com/search?q=license%3Aofl-1.1

### DIFF
--- a/src/OFL-1.1.xml
+++ b/src/OFL-1.1.xml
@@ -6,10 +6,11 @@
          <crossRef>http://www.opensource.org/licenses/OFL-1.1</crossRef>
       </crossRefs>
       <notes>This license was released: 26 February 2007.</notes>
-      <optional>
-         <copyrightText><p>Copyright (c) &lt;dates&gt;, &lt;Copyright Holder&gt; (&lt;URL|email&gt;),
+      <copyrightText><p>Copyright (c) &lt;dates&gt;, &lt;Copyright Holder&gt; (&lt;URL|email&gt;),
             <br/>with Reserved Font Name &lt;Reserved Font Name&gt;.</p></copyrightText>
-         <p>This Font Software is licensed under the SIL Open Font License, Version 1.1. This license is copied below, and is also available with a FAQ at: http://scripts.sil.org/OFL</p>
+      <optional>
+         <p>This Font Software is licensed under the SIL Open Font License, Version 1.1.</p>
+         <p>This license is copied below, and is also available with a FAQ at: http://scripts.sil.org/OFL</p>
       </optional>
       <titleText>
          <p>SIL OPEN FONT LICENSE</p>
--- a/src/OFL-1.1.xml
+++ b/src/OFL-1.1.xml
@@ -6,6 +6,11 @@
          <crossRef>http://www.opensource.org/licenses/OFL-1.1</crossRef>
       </crossRefs>
       <notes>This license was released: 26 February 2007.</notes>
+      <optional>
+         <copyrightText><p>Copyright (c) &lt;dates&gt;, &lt;Copyright Holder&gt; (&lt;URL|email&gt;),
+            <br/>with Reserved Font Name &lt;Reserved Font Name&gt;.</p></copyrightText>
+         <p>This Font Software is licensed under the SIL Open Font License, Version 1.1. This license is copied below, and is also available with a FAQ at: http://scripts.sil.org/OFL</p>
+      </optional>
       <titleText>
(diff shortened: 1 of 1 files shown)


src/OFL-1.1.xml
```diff
@@ -6,6 +6,11 @@
          <crossRef>http://www.opensource.org/licenses/OFL-1.1</crossRef>
       </crossRefs>
       <notes>This license was released: 26 February 2007.</notes>
+      <optional>
```
wking: `copyrightText` is already optional ([guideline 10.1.1][1]).  I'm in favor of explicit `<optional>` tags for optional copyright text, but that's not where this repo is now.  I think we should shift this line after `<copyrightText>` for this PR and move it back to here it if/when the whole repository moves to use explicit `<optional>` tags for `<copyrightText>`.

[1]: https://spdx.org/spdx-license-list/matching-guidelines

src/OFL-1.1.xml
```diff
@@ -6,6 +6,11 @@
...
+         <copyrightText><p>Copyright (c) &lt;dates&gt;, &lt;Copyright Holder&gt; (&lt;URL|email&gt;),
+            <br/>with Reserved Font Name &lt;Reserved Font Name&gt;.</p></copyrightText>
+         <p>This Font Software is licensed under the SIL Open Font License, Version 1.1. This license is copied below, and is also available with a FAQ at: http://scripts.sil.org/OFL</p>
```
wking: The “This license is copied below…” bit is a separate paragraph [upstream][1].

[1]: http://scripts.sil.org/cms/scripts/page.php?item_id=OFL_web

src/OFL-1.1.xml
mlinksva: Thanks, agreed on all points.

mlinksva:
OK, I removed 'fixes'.

PR #494, opened 2017-11-25
Add angle brackets missing from LGPL-2.1 how-to trailer

mlinksva:
cf https://www.gnu.org/licenses/old-licenses/lgpl-2.1.txt

(GPL-2.0 has the same issue.)

### DIFF
--- a/src/LGPL-2.1.xml
+++ b/src/LGPL-2.1.xml
@@ -421,8 +421,8 @@
          <p>To apply these terms, attach the following notices to the library. It is safest to attach them to the
          start of each source file to most effectively convey the exclusion of warranty; and each file should
          have at least the "copyright" line and a pointer to where the full notice is found.</p>
-         <p>one line to give the library's name and an idea of what it does. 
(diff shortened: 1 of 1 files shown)


src/LGPL-2.1.xml
```diff
@@ -421,8 +421,8 @@
          <p>To apply these terms, attach the following notices to the library. It is safest to attach them to the
          start of each source file to most effectively convey the exclusion of warranty; and each file should
          have at least the "copyright" line and a pointer to where the full notice is found.</p>
-         <p>one line to give the library's name and an idea of what it does. 
-        <br/>Copyright (C) year name of author 
+         <p>&lt;one line to give the library's name and an idea of what it does.&gt;
```
wking: I've changed my mind and now think the new `&lt;` and similar should be `<optional>`.  More on why [here][1], and also to preserve matching for folks who used our old license text.

[1]: https://github.com/spdx/license-list-XML/pull/496/files#r153306757

PR #496, opened 2017-11-26
Add angle brackets and br missing from GPL-2.0 how-to trailer

mlinksva:
Same as #494 and LGPL-2.1, plus the missing br

### DIFF
--- a/src/GPL-2.0.xml
+++ b/src/GPL-2.0.xml
@@ -280,7 +280,8 @@
          <p>To do so, attach the following notices to the program. It is safest to attach them to the start of each
          source file to most effectively convey the exclusion of warranty; and each file should have at least
          the "copyright" line and a pointer to where the full notice is found.</p>
-         <p>one line to give the program's name and an idea of what it does. Copyright (C) yyyy name of author</p>
+         <p>&lt;one line to give the program's name and an idea of what it does.&gt;
+         <br/>Copyright (C) &lt;yyyy&gt; &lt;name of author&gt;</p>
          <p>This program is free software; you can redistribute it and/or modify it under the terms of the GNU
          General Public License as published by the Free Software Foundation; either version 2 of the License,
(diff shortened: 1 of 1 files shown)


src/GPL-2.0.xml
```diff
@@ -280,7 +280,8 @@
          <p>To do so, attach the following notices to the program. It is safest to attach them to the start of each
          source file to most effectively convey the exclusion of warranty; and each file should have at least
          the "copyright" line and a pointer to where the full notice is found.</p>
-         <p>one line to give the program's name and an idea of what it does. Copyright (C) yyyy name of author</p>
+         <p>&lt;one line to give the program's name and an idea of what it does.&gt;
+         <br/>Copyright (C) &lt;yyyy&gt; &lt;name of author&gt;</p>
```
wking: `<br/>` looks good to me.  I'm having second thoughts about the `&lt;` and such.  They're in the FSF's text version:

```
$ curl -s https://www.gnu.org/licenses/old-licenses/gpl-2.0.txt | grep -A1 'what it does'
    <one line to give the program's name and a brief idea of what it does.>
    Copyright (C) <year>  <name of author>
```

But the FSF seems to use [`<var>`][1] in HTML:

```
$ curl -s https://www.gnu.org/licenses/gpl-2.0.html | grep -A1 'what it does'
<var>one line to give the program's name and an idea of what it does.</var>
Copyright (C) <var>yyyy</var>  <var>name of author</var>
```

which is perfect, with the W3C including “[a term used as a placeholder in prose][1]” as one possible use-case.  Although Firefox doesn't seem do style it when used with a `<pre>`, and we can't put flow content inside `<var>` (spdx/tools#121).

So my personal preference is to use `<optional>&lt;</optional>` and similar for these.

[1]: http://w3c.github.io/html/textlevel-semantics.html#the-var-element

Issue #391, opened 2017-02-20
Document the XML schema somewhere in this repository


wking:
Also documenting more about how this project is supposed to work (e.g. what content is under which licenses?  How folks should go about contributing, etc.) would be nice ;).

bradleeedmondson:
Thanks Trevor, those are definitely in the plans. As you may have gleaned, SPDX is in the midst of two migrations: one to GitHub (from LF-hosted git + bugzilla) and another to XML. Neither migrations are complete yet.

The XML schema isn't finalized yet (and this repo is marked not-authoritative in its readme), so it is likely the implicit schema embodied in these files will change as the schema attribute and element names are finalized. At the same time, we are working on documentation for contributors and license requestors, which we hope to have ready once the schema is stable and all the existing licenses have been imported.

You can find an overview of the XML project here:
http://wiki.spdx.org/view/Legal_Team/Templatizing/ReviewXML

And a fairly recently updated status on the XML project (with outstanding action items) here:
http://wiki.spdx.org/view/Legal_Team/Templatizing/ActionPlan

wking:
On Wed, Feb 22, 2017 at 11:53:44AM -0800, Bradlee H. Edmndson wrote:
> The XML schema isn't finalized yet (and this repo is marked
> not-authoritative in its readme)…

Yup, non-authoritative-ness is well marked.  That doesn't mean you
need to keep the schema out of version control, and in fact, is
probably even more reason to keep the schema *in* version control ;).
I'm happy to try and assemble a pull request for the currently-landed
licenses if that would be useful.  Do you have a preference for
tooling?  RELAX NG?  XSD?

> And a fairly recently updated status on the XML project (with outstanding action items) here:
> http://wiki.spdx.org/view/Legal_Team/Templatizing/ActionPlan

How can I get involved in this?  For example, I'd rather have a
different solution for deprecated licenses (#392) than the one listed
in the action plan, but I'm not sure how/where to advocate for it.

wking:
Following the wiki lead turned up existing docs for the XML in [1]
which links the existing field notes in [2].  I recommend moving [1]
from the wiki into this repository as a first pass at resolving this
issue.  The wiki page has a browsable history [3], but the UI for that
is a lot harder to use than Git history, and wiki changes are probably
not peer-reviewed before they land (while Git commits can be reviewed
before they land).  If this sounds useful, I'm happy to provide a pull
request, with or without preserving the wiki's history, just let me
know.

[...]

Issue #314, opened 2016-05-12
Programmatically replace HTML-escaped characters?


bradleeedmondson:
Opening this issue to confirm that @myndzi plans to programmatically replace HTML-escaped characters like `&quot;`, `&apos;`, etc., on a later pass (i.e. there is no need to replace these by hand). I believe I recall correctly (though could always be mistaken) that he planned to do so at some point and waved us off modifying these manually.

If so, however, do we need to worry about straight quotes in ascii vs. UTF-8 curly quotes? Or is that all taken care of?

myndzi:
I assume curly quotes will be left curly and straight quotes will be left straight. This presents a problem for matching, but leaving them as they were in the source file helps reduce this for now. Later, they might be added to the synonym/equivalence list to allow them to match each other when building a match...

bradleeedmondson:
Note that we need to leave greater-than and less-than characters escaped, but can replace quotes, apostrophes, and other characters with their ASCII characters.

bradleeedmondson:
Brad to find example of UTF-8 character and check schemadev branch to see whether tag-conversion tool has taken care of this already and let Gary know.

silverhook:
Keep in mind that some of the curly and straight quotes seem to stem from different ways and sources of gathering the original texts in the first place.

An obvious example are GPL-3.0 and AGPL-3.0 which in both this and the _license-list_ repository use different quotation marks. I’d suspect that one was copied in plaintext format and the other was a copy-paste from the HTML website. This might also explain the difference in word wrapping between the two, otherwise near-identical licenses.

For consistancy’s sake and easier `diff`-ing, I’d prefer choosing simply one type of brackets.

bradleeedmondson:
We do have a synonyms/equivalents list already, including mostly words (Programme=Program) but also (C) = ©, so I think the plan to reduce these to ascii and then mark curly/straight quotes as synonyms should probably work well.

https://spdx.org/spdx-license-list/matching-guidelines

silverhook:
@bradleeedmondson Right, but the license texts in the SPDX list may be used also by less sophisticated tools, so reducing them to whatever makes sense, sounds great :)

goneall:
I added an issue to the SPDX tools to normalize the quotes when we product the license list data and website from the XML input format: https://github.com/spdx/tools/issues/95

I'll try to include this improvement before the next release of the license list.

goneall:
[...]

PR #499, opened 2017-11-30
Add the EPL-2.0

waynebeaton:


src/EPL-2.0.xml
```diff
@@ -0,0 +1,635 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<SPDXLicenseCollection xmlns="http://www.spdx.org/license">
+	<license isOsiApproved="true" licenseId="EPL-2.0"
+		name="Eclipse Public License 2.0">
+		<crossRefs>
+			<crossRef>http://www.eclipse.org/legal/epl-2.0</crossRef>
+			<crossRef>http://www.opensource.org/licenses/EPL-2.0</crossRef>
+		</crossRefs>
+		<notes></notes>
```
wking: If we don't have any notes on this license, I think we should leave these off ([they're optional][1]).

[1]: https://github.com/spdx/license-list-XML/blob/500c834cddce63632bb839fb5502575551753042/schema/ListedLicense.xsd#L22

src/EPL-2.0.xml
```diff
@@ -0,0 +1,635 @@
...
+							such changes and/or additions to the Program originate
+							from
+							and
+							are Distributed by that particular Contributor. A
+							Contribution
+							"originates"
+							from a Contributor if it was added to the
+							Program by such
+							Contributor itself or anyone acting on such
+							Contributor&#039;s
```
wking: We would prefer UTF-8 for the apostrophe (`'`).  See #314.

waynebeaton: I'll remove the empty note and replace the apostrophes.

In the process of converting the HTML document into the license schema, I ended up converting a blockquote into a paragraph (p). Does it instead make more sense to using a list/item combination?

Out of curiosity, why not just wrap the existing HTML in a CDATA block? Converting to the license schema seems unnecessary at best and potentially problematic (in, e.g., cases like mine when the original license text uses a blockquote).

goneall: When we convert the XML to the HTML for the spdx.org/licenses site, we add styling for the `<alt>` and `<optional>`.  This would be challenging to process if we were working from the original HTML.

The tool that generates the website does have hooks for passing in original HTML if we wanted to go that route - we would just lose the optional and alt highlighting.

wking: > Out of curiosity, why not just wrap the existing HTML in a CDATA block?

Besides the issues mentioned by @goneall, I also spent a bit of time working on XHTML embedding and couldn't get validation working.  See [here][1] and later comments in that issue mentioning “XHTML”.

> Converting to the license schema seems unnecessary at best and potentially problematic (in, e.g., cases like mine when the original license text uses a blockquote).

EPL-2.0 seems to use `<blockquote>` for [Exhibit A][2], but I'm a bit confused by its placement:

```console
$ curl -s 'https://www.eclipse.org/org/documents/epl-2.0/EPL-2.0.html' | grep -B6 -A2 '<blockquote>'
    <h2 id="exhibit-a">Exhibit A &ndash; Form of Secondary Licenses Notice</h2>
    <p>&ldquo;This Source Code may also be made available under the following 
        Secondary Licenses when the conditions for such availability set forth 
        in the Eclipse Public License, v. 2.0 are satisfied: {name license(s),
        version(s), and exceptions or additional permissions here}.&rdquo;
    </p>
    <blockquote>
      <p>Simply including a copy of this Agreement, including this Exhibit A
        is not sufficient to license the Source Code under Secondary Licenses.
```

I'd have expected:

```xml
<blockquote>
  <p>This Source Code may also be made available under the following 
    Secondary Licenses when the conditions for such availability set forth 
    in the Eclipse Public License, v. 2.0 are satisfied: {name license(s),
    version(s), and exceptions or additional permissions here}.
  </p>
</blockquote>
<p>Simply including a copy of this Agreement, including this Exhibit A…
```

Are you sure the upstream blockquote is in the right place?

[1]: https://github.com/spdx/license-list-XML/issues/391#issuecomment-282795938
[2]: https://www.eclipse.org/org/documents/epl-2.0/EPL-2.0.html#exhibit-a

src/EPL-2.0.xml
```diff
@@ -0,0 +1,635 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<SPDXLicenseCollection xmlns="http://www.spdx.org/license">
+	<license isOsiApproved="true" licenseId="EPL-2.0"
+		name="Eclipse Public License 2.0">
+		<crossRefs>
+			<crossRef>http://www.eclipse.org/legal/epl-2.0</crossRef>
+			<crossRef>http://www.opensource.org/licenses/EPL-2.0</crossRef>
```
wking: This should use `https://opensource.org/licenses/EPL-2.0`, since that's the OSI's canonical location:

```console
$ curl -sLI http://www.opensource.org/licenses/EPL-2.0 | grep '^HTTP\|Location'
HTTP/1.1 301 Moved Permanently
Location: http://opensource.org/licenses/EPL-2.0
HTTP/1.1 301 Moved Permanently
Location: https://opensource.org/licenses/EPL-2.0
HTTP/1.1 200 OK
```

I'd like to use HTTPS for eclipse.org too (https://www.eclipse.org/legal/epl-2.0/), although they allow both HTTP and HTTPS without redirects.

jmini:
Thank you.

What are the next steps in order for the license to appear on the website? https://spdx.org/licenses/

goneall:
> What are the next steps in order for the license to appear on the website?

We are working on a release of the license list (version 3.0) with a target release date before the end of the year.

I'll publish a preview on the website https://spdx.org/licenses/preview a few days prior to the actual release.  I'll send an email out to the legal distribution list when the preview site is available.

src/EPL-2.0.xml
```diff
@@ -0,0 +1,634 @@
...
+							</item>
+						</list>
+					</item>
+					<item>
+						<bullet>3.3</bullet>
+						Contributors may not remove or alter any copyright, patent,
+						trademark,
+						attribution notices, disclaimers of warranty, or
+						limitations of liability
+						("notices")
```
wking: I finally got around to comparing this with the [upstream text][1], and the only differences I see are curly/straight quotes (which I'm fine with ignoring) and then this line, which has double quotes here, but is using curly single quotes upstream:

```console
$ curl -s https://www.eclipse.org/legal/epl-2.0/ | grep '(.notices.)'
      (‘notices’) contained within the Program from any copy of the Program which
```

@waynebeaton, do you know if upstream intended to use single quotes there?  Or is the intention to use double (curly) quotes?

[1]: https://www.eclipse.org/legal/epl-2.0/

waynebeaton: As it was delivered to me, the text ony version uses straight double quotes, and the HTML version uses single curly quotes. That must have happened in translation. 

My vote is to ignore.

jlovejoy: curly v. straight quote doesn't matter for matching purposes: see 5.1.3 https://spdx.org/spdx-license-list/matching-guidelines

wking: > curly v. straight quote doesn't matter for matching purposes…

I agree.  I just wanted to make sure we represent the upstream preference for single vs. double (even though that doesn't impact matching, it's still nice to be consistent).  But as @waynebeaton points out, upstream uses double in their text/plain version and single in their text/html version, so (without a consistent upstream pattern to follow) it doesn't really matter which version we use here.

PR #512, opened 2017-12-12
Create Bootloader-exception

jlovejoy:
add file for Bootloader-exception

### DIFF
--- a/src/exceptions/Bootloader-exception
+++ b/src/exceptions/Bootloader-exception
@@ -1,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<SPDXLicenseCollection xmlns="http://www.spdx.org/license">
+   <exception licenseId="Bootloader-exception" name="Bootloader Distribution Exception">
+      <crossRefs>
+         <crossRef>https://github.com/pyinstaller/pyinstaller/blob/develop/COPYING.txt</crossRef>
+      </crossRefs>
+  
+      <titleText>Bootloader Exception</titleText>
+<optional>--------------------</optional>
+<p>In addition to the permissions in the GNU General Public License, the
+authors give you unlimited permission to link or embed compiled bootloader
+and related files into combinations with other programs, and to distribute
+those combinations without any restriction coming from the use of those
+files. (The General Public License restrictions do apply in other respects;
+for example, they cover modification of the files, and distribution when
+not linked into a combine executable.)</p>
+  </exception>
+</SPDXLicenseCollection>


jlovejoy:
@goneall  or @zvr - could one of you review this and merge? (I'm not sure I have the line spacing/characters correct)

src/exceptions/Bootloader-exception
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<SPDXLicenseCollection xmlns="http://www.spdx.org/license">
+   <exception licenseId="Bootloader-exception" name="Bootloader Distribution Exception">
+      <crossRefs>
+         <crossRef>https://github.com/pyinstaller/pyinstaller/blob/develop/COPYING.txt</crossRef>
+      </crossRefs>
+  
+      <titleText>Bootloader Exception</titleText>
+<optional>--------------------</optional>
```
wking: Can we follow the [Multics precedent][1] here and allow different numbers of hyphens?  Previous discussion [here][2].

And this underline is really title markup, so:

* I'd prefer it to go inside `<titleText>`.
* I'd rather not include it in the canonical HTML at all.

Putting all of those suggestions together gives something like:

```xml
<titleText>
  <p>Bootloader Exception<alt match="-*" name="titleUnderline"/></p>
</titleText>
```

We don't currently distinguish between “canonical HTML content” (which should not have the underline) and “canonical text/plain content” (which maybe should have the underline), but I think not having the underline in text/plain generated from our XML is a lesser evil than having underscores in HTML generated from our XML.

[1]: https://github.com/spdx/license-list-XML/blob/52f17985f621530754add25316605748d9f19c82/src/Multics.xml#L33-L35
[2]: https://github.com/spdx/license-list-XML/pull/474#discussion_r148378189

src/exceptions/Bootloader-exception
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<SPDXLicenseCollection xmlns="http://www.spdx.org/license">
+   <exception licenseId="Bootloader-exception" name="Bootloader Distribution Exception">
+      <crossRefs>
+         <crossRef>https://github.com/pyinstaller/pyinstaller/blob/develop/COPYING.txt</crossRef>
```
wking: Do we have a policy on pinned vs. floating links for these references?  See discussion in github/choosealicense.com#555 for more on this, but if we only use a single link, I'd rather pin it to a specific version (like [this][1]) unless we are automatically scanning the floating target and confirming that it does still include the expected material (in this case, the exception text).

[1]: https://github.com/pyinstaller/pyinstaller/blob/v3.3/COPYING.txt#L16-L25

src/exceptions/Bootloader-exception
```diff
@@ -0,0 +1,18 @@
...
+  
+      <titleText>Bootloader Exception</titleText>
+<optional>--------------------</optional>
+<p>In addition to the permissions in the GNU General Public License, the
+authors give you unlimited permission to link or embed compiled bootloader
+and related files into combinations with other programs, and to distribute
+those combinations without any restriction coming from the use of those
+files. (The General Public License restrictions do apply in other respects;
+for example, they cover modification of the files, and distribution when
+not linked into a combine executable.)</p>
```
wking: [Upstream][1] and [the GCC exception 2.0][2] both have this typo too, but I'm pretty sure “combine executable” should be “combined executable” (compare “[combined work][3]”).  If folks agree, I think we want to mark that up with:

```
not linked into a <alt name="combined" match="combined|combine">combined</alt> executable.)</p>
```

[1]: https://github.com/pyinstaller/pyinstaller/blob/v3.3/COPYING.txt#L25
[2]: https://github.com/spdx/license-list-XML/blob/master/src/exceptions/GCC-exception-2.0.xml#L18
[3]: https://github.com/spdx/license-list-XML/blob/master/src/GPL-3.0.xml#L498

src/exceptions/Bootloader-exception
jlovejoy: no, we don't have a policy and it's usually (always?) been a regular link. I don't care either way and not sure we need an official policy. In this case, the text is at the top/easy to find. A pinned link could be useful if the file was long and the specific text hard to find. let's just leave this as is.

src/exceptions/Bootloader-exception
jlovejoy: fine. let's make consistent with Multics then.

src/exceptions/Bootloader-exception
jlovejoy: we usually don't "fix" typos especially when they appear consistently, but in the chance that someone else has fixed it downstream, I'm fine with adding this. (can you do so?)

jlovejoy: that's fine, please add :)

src/exceptions/Bootloader-exception
wking: > A pinned link could be useful if the file was long and the specific text hard to find.

Linking to a specific part of the file is convenient, and probably only a good idea for pinned links.  But I'm more concerned about “what happens after pyinstaller changes their license and drops the exception or moves the exception into a different file?”.  Projects don't relicense or rename their license files often, but it does happen.

src/exceptions/Bootloader-exception
wking: > … I'm fine with adding this. (can you do so?)

I don't have push access to this repo, so I can't update your branch directly.  Instead, I've filed #515 against your branch; merging #515 will update this PR.

wking:
Can someone with push access delete the `jlovejoy-patch-1` branch now that this has landed?

PR #519, opened 2017-12-16
.travis.yml: Add gulp validation

### LINKED ISSUE #517: CDLA-Sharing-1.0 Parsing Error
Following is an XML schema validation error reported when generating the license list data:

Parsing error in XML file CDLA-Sharing-1.0.xml at line 170, column 23:cvc-complex-type.2.4.a: Invalid content was found starting with element 'bullet'. One of '{"http://www.spdx.org/license":item, "http://www.spdx.org/license":list}' is expected., Skipping file CDLA-Sharing-1.0.xml

### LINKED ISSUE #518: Missing required name attribute for ISC
The following error was reported when generating the license list data:

Missing name attribute for variable text, Skipping file ISC.xml

wking:
Details on the individual pivots in the commit messages, but this pull requests gets automatic schema validation to avoid issues like #517 ~~(which it fixes)~~.

The tests aren't as extensive as those in #513.  I think we want that sort of XML-template vs. known good/bad sample testing soon, but filling in gulp-based tests seems like something we want as well, and it's easier to implement.

Fixes #518.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,8 @@
 .DS_Store
-# Ignore all hidden files/dirs except .gitignore
+# Ignore all hidden files/dirs except .gitignore and .travis.yml
 .*
 !/.gitignore
+!/.travis.yml
 
 # nodeJS
 node_modules/
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,5 @@
+language: node_js
+node_js:
+  - "node"
+before_script:
+  - npm run preinstall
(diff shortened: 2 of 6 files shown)


wking:
And I've updated the schema with [`use="required"`][1] to catch issues like #518, and also added a commit to fix #518.

[1]: https://www.w3.org/TR/xmlschema11-1/#declare-attribute

wking:
I've dropped the CDLA-Sharing-1.0 patch with 4cb164a → 3ddcfc9, now that the `<item>` approach is in #521.

goneall:
I'm going to merge this so that I can run the license tests again.  I'll submit a separate PR for issue #520

PR #521, opened 2017-12-17
fixed CDLA-Sharing-1.0

zvr:
added missing <item>s in a <list>
https://github.com/spdx/license-list-XML/issues/517

### DIFF
--- a/src/CDLA-Sharing-1.0.xml
+++ b/src/CDLA-Sharing-1.0.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <SPDXLicenseCollection xmlns="http://www.spdx.org/license">
   <license licenseId="CDLA-Sharing-1.0"
-   name="Community Data License Agreement Sharing 1.0">
+  name="Community Data License Agreement Sharing 1.0">
     <crossRefs>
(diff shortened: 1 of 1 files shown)


src/CDLA-Sharing-1.0.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <SPDXLicenseCollection xmlns="http://www.spdx.org/license">
   <license licenseId="CDLA-Sharing-1.0"
-   name="Community Data License Agreement Sharing 1.0">
+  name="Community Data License Agreement Sharing 1.0">
```
wking: This dedent looks odd to me.  I'd rather this be on the same line as `<license …` or have at least a space of additional input.  For what it's worth, lxml (as scripted in #432) is taking the single-line approach here:

```xml
<license licenseId="CDLA-Sharing-1.0" name="Community Data License Agreement Sharing 1.0">
```

zvr: How it works: Indent levels increase (in this run by 2 spaces but it's a command-line option) whenever you enter a new XML block element. Lines are split and wrapped to max characters (80, also a command-line option) with same initial indent (Knuth and Plass algorithm, optimized).
As long as the start tag (license in this case) has not yet closed, indent remains same.

Issue #506, opened 2017-12-07
Add BSD 1 Clause License


bradleeedmondson:
@zvr volunteered to create initial PR for BSD 1-clause

I can review

zvr:
Added it in #522

waldyrious:
Cross-referencing https://github.com/spdx/license-list/issues/13 (and also https://github.com/github/choosealicense.com/issues/464 for related context and background)

PR #522, opened 2017-12-17
Adding BSD-1-Clause

zvr:
Taking care of #506 

I have used "BSD-1-Clause" for identifier and "BSD 1-clause License" for name.
We haven't always been consistent on capitalization of "clause" in names.

I have also templatized the license, matching the other BSD ones.

### DIFF
--- a/src/BSD-1-Clause.xml
+++ b/src/BSD-1-Clause.xml
@@ -1,0 +1,50 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<SPDXLicenseCollection xmlns="http://www.spdx.org/license">
+   <license licenseId="BSD-1-Clause"
+            name="BSD 1-clause License">
+      <crossRefs>
+	      <crossRef>
+		      https://svnweb.freebsd.org/base/head/include/ifaddrs.h?revision=326823
+	      </crossRef>
+      </crossRefs>
+      <copyrightText>
+         <p>Copyright (c) 
+        <alt match=".+" name="copyright"> &lt;year&gt; &lt;owner&gt;</alt> All rights reserved. 
+      </p>
+      </copyrightText>
+    
+      <p>
+ Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions
+are met:
+      </p>
+      <list>
+        <item>
+            <bullet>1.</bullet>
+Redistributions of source code must retain the above copyright
+   notice, this list of conditions and the following disclaimer.
+        </item>
+      </list>
+
+      <p>
+      THIS SOFTWARE IS PROVIDED BY
+        <alt match=".+" name="copyrightHolderAsIs">Berkeley Software Design, Inc.</alt>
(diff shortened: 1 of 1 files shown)


src/BSD-1-Clause.xml
```diff
@@ -0,0 +1,42 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<SPDXLicenseCollection xmlns="http://www.spdx.org/license">
+  <license licenseId="BSD-1-Clause" name="BSD 1-clause License">
```
wking: We don't seem to be particularly consistent about this:

```
$ git grep 'name="BSD .*-clause' | wc -l
6
$ git grep 'name="BSD .*-Clause' | wc -l
3
```

but despite its current lower popularity, I prefer `1-Clause`.  But I'm fine leaving this alone until we have an official policy (I've opened #524 to get that policy).

src/BSD-1-Clause.xml
```diff
@@ -0,0 +1,42 @@
...
+    <crossRefs>
+      <crossRef>https://svnweb.freebsd.org/base/head/include/ifaddrs.h?revision=326823</crossRef>
+    </crossRefs>
+    <copyrightText>
+      <p>
+        Copyright (c)<alt name="copyright" match=".+"><year> <owner></alt>
+        All rights reserved.
```
wking: It's a bit strange that “All” looks like the start of a new sentence, but the previous line doesn't end in a period or other terminal punctuation.  Looking [upstream][1], the copyright owner ends with “Inc.”, so I think we want:

```xml
Copyright (c)<alt name="copyright" match=".+"><year> <owner></alt><optional>.</optional>
```

to cover cases where the owner ends with a period (and you don't want to double up) as well as cases where the owner does not (so you need some additional terminal punctuation).
[1]: https://svnweb.freebsd.org/base/head/include/ifaddrs.h?revision=250887&view=markup

src/BSD-1-Clause.xml
```diff
@@ -0,0 +1,42 @@
...
+    </p>
+    <list>
+      <item>
+        <bullet>1.</bullet>
+        Redistributions of source code must retain the above copyright
+        notice, this list of conditions and the following disclaimer.
+      </item>
+    </list>
+    <p>
+      THIS SOFTWARE IS PROVIDED BY<alt name="copyrightHolderAsIs"
```
wking: You need a space here: `BY<alt` → `BY <alt`.

src/BSD-1-Clause.xml
```diff
@@ -0,0 +1,42 @@
...
+      match=".+">Berkeley Software Design, Inc.</alt>
```
wking: I think it's odd to template `<owner>` in the Copyright line, but not template it here.  And as long as we're templating it, it would be nice to use `THE COPYRIGHT HOLDERS AND CONTRIBUTORS` (like BSD-2-Clause) or some other generic text as the default here.

PR #534, opened 2017-12-21
Create LLVM-exception

jlovejoy:
adding LLVM exception to Apache-2.0
need to check with LLVM folks on name (and update URL later)


wking:
With this PR merged, can someone with write access delete branch `jlovejoy-patch-1`?

Issue #543, opened 2017-12-21
deprecate GNU licenses with current identifiers


jlovejoy:
GPL-1.0
GPL-2.0
GPL-3.0

LGPL-2.0
LGPL-2.1
LGPL-3.0

FDL-1.1
FDL-1.2
FDL-1.3

AGPL-3.0

goneall:
The notes for the recently added [L]GPL-N.N+.xml files need to be updated to reference the new ID's.

zvr:
The FDL ones are actually GFDL ;)

zvr:
Submitted #553

zvr:
@goneall I am not sure I understood your comment... Which recently added files, which notes and which IDs?

goneall:
> Which recently added files, which notes and which IDs?

The LGPL-2.1+, GPL-2.0+, etc.

jlovejoy:
include AGPL-1.0 and then close this

PR #545, opened 2017-12-21
deprecated GPL-1.0.xml

jlovejoy:
added isDeprecated tag and note

### DIFF
--- a/src/GPL-1.0.xml
+++ b/src/GPL-1.0.xml
@@ -1,10 +1,11 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <SPDXLicenseCollection xmlns="http://www.spdx.org/license">
-   <license isOsiApproved="false" licenseId="GPL-1.0"
+   <license isDeprecated="true" isOsiApproved="false" licenseId="GPL-1.0"
             name="GNU General Public License v1.0 only">
       <crossRefs>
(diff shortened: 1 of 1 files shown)


jlovejoy:
I did this one as a pull request to make sure it's right and the notes text is okay (so we can then do same for all others as per https://github.com/spdx/license-list-XML/issues/543

still need to notate version for which it was deprecated (v3.0)
@kestewart can you review text in note?

src/GPL-1.0.xml
```diff
@@ -1,10 +1,11 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <SPDXLicenseCollection xmlns="http://www.spdx.org/license">
-   <license isOsiApproved="false" licenseId="GPL-1.0"
+   <license isDeprecated="true" isOsiApproved="false" licenseId="GPL-1.0"
             name="GNU General Public License v1.0 only">
       <crossRefs>
          <crossRef>http://www.gnu.org/licenses/old-licenses/gpl-1.0-standalone.html</crossRef>
       </crossRefs>
+      <notes>DEPRECATED: Deprecated in lieu of more explicit license identifier of GPL-1.0-only
```
goneall: Missing </notes> at the end of this line

goneall: Missing </notes>

jlovejoy: added (and deprecatedVersion tag)

PR #551, opened 2017-12-22
Add OpenJDK Assembly exception

waynebeaton:
At least two Eclipse projects use the Open JDK Assembly exception. I'd like to add it to the exception list.

I've labelled it 2.0 because it is an exception specifically to the GPL-2.0. I am hopeful that this was the right thing to do.

I'll address this to the SPDX legal mailing list.

### DIFF
--- a/src/exceptions/OpenJDK-Assembly-exception-2.0.xml
+++ b/src/exceptions/OpenJDK-Assembly-exception-2.0.xml
@@ -1,0 +1,50 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<SPDXLicenseCollection xmlns="http://www.spdx.org/license">
+   <exception licenseId="OpenJDK-assembly-exception-2.0" name="Open JDK Assembly exception 2.0">
+      <crossRefs>
+         <crossRef>http://openjdk.java.net/legal/assembly-exception.html
+         </crossRef>
+      </crossRefs>
+      <titleText>OpenJDK Assembly Exception</titleText>
+      <p>
+         The OpenJDK source code made available by Oracle America, Inc.
+         (Oracle) at openjdk.java.net ("OpenJDK Code") is distributed
+         under the terms of the GNU General Public License
+         &lt;http://www.gnu.org/copyleft/gpl.html&gt; version 2 only
+         ("GPL2"), with the following clarification and special
+         exception.
+      </p>
+
+      <list>
+         <item>
+            <p>Linking this OpenJDK Code statically or dynamically with
+               other code is making a combined work based on this
+               library. Thus, the terms and conditions of GPL2 cover the
+               whole combination.
+            </p>
+
+            <p>As a special exception, Oracle gives you permission to
+               link this OpenJDK Code with certain code licensed by
+               Oracle as indicated at
+               http://openjdk.java.net/legal/exception-modules-2007-05-08.html
+               ("Designated Exception Modules") to produce an
+               executable, regardless of the license terms of the
+               Designated Exception Modules, and to copy and distribute
+               the resulting executable under GPL2, provided that the
+               Designated Exception Modules continue to be governed by
+               the licenses under which they were offered by Oracle.
+            </p>
+         </item>
+      </list>
+
+      <p>As such, it allows licensees and sublicensees of Oracle's GPL2
+         OpenJDK Code to build an executable that includes those
+         portions of necessary code that Oracle could not provide under
+         GPL2 (or that Oracle has provided under GPL2 with the Classpath
+         exception). If you modify or add to the OpenJDK code, that new
+         GPL2 code may still be combined with Designated Exception
+         Modules if the new code is made subject to this exception by
+         its copyright holder.
+      </p>
+   </exception>
+</SPDXLicenseCollection>


src/exceptions/OpenJDK-Assembly-exception-2.0.xml
```diff
@@ -0,0 +1,50 @@
...
+         (Oracle) at openjdk.java.net ("OpenJDK Code") is distributed
+         under the terms of the GNU General Public License
+         &lt;http://www.gnu.org/copyleft/gpl.html&gt; version 2 only
+         ("GPL2"), with the following clarification and special
+         exception.
+      </p>
+
+      <list>
+         <item>
+            <p>Linking this OpenJDK Code statically or dynamically with
```
wking: I haven't looked at our existing exceptions for precedent here, but to me it feels like the text you've included in the `<list>` element is the legal text of the exception.  The earlier “The OpenJDK source code made available…” paragraph sounds like part of a grant that associates the exception with a particular set of code (much like a GPL header at the beginning of a file associates the GPL license with the code in the file).  The final “As such, it allows licensees and sublicensees…” seems like a combination non-normative notes and a contributor license agreement.

jlovejoy:
@waynebeaton as for the numbering - I just had a review of all our existing exceptions: for the most part they have a number if the exception itself is versioned, or if there was more than one version of the exception and a number was needed to differentiate (even if the exception text didn't necessarily denote this). There are a couple that don't completely follow this pattern. 
In this case, it doesn't seem like it really needs a number - there is only one version of this exception out there, is that correct?

wking:
On Tue, Dec 26, 2017 at 01:41:25PM -0800, Jilayne Lovejoy wrote:
> In this case, it doesn't seem like it really needs a number - there
> is only one version of this exception out there, is that correct?

I am in favor of versioning it somehow.  If you want to start out with
a 2.0 verison because it's designed for GPL-2.0-only, that's fine with
me.  If you decide to release a new exception also targetting
GPL-2.0-only, you'd have OpenJDK-Assembly-exception-2.1 or whatever,
and you can release an exception that targets the GPL-3.0, etc., etc.

I'm also ok with calling this OpenJDK-Assembly-exception-1.0 (for “the
first release of this exception”).

I'd rather avoid our current W3C situation (with W3C, W3C-19980720,
and W3C-20150513) by assuming “we'll never cut another version of
this” and skipping a version number.

waynebeaton:
It does seem intuitive that having some sort of version number will help avoid confusion later.

Version 2.0 isn't confusing when it matches against GPL-2.0, but will become confusing when/if the exception is updated. e.g. what does it mean to have version 2.1 of the exception matched against the GPL-2.0?

I'm thinking that the version number should apply to the exception itself, and that we should make it 1.0.

src/exceptions/OpenJDK-Assembly-exception-2.0.xml
waynebeaton: The Eclipse OMR and Eclipse OpenJ9 projects both use the OpenJDK assembly exception. I'm not happy with their LICENSE files and will get them to update them.

e.g.

https://github.com/eclipse/openj9/blob/master/jcl/src/openj9.dataaccess/share/classes/com/ibm/dataaccess/ByteArrayMarshaller.java

wking: > https://github.com/eclipse/openj9/blob/master/jcl/src/openj9.dataaccess/share/classes/com/ibm/dataaccess/ByteArrayMarshaller.java

The Eclipse OpenJ9 project seems to cover the grant association [in their per-file header][1], which seems like it make's the “The OpenJDK source code made available…” paragraph superfluous for that file.  And they cover the contributor license agreement section with [an explicit CLA and sign-off requirement][2].  So at least some parts of the [upstream exception page][3] seem redundant/misleading in that context.  For OpenJ9 (which is not available under openjdk.java.net?), the exception seems to allow folks to link OpenJ9 against the [designated exception modules][4], but it's the copyright holder (not always Oracle) granting the exception.  A generic form of the exception would be closer to:

> Linking this OpenJDK Code statically or dynamically with other code is making a combined work based on this library.  Thus, the terms and conditions of the GNU General Public License <http://www.gnu.org/copyleft/gpl.html> version 2 only ("GPL2") cover the whole combination.
>
> As a special exception, the copyright holders and contributors give you permission to link this OpenJDK Code with certain code as indicated at http://openjdk.java.net/legal/exception-modules-2007-05-08.html ("Designated Exception Modules") to produce an executable, regardless of the license terms of the Designated Exception Modules, and to copy and distribute the resulting executable under GPL2, provided that the Designated Exception Modules continue to be governed by the licenses under which they were offered by Oracle.

[1]: https://github.com/eclipse/openj9/blob/8fc87280857fe06e0d91409abce26e751c1dc6ae/jcl/src/openj9.dataaccess/share/classes/com/ibm/dataaccess/ByteArrayMarshaller.java#L15-L16
[2]: https://github.com/eclipse/openj9/blob/8fc87280857fe06e0d91409abce26e751c1dc6ae/CONTRIBUTING.md#legal-considerations
[3]: http://openjdk.java.net/legal/assembly-exception.html
[4]: http://openjdk.java.net/legal/exception-modules-2007-05-08.html

wking:
This PR should probably be tagged as a “new license/exception request” by someone with write access to this repository.

src/exceptions/OpenJDK-Assembly-exception-2.0.xml
jlovejoy: @waynebeaton  - @wking  has a good point: considering the inbound policy is DCO sign-off (not sure what the need for the ECA is on top of that, but that's another story), all contributors retain their copyright. Thus the exception shouldn't be specific to Oracle. I don't know the history here is and maybe it was appropriate at one time to only come from Oracle if they were the only copyright holder (I believe, they use an assignment agreement for contributions to their project, so this could have worked under that paradigm). But given the Eclipse Foundation licensing policy this exception may need to be updated.

wking:
This PR should probably be milestoned 3.1 to [match][1] #567 (it's currently milestoned “Immediate Release - 3.0”), although:

* I'm not clear on why we need a separate issue at all.  We could always open a new issue if/when someone decides to carry this PR, but as far as I can tell the current plan is to work on this PR until it's ready.

* I'm not clear on [why we're using 3.0 for the next release][2].  I'd be fine with this landing in 3.0 if that happened to be a later release.

[1]: https://github.com/spdx/license-list-XML/issues/567#issuecomment-354178445
[2]: https://lists.spdx.org/pipermail/spdx-legal/2017-December/002437.html

src/exceptions/OpenJDK-Assembly-exception-2.0.xml
```diff
@@ -0,0 +1,50 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<SPDXLicenseCollection xmlns="http://www.spdx.org/license">
+   <exception licenseId="OpenJDK-assembly-exception-1.0" name="OpenJDK Assembly exception 1.0">
+      <crossRefs>
+         <crossRef>http://openjdk.java.net/legal/assembly-exception.html
+         </crossRef>
+      </crossRefs>
+      <titleText>OpenJDK Assembly Exception</titleText>
```
wking: This is `<titleText>` outside of `<text>`, which is why you're getting “[This element is not expected][1]”.  You need to add wrapping `<text>` like 4af440270 did for `LLVM-exception`.

[1]: https://travis-ci.org/spdx/license-list-XML/jobs/357048314#L653

src/exceptions/OpenJDK-Assembly-exception-2.0.xml
```diff
@@ -0,0 +1,52 @@
...
+         <crossRef>http://openjdk.java.net/legal/assembly-exception.html
+         </crossRef>
+      </crossRefs>
+      <text>
+      <titleText>OpenJDK Assembly Exception</titleText>  
+      <p>
+         The OpenJDK source code made available by Oracle America, Inc.
+         (Oracle) at openjdk.java.net ("OpenJDK Code") is distributed
+         under the terms of the GNU General Public License
+         &lt;http://www.gnu.org/copyleft/gpl.html&gt; version 2 only
```
wking: Perhaps this link should be to (or at least allow) https://www.gnu.org/licenses/old-licenses/gpl-2.0.html?  Currently https://www.gnu.org/copyleft/gpl.html has the v3 text, and you have to click through “Old versions of the GNU GPL” and “GNU General Public License, version 2” to get from there to the v2 text.

wking: Also allowing `https` as well would be nice (like we do in the GPL licenses themselves since #450).

wking: > I like allowing http|https matches (let's encrypt! and all that), but that's broader than just this license. Let's open a separate issue for that.

With this sort of thing, I think the easiest approach is to gate-keep new submissions to hold the new standards while slowly chipping away at the existing instances.  Landing new code that violates our intended approach just makes the existing issues more trouble to fix ;).  This lets you solve the issues incrementally without having huge-change flag-days.

bradleeedmondson: I can see the case for that, but we don't want to hold up substantive additions to the license list while deciding. Also, if we want to systematically change this, there wouldn't appear to be much effort difference in (N licenses affected) vs. (N-1 licenses affected).

Set up in #633 to address systematically later.

Issue #544, opened 2017-12-21
missing some deprecated licenses


jlovejoy:
I was checking the list of deprecated licenses against what's in the src files here and it appears that these are missing:

GPL-1.0+
GPL-2.0+
GPL-3.0+
GPL-3.0-with-GCC-exception
LGPL-2.1+
LGPL-3.0+
LGPL-2.0+
StandardML-NJ
WXwindows

@goneall  i think you (or Kris) had added these? Am I missing something?

goneall:
Yes - these should have been added.  @jlovejoy glad you checked this.  I'll add PR's for the missing deprecated licenses.

jlovejoy:
GPL-1.0+
~~GPL-2.0+~~
~~GPL-3.0+~~
~~GPL-3.0-with-GCC-exception~~
~~LGPL-2.1+~~
~~LGPL-3.0+~~
~~LGPL-2.0+~~
~~StandardML-NJ~~
~~WXwindows~~

Merged commits for 8 files, but looks like GPL-1.0+ was missing?

goneall:
Adding a PR for GPL-1.0+

PR #552, opened 2017-12-22
Add remaining deprecated licenses. 

goneall:
Note: the notes for the GPL licenses need to be updated once the new GPL license ID have been added.

Resolves issue #544 


goneall:
> Note: the notes for the GPL licenses need to be updated once the new GPL license ID have been added.

These are now updated.

src/GPL-2.0+.xml
```diff
@@ -0,0 +1,311 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<SPDXLicenseCollection xmlns="http://www.spdx.org/license">
+   <license isOsiApproved="true" licenseId="GPL-2.0+" isDeprecated="true" deprecatedVersion="2.0rc2"
```
wking: I'd rather not use RC versions in `deprecatedVersion`, since those are a lower tier than non-RC versions.  For example, if a license was marked as deprecated in 2.0rc2 but we changed our minds and removed the deprecation by 2.0, I'd consider the license “never deprecated”.  If that argument makes sense, I think we want to use `deprecatedVersion="2.0"` for the licenses where you currently have `2.0rc2`.

jlovejoy: 2.0rc2 was an actual version/release. I can't remember the details of all that now, but we can rewrite history, so this needs to be left as is. :)

wking: > 2.0rc2 was an actual version/release. I can't remember the details of all that now, but we can rewrite history, so this needs to be left as is. :)

I think you meant “can't”.  I'm not contesting the existence of the rc2 tag, I just think that tag was a pre-release (i.e. release candidate) and not a release.  [From SemVer][1] (which we don't formally follow, but which serves as a good example of conventional use):

> Pre-release versions have a lower precedence than the associated normal version. A pre-release version indicates that the version is unstable and might not satisfy the intended compatibility requirements as denoted by its associated normal version.

Are you aware of any consumers (e.g. the SPDX spec) that reference the 2.0-rc2 license list?  I'd have expected consumers to wait and reference the normal 2.0 release.

[1]: https://semver.org/spec/v2.0.0.html#spec-item-9

wking:
With this PR merged, can someone with write access delete branch `adddeprecated`?

PR #555, opened 2017-12-26
Giftware|SMLNJ: Update URLs to avoid double-slashes

wking:
I turned up these two double-slash URLs while looking for more issues like the one fixed by #540.


goneall:
Seems straightforward and only involves cross reference links.  Going ahead and merging the PR.

PR #130, opened 2016-03-30
Automated PR for MS-RL

myndzi:

PR #558, opened 2017-12-27
Allow matching spaces between '-' and works around SPDX tools issue #130

goneall:

### DIFF
--- a/src/RPL-1.5.xml
+++ b/src/RPL-1.5.xml
@@ -167,7 +167,7 @@
         <item>
             <bullet>3.0</bullet>
           Grant of License From Licensor. Subject to the terms and conditions of this License, Licensor
-             hereby grants You a world-wide, royalty-free, non- exclusive license, subject to Licensor's
+             hereby grants You a world-wide, royalty-free, <alt match="non ?- ?exclusive" name="anotherExtraSpace">non-exclusive</alt> license, subject to Licensor's
              intellectual property rights, and any third party intellectual property claims derived from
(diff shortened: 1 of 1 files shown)


src/RPL-1.5.xml
```diff
@@ -200,7 +200,7 @@
         <item>
             <bullet>4.0</bullet>
           Grant of License From Contributor. By application of the provisions in Section 6 below, each
-             Contributor hereby grants You a world-wide, royalty-<alt match=" ?" name="extraSpace"/>free, non-exclusive license, subject to
+             Contributor hereby grants You a world-wide, <alt match="royalty ?- ?free" name="extraSpace">royalty-free</alt>, non-exclusive license, subject to
```
wking: I like these better with the shorter `<alt>`, since I expect the marked-up HTML will be easier to read with a single character of variable text.

And even with the longer `<alt>`, I see no reason to allow a space before the hyphen.

goneall: > I like these better with the shorter <alt>, since I expect the marked-up HTML will be easier to read with a single character of variable text.

Me to, but the tools will not currently process this correctly.

> I see no reason to allow a space before the hyphen.

Removed

goneall:
Just a note that without this fix, the original license with the line breaks after the hyphen ("-") will match, but the more common license text without the line breaks will not match.

I'm going to go ahead and merge this since it is more of a technical fix / workaround.

wking:
With this PR merged, can someone with write access delete branch `fixrpl15`?

PR #562, opened 2017-12-27
Add missing word to Bootloader-exception.

sam-ellis:
Added missing word as compared to original license at https://github.com/pyinstaller/pyinstaller/blob/develop/COPYING.txt.


jlovejoy:
good catch, thanks!

PR #565, opened 2017-12-27
*: Fix missing space between (c) and <alt>

wking:
Generated with:

    $ sed -i 's/\(([cC])\)<alt/\1 <alt/' $(git grep -l '[cC])<alt')

Like 7af2b40 (#526), but with more licenses.  The ones I'm fixing here were mostly broken by a0879e6e (#553).

### DIFF
--- a/src/AGPL-3.0-only.xml
+++ b/src/AGPL-3.0-only.xml
@@ -7,7 +7,7 @@
       <crossRef>http://www.opensource.org/licenses/AGPL-3.0</crossRef>
     </crossRefs>
     <standardLicenseHeader>
-      Copyright (C)<alt name="copyright" match=".+">[year] [name of author]</alt>
+      Copyright (C) <alt name="copyright" match=".+">[year] [name of author]</alt>
       <p>
 	This program is free software: you can redistribute it and/or modify it
         under the terms of the GNU Affero General Public License as published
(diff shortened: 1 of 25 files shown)


wking:
CC @zvr about fixing his pretty-printer before we get more of these ;).

jlovejoy:
some of these have already been fixed and seems to be viewing okay as per Preview check

wking:
On Thu, Dec 28, 2017 at 11:16:56PM +0000, Jilayne Lovejoy wrote:
> some of these have already been fixed and seems to be viewing okay as per Preview check

https://spdx.org/licenses/preview/GPL-1.0 doesn't even render in preview:

  XML Parsing Error: mismatched tag. Expected: </br>.
  Location: https://spdx.org/licenses/preview/GPL-1.0
  Line Number 176, Column 9:      </p>
  --------^

Looks like an unclosed <br>.

wking:
Ah, I was missing the .html.  Looking at that:

  $ curl -s https://spdx.org/licenses/preview/GPL-1.0.html | grep -B1 '<var.*19xx'
        Copyright (C)
  <var class="replacable-license-text">19xx name of author</var>

the tool seems to be injecting a newline before the <var> (which fixes
this case, but introduces an unwanted space in other places.  For
example:

  $ curl -s https://spdx.org/licenses/preview/CC-BY-4.0.html | grep -B1 'wiki.creativecommons.org/Considerations_for_licensees'
  <var class="replacable-license-text">More considerations</var> for the public
  <var class="replacable-license-text">: wiki.creativecommons.org/Considerations_for_licensees</var></p>

should not have whitespace before the colon.

wking:
Rebased onto master with 62ad7e2 → 5d907de.

jlovejoy:
@goneall or @zvr can one of your give this last check and merge? looks good to go to me!

PR #566, opened 2017-12-27
*: Collapse to a single note entry

wking:
The XSD schema is currently buggy on this point, but the plural element name (`<notes>` vs. `<note>`) and `maxOccurs` in the XSD entry:

```xml
<element name="notes" type="tns:formattedFixedTextType" minOccurs="0" maxOccurs="1"/>
```

suggest that we expect only a single note entry (see also discussion [here][1]).  This PR brings us back to a single `<notes>` entry for our licenses, preparing us for #452, which fixes the XSD schema.

The wxWindows change, I created paragraphs for the previously separate `<notes>` content.

For the GPL-family changes, I dropped the release-date/version `<notes>`, because:

* Licenses with a release date in their `<notes>` entry had that same release date in the license title.  I don't see a need to include that release date as unstrucutured information in two places, and the license title is clearly the more important location.

* Whether the license is only or or-later is covered explicitly in the long name and implicitly in the new deprecation notice (e.g. as added in #552).  Once we get explicit obsoleted-by markup (#392), those deprecation notices will become machine readable.  So I don't think we need to call out the versioning again in the notes.

[1]: https://github.com/spdx/license-list-XML/pull/553#discussion_r158600187

### DIFF
--- a/src/AGPL-3.0.xml
+++ b/src/AGPL-3.0.xml
@@ -23,9 +23,6 @@
       along with this program. If not, see
       &lt;http<optional>s</optional>://www.gnu.org/licenses/&gt;
     </standardLicenseHeader>
-    <notes>
-      This version was released: 19 November 2007
-    </notes>
     <titleText>
       <p>
         GNU AFFERO GENERAL PUBLIC LICENSE<br></br>
--- a/src/GFDL-1.1.xml
+++ b/src/GFDL-1.1.xml
@@ -20,9 +20,6 @@
       Back-Cover Texts being LIST. A copy of the license is included
       in the section entitled "GNU Free Documentation License".
     </standardLicenseHeader>
-    <notes>
-      This license was released March 2000
-    </notes>
     <titleText>
       <p>
         GNU Free Documentation License<br></br>
--- a/src/GFDL-1.2.xml
+++ b/src/GFDL-1.2.xml
@@ -19,9 +19,6 @@
       and no Back-Cover Texts. A copy of the license is included
       in the section entitled "GNU Free Documentation License".
     </standardLicenseHeader>
-    <notes>
-      This license was released November 2002
-    </notes>
     <titleText>
       <p>
         GNU Free Documentation License<br></br>
--- a/src/GFDL-1.3.xml
+++ b/src/GFDL-1.3.xml
@@ -19,9 +19,6 @@
       and no Back-Cover Texts. A copy of the license is included
       in the section entitled "GNU Free Documentation License".
     </standardLicenseHeader>
-    <notes>
-      This license was released 3 November 2008.
-    </notes>
     <titleText>
       <p>
         GNU Free Documentation License<br></br>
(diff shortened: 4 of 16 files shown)


wking:
This also addresses (by removing the line entirely) all of the broken `"or later)` notes as pointed out by @goneall [here][1].

[1]: https://github.com/spdx/license-list-XML/pull/553#discussion_r158591685

goneall:
3.1 release?  A lot of changes to take in the day of the release and the tools currently merges the notes.

wking:
> A lot of changes to take in the day of the release…

True, but a lot of these are fixing the `"or later)` typos.  I can file a PR to just fix those, but this one is about as big and I think it's more helpful than the narrower change.  I'd rather get the `"or later)` typos fixed somehow before we cut the next release.

And regardless of this PR, there's already a crazy amount of change going on right before a release that will apparently be a major version bump.  Can we get an RC or something to give things a day or two to cool off?

goneall:
@jlovejoy If you're OK with the notes updates, I'm OK including in the release.

I did read through the changes and it looks pretty straightforward.

jlovejoy:
I'm not okay with removing this info full stop. would need individual review.

wking:
On Thu, Dec 28, 2017 at 03:14:50PM -0800, Jilayne Lovejoy wrote:
> I'm not okay with removing this info full stop. would need
> individual review.

The wxWindows change was already in its own commit, so I've filed #577
with just that change.  If/when that lands, I can proceed to file
single-license PRs for the GPL family.

wking:
I've rebased this onto master now that #577 has landed the wxWindows change.  I've also spun off the AGPL-3.0 change into a single-license PR with #597, in case folks want to continue with [per-license review][1].  I'll keep this branch up to date in case you want to merge all of these very similar changes in one go.

[1]: https://github.com/spdx/license-list-XML/pull/566#issuecomment-354372728

jlovejoy:
ok, so let me confirm I'm getting this in it's current form: the PR now updates all the deprecated licenses that ended up with two <notes> tags - one that has the deprecated info and one that has the old/original "notes" - the PR simply removes the latter for all, so that all that will remain is the notes related to deprecation. right?
and I agree - we should only have one notes tag with all notes contained therein.

wking:
On Thu, Jan 11, 2018 at 01:43:17PM -0800, Jilayne Lovejoy wrote:
> … the PR now updates all the deprecated licenses that ended up with
> two <notes> tags - one that has the deprecated info and one that has
> the old/original "notes" - the PR simply removes the latter for all,
> so that all that will remain is the notes related to
> deprecation. right?  and I agree - we should only have one notes tag
> with all notes contained therein.

Yup.

I'm in favor of removing <notes> from other places as well, e.g. [1]
is redundant with [2].  But those are harder to grep for than doubled
<notes>, so I've left them alone for now.

[1]: https://github.com/spdx/license-list-XML/blob/57941c7c5c0a1f8aae1293f60d5ab9bd2988b84b/src/GPL-3.0-or-later.xml#L31
[2]: https://github.com/spdx/license-list-XML/blob/57941c7c5c0a1f8aae1293f60d5ab9bd2988b84b/src/GPL-3.0-or-later.xml#L36

jlovejoy:
yeah, I'm really not in favor of removing notes that have been there all along (literally, in these cases since v1.0!!) and only was okay with it here as we ended up with two sets of notes and this leaves the more important one (about deprecation).  
but I also may be biased to leaving stuff that has been there all along, b/c I was the one who put those notes there to begin with :-p
In any case, unless the notes are wrong or confusing, it's really not important to change them.

PR #568, opened 2017-12-27
Add title to LLVM-exception

goneall:
The upstream text includes a title.  Adding to the exception text.

### DIFF
--- a/src/exceptions/LLVM-exception.xml
+++ b/src/exceptions/LLVM-exception.xml
@@ -5,6 +5,9 @@
          <crossRef>http://lists.llvm.org/pipermail/llvm-dev/2017-August/116266.html</crossRef>
       </crossRefs>
       <notes>This exception was created specifically to be used with Apache-2.0</notes>
+	  <titleText>---- LLVM Exceptions to the Apache 2.0 License ----
+			<br/>
+	  </titleText>
       <p> As an exception, if, as a result of your compiling your source 
       code, portions of this Software are embedded into an Object form of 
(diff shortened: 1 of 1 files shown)


src/exceptions/LLVM-exception.xml
```diff
@@ -5,6 +5,9 @@
          <crossRef>http://lists.llvm.org/pipermail/llvm-dev/2017-August/116266.html</crossRef>
       </crossRefs>
       <notes>This exception was created specifically to be used with Apache-2.0</notes>
+	  <titleText>---- LLVM Exceptions to the Apache 2.0 License ----
+			<br/>
```
wking: I'd rather wrap with `<p>` and drop the `<br/`> to match the bulk of our existing licenses.

src/exceptions/LLVM-exception.xml
```diff
@@ -5,6 +5,9 @@
          <crossRef>http://lists.llvm.org/pipermail/llvm-dev/2017-August/116266.html</crossRef>
       </crossRefs>
       <notes>This exception was created specifically to be used with Apache-2.0</notes>
+	  <titleText>---- LLVM Exceptions to the Apache 2.0 License ----
```
wking: I think these dashes are part of header markup.  I'd rather see them optional inside content-less `<alt>` tags (see 8b32737, #515, for a similar change to a different license.

goneall:
@wking Updated, although I didn't follow the pattern on the dashes from the other licenses.  I just kept them as optional rather than as a variable text allowing for multiple dashes.

Issue #541, opened 2017-12-21
update LLVM name / identifier


jlovejoy:
pending feedback from LLVM people (sent email to Chris Lattner, et al)

jlovejoy:
moving to 3.1 release for end of Jan due to not hearing back from LLVM yet on name / short identifier

wking:
Does that mean we should temporarily remove `src/exceptions/LLVM-exception.xml` (originally added in a510984f5)?

jlovejoy:
yes, @goneall is taking care of that

goneall:
Deleted LLVM for the 3.0 release and added pull request #570 to add it back in once the LLVM issues are resolved.

wking:
> merged PR, issue resolved

I still have [an issue][1] with the PR, but it's not actually merged yet.  Do we want to re-open this issue until the PR lands?  I'm also fine just using the PR to track, without keeping the issue open.

[1]: https://github.com/spdx/license-list-XML/pull/570#discussion_r158875572

jlovejoy:
I can only click on things so quickly... obviously not as quick as you are at reading new comments! 
response in PR... ;)

wking:
> I can only click on things so quickly... obviously not as quick as you are at reading new comments!

You can always merge the PR first, and then close the issue with “merged PR” ;).  But in this case, we can say the issue was closed because [the PR is a sufficient placeholder][1] (which works regardless of whether the PR is open or closed).

[1]: https://lists.spdx.org/pipermail/spdx-legal/2018-March/002504.html

PR #570, opened 2017-12-27
Add back LLVM exception for release 3.1

goneall:
This was temporarily removed from the 3.0 release.  To be added once issue #541 is resolved.

### DIFF
--- a/src/exceptions/LLVM-exception.xml
+++ b/src/exceptions/LLVM-exception.xml
@@ -1,0 +1,28 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<SPDXLicenseCollection xmlns="http://www.spdx.org/license">
+   <exception licenseId="LLVM-exception" name="LLVM Exceptions to the Apache 2.0">
+      <crossRefs>
+         <crossRef>http://lists.llvm.org/pipermail/llvm-dev/2017-August/116266.html</crossRef>
+      </crossRefs>
+      <notes>This exception was created specifically to be used with Apache-2.0</notes>
+	  <titleText>
+		<p>
+		<optional>---- </optional>LLVM Exceptions to the Apache 2.0 License<optional> ----</optional>
+		</p>
+	  </titleText>
+      <p> As an exception, if, as a result of your compiling your source 
+      code, portions of this Software are embedded into an Object form of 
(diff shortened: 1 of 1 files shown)


src/exceptions/LLVM-exception.xml
```diff
@@ -0,0 +1,28 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<SPDXLicenseCollection xmlns="http://www.spdx.org/license">
+   <exception licenseId="LLVM-exception" name="LLVM Exceptions to the Apache 2.0">
+      <crossRefs>
+         <crossRef>http://lists.llvm.org/pipermail/llvm-dev/2017-August/116266.html</crossRef>
+      </crossRefs>
+      <notes>This exception was created specifically to be used with Apache-2.0</notes>
+	  <titleText>
+		<p>
+		<optional>---- </optional>LLVM Exceptions to the Apache 2.0 License<optional> ----</optional>
```
wking: I'd rather have the dashes inside an empty `<alt>` (see previous discussion [here][1] and [here][2]), because I see the dashes as header markup.  I'd much rather they *not* be part of our canonical HTML, and instead would rather use `<h1>` or other canonical HTML markup for “this is a header”.  We don't have “this is a header” markup in our XML schema yet, but I'd still rather not have this be a part of the canonical HTML we display under spdx.org/licenses/.

[1]: https://github.com/spdx/license-list-XML/pull/568#discussion_r158869521
[2]: https://github.com/spdx/license-list-XML/pull/568#issuecomment-354181552

goneall:
This PR has been rebased and updated to comply with the schema changes in pull request #452

jlovejoy:
got confirmation from LLVM project as to identifier

src/exceptions/LLVM-exception.xml
jlovejoy: leaving as optional - we can update to <alt> later if need be, but this is not substantive and the dashes are taken from the email confirming the new exception, which is not yet showing up in the source repo. We'll update the URL when that hits and can review this then.

wking: > leaving as optional - we can update to later if need be, but this is not substantive…

That's exactly why I prefer `<alt>`.  Why should we have an opinion on the number of dashes?

PR #571, opened 2017-12-28
GPL-2.0*: Remove erroneous whitespace from around <optional>

wking:
Recover from some whitespace which was accidentally introduced by a0879e6e (#553).

This PR is similar to #565, which is restoring whitespace which was accidentally removed by a0879e6e.

### DIFF
--- a/src/GPL-2.0-only.xml
+++ b/src/GPL-2.0-only.xml
@@ -22,8 +22,7 @@
       <p>
         You should have received a copy of the GNU General Public License
         along with this program; if not, write to the Free Software Foundation,
-        Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301
-        <optional>,</optional>
+        Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301<optional>,</optional>
         USA.
       </p>
     </standardLicenseHeader>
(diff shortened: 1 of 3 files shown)


wking:
Does milestoning this 3.1 mean that I should switch to `<alt>` to match folks who decide to use `02110-1301 ,` (with the space that I'm trying to use in this commit)?  That would now technically match our template and *recommended* text, while `02110-1301,` would not match our template.

goneall:
> Does milestoning this 3.1 mean that I should switch to <alt> to match folks who decide to use 02110-1301 , (with the space that I'm trying to use in this commit)? 

What space is ignored when matching, so no need to introduce an `<alt...` tag

wking:
> [White] space is ignored when matching...

I agree that a single space, multiple spaces, tabs and spaces, etc. are equivalent.  I'm less clear on adding whitespace where there was none, or removing whitespace where the was some.  Do we really want to ignore those too?

jlovejoy:
@goneall is the one to give the definitive answer here, as it depends on how tools deal with the given white space scenario.

goneall:
Finally getting back to this.  

This is a bit of a grey area for the tooling.  The matching guidelines state that whitespace can be ignored.   Most tool implementations tokenize based on the punctuation characters in addition to the white space, however this isn't specified in the matching guidelines.  The SPDX tools will handle this fine.  There is a small probability that other tools using the matching guidelines will not be able to match.

Since there is also a cosmetic improvement removing extra spaces, I would suggest merging this PR.

@wking If you wouldn't mind rebasing, I'll merge.

wking:
Rebased with e407fc6 → 744e9ee.

PR #514, opened 2017-12-14
GPL-2.0: Mark postal-code commas as optional

wking:
Upstream is not consistent about this:

```console
$ curl -s https://www.gnu.org/licenses/old-licenses/gpl-2.0-standalone.html | grep USA
51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA
Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
$ curl -s https://www.gnu.org/licenses/old-licenses/gpl-2.0.txt | grep USA
 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
    51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
```

so support both forms.  I've stuck with our old comma version as canonical.

[Reported][1] by @1138-4EB.

[1]: https://github.com/benbalter/licensee/issues/247#issuecomment-351756955

PR #572, opened 2017-12-28
GPL-2.0*: More alt/optional entries to match FSF's canonical text/plain

wking:
These changes address additional differences between [the FSF's text/plain version][1] (unchanged since 2007-07-16 [according to the Internet Archive][2]) and [the FSF's HTML version][3].  There has been previous work in this direction in #496 and #514.

The final paragraph (“This General Public License does not permit…”) is in both the text/plain and HTML FSF versions.  I'm not clear on why it wasn't included in our template (our GPL-3.0 template does include a similar paragraph).  I've put it in a new `<optional>` block to cover folks who were using the text we previously recommended (which lacked the paragraph).

The WIP tag is because this will conflict with #571, which I think is more important (or at least, easier to review), and because I haven't touched GPL-2.0-only or GPL-2.0-or-later (which share the same license content as GPL-2.0).

[1]: https://www.gnu.org/licenses/old-licenses/gpl-2.0.txt
[2]: https://web.archive.org/web/20070716031727/https://www.gnu.org/licenses/old-licenses/gpl-2.0.txt
[3]: https://www.gnu.org/licenses/old-licenses/gpl-2.0.en.html

### DIFF
--- a/src/GPL-2.0-only.xml
+++ b/src/GPL-2.0-only.xml
@@ -38,7 +38,7 @@
       </p>
     </titleText>
     <p>
-      Copyright (C) 1989, 1991 Free Software Foundation, Inc.<br></br>
+      Copyright (C) 1989, 1991 Free Software Foundation, Inc.<alt name="incComma" match="|,"/><br/>
       51 Franklin Street, Fifth Floor, Boston, MA 02110-1301<optional>,</optional>
       USA
     </p>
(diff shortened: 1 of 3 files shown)


goneall:
@wking Please rebase/resolve merge conflict

wking:
On Fri, Feb 16, 2018 at 12:08:15PM -0800, goneall wrote:
> Please rebase/resolve merge conflict

Rebased and extended to `GPL-2.0-only` or `GPL-2.0-or-later` with
7bc88e9 → 91044d8.

PR #573, opened 2017-12-28
Update URL for Imlib2

sam-ellis:
URL no longer valid. Provided a new URL that works.


bradleeedmondson:
To @goneall's point in https://github.com/spdx/license-list-XML/pull/653#issuecomment-397712826 I would say that the original link should be added back -- at this stage, alongside the new, working link, and without known-dead markup (as we haven't come to consensus on whether & how to do this markup).

filed #659 to add the old URL back

PR #578, opened 2017-12-29
GPL-3.0*: Copy the standard header from the license body

wking:
I've copy/pasted exactly from the license body, except for:

* Some tab → space replacements for consistent indenting.
* Adjusting the version text for the `-only` forms.  I think the version text we use is pretty strange, but I'll deal with that in follow-up work.

Semantic changes due to this standardization:

* GPL-3.0 gets “version.” → “version 3.”.

* GPL-3.0+ gets “either version 3 of the License, or (at your option) any later” where it previously had nothing.

* GPL-3.0-or-later gets “version 3 or any later version” → “either version 3 of the License, or (at your option) any later version”.

* Every license I touch gets `<…>` around its URL, which has been in the license templates since f6c34536 (#85) and similar.

* Every license I touch gets the optional HTTPS from df182a1a (#450) and similar.

I haven't bothered adding alt/optional to support our old forms, because I'm expecting folks in the wild to be copy/pasting from the FSF license text instead of our `<standardLicenseHeader>`s.  But I can work up alt/optional entries if folks feel they are needed.

### DIFF
--- a/src/GPL-3.0+.xml
+++ b/src/GPL-3.0+.xml
@@ -9,12 +9,23 @@
 	  <notes>DEPRECATED: Use the license identifier GPL-3.0-or-later</notes>
       <standardLicenseHeader> Copyright (C) 
     <alt match=".+" name="copyright">&lt;year&gt; &lt;name of author&gt;</alt>
-    This program is free software:
-    you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the
-    Free Software Foundation, version. This program is distributed in the hope that it will be useful, but WITHOUT
-    ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
-    GNU General Public License for more details. You should have received a copy of the GNU General Public License
-    along with this program. If not, see http://www.gnu.org/licenses/
+      <p>
+        This program is free software: you can redistribute it and/or
+        modify it under the terms of the GNU General Public License
+        as published by the Free Software Foundation, either version
+        3 of the License, or (at your option) any later version.
+      </p>
+      <p>
+        This program is distributed in the hope that it will be
+        useful, but WITHOUT ANY WARRANTY; without even the implied
+        warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR
+        PURPOSE. See the GNU General Public License for more details.
+      </p>
+      <p>
+        You should have received a copy of the GNU General
+        Public License along with this program. If not, see
+        &lt;http<optional>s</optional>://www.gnu.org/licenses/&gt;.
+      </p>
   </standardLicenseHeader>
       <notes>This license was released: 29 June 2007 This refers to when this GPL 3.0 only is being used (as opposed to
          "or later).</notes>
--- a/src/GPL-3.0-or-later.xml
+++ b/src/GPL-3.0-or-later.xml
@@ -9,15 +9,23 @@
     <standardLicenseHeader>
       Copyright (C)<alt name="copyright"
       match=".+">&lt;year&gt; &lt;name of author&gt;</alt>
-	<p>This program is free software: you can redistribute it and/or
-      	modify it under the terms of the GNU General Public License as
-	published by the Free Software Foundation, version 3 or any later version.</p>
-      	<p>This program is distributed in the hope that it will be useful, but
-      	WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
-      	or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License
-	for more details.</p>  
-    	<p>You should have received a copy of the GNU General
-	Public License along with this program. If not, see http://www.gnu.org/licenses/</p>
+      <p>
+        This program is free software: you can redistribute it and/or
+        modify it under the terms of the GNU General Public License
+        as published by the Free Software Foundation, either version
+        3 of the License, or (at your option) any later version.
+      </p>
+      <p>
+        This program is distributed in the hope that it will be
+        useful, but WITHOUT ANY WARRANTY; without even the implied
+        warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR
+        PURPOSE. See the GNU General Public License for more details.
+      </p>
+      <p>
+        You should have received a copy of the GNU General
+        Public License along with this program. If not, see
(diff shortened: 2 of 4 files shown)


src/GPL-3.0+.xml
```diff
@@ -9,12 +9,23 @@
 	  <notes>DEPRECATED: Use the license identifier GPL-3.0-or-later</notes>
       <standardLicenseHeader> Copyright (C) 
     <alt match=".+" name="copyright">&lt;year&gt; &lt;name of author&gt;</alt>
-    This program is free software:
-    you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the
-    Free Software Foundation, version. This program is distributed in the hope that it will be useful, but WITHOUT
```
wking: This is the broken “, version.”.

src/GPL-3.0+.xml
```diff
@@ -9,12 +9,23 @@
...
-    ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
-    GNU General Public License for more details. You should have received a copy of the GNU General Public License
-    along with this program. If not, see http://www.gnu.org/licenses/
+      <p>
+        This program is free software: you can redistribute it and/or
+        modify it under the terms of the GNU General Public License
+        as published by the Free Software Foundation, either version
+        3 of the License, or (at your option) any later version.
```
wking: This is the fixed version text.

src/GPL-3.0-or-later.xml
```diff
@@ -9,15 +9,23 @@
     <standardLicenseHeader>
       Copyright (C)<alt name="copyright"
       match=".+">&lt;year&gt; &lt;name of author&gt;</alt>
-	<p>This program is free software: you can redistribute it and/or
-      	modify it under the terms of the GNU General Public License as
-	published by the Free Software Foundation, version 3 or any later version.</p>
```
wking: This is the broken version text.

src/GPL-3.0-or-later.xml
```diff
@@ -9,15 +9,23 @@
...
-      	WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
-      	or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License
-	for more details.</p>  
-    	<p>You should have received a copy of the GNU General
-	Public License along with this program. If not, see http://www.gnu.org/licenses/</p>
+      <p>
+        This program is free software: you can redistribute it and/or
+        modify it under the terms of the GNU General Public License
+        as published by the Free Software Foundation, either version
+        3 of the License, or (at your option) any later version.
```
wking: This is the fixed version text.

src/GPL-3.0.xml
```diff
@@ -14,14 +14,22 @@
     <standardLicenseHeader>
       Copyright (C)<alt name="copyright"
       match=".+">&lt;year&gt; &lt;name of author&gt;</alt>
-      This program is free software: you can redistribute it and/or
-      modify it under the terms of the GNU General Public License as
-      published by the Free Software Foundation, version. This program is
```
wking: This is the broken version text.

src/GPL-3.0.xml
```diff
@@ -14,14 +14,22 @@
...
-      distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
-      without even the implied warranty of MERCHANTABILITY or FITNESS FOR A
-      PARTICULAR PURPOSE. See the GNU General Public License for more details.
-      You should have received a copy of the GNU General Public License
-      along with this program. If not, see http://www.gnu.org/licenses/
+      <p>
+        This program is free software: you can redistribute it and/or
+        modify it under the terms of the GNU General Public License
+        as published by the Free Software Foundation, version 3.
```
wking: This is the fixed-to-match `GPL-3-only` version text.

There are a number of forms of this grant in the Linux kernel, and we probably want to support more of them.  But that is a bigger change than I wanted to bite off in this PR.

wking:
And these header issues also affect the other GPL versions, but I thought I'd PR the 3.0 changes first.  I'll follow up with 1.0 and 2.0 fixups if/when this PR lands.

jlovejoy:
yikes, good catch! I'm not sure how no one noticed this all these years... I have seen the modified header that was included here, but it is not the standard one as recommended by the license!

jlovejoy:
oh, I'm not really sure it matters if we fix deprecated licenses for this kind of thing? @goneall @bradleeedmondson - do you have opinions on that?
(will merge this shortly)

bradleeedmondson:
I would say go ahead and fix. It will be wrong in everything before 3.1 and
correct in 3.1, and consumers can draw their own conclusions from that (if
they even notice).

wking:
This same issue affects at least `GPL-2.0-or-later` as well (compare [here][1] and [here][2]).  Should I file a similar PR for the 2.0 series, or do we want to wait for #581 (which will allow us to drop this copy/pasting entirely)?  I think the trade offs are:

* How much review time do we want to spend?  A separate 2.0 PR would take more review time that would be wasted if we end up going with #581.
* How quickly do we want the other headers fixed?  #581 and its #452 dependency are both WIP, and #452 has been marinating for a few months.  If we think those PRs are going to miss 3.1, we may want to spend the extra review time to handle a separate 2.0 PR.

[1]: https://github.com/spdx/license-list-XML/blob/0e6a10d6528cb3b30c88572820a19d3d338b26d9/src/GPL-3.0-or-later.xml#L16
[2]: https://github.com/spdx/license-list-XML/blob/0e6a10d6528cb3b30c88572820a19d3d338b26d9/src/GPL-3.0-or-later.xml#L803

goneall:
Since the license is deprecated, I don't think we should spend too much time/effort.  Since this PR is already created, I agree with @bradleeedmondson we should merge it.  I wouldn't put in any additional effort until #581 is resolved.

wking:
> Since the license is deprecated…

`GPL-2.0-or-later` is not deprecated.  Although I see I dropped `GPL-3.0-or-later` links by mistake.  Compare:

* https://github.com/spdx/license-list-XML/blob/0e6a10d6528cb3b30c88572820a19d3d338b26d9/src/GPL-2.0-or-later.xml#L15
* https://github.com/spdx/license-list-XML/blob/0e6a10d6528cb3b30c88572820a19d3d338b26d9/src/GPL-2.0-or-later.xml#L420

PR #580, opened 2017-12-29
validate-schema.js: Drop Gulp for raw Node.js

wking:
Gulp doesn't help us with much, and it makes passing arguments hard, with complaints like:

```console
$ gulp validate src/Apache-2.0.xml
[10:02:19] Using gulpfile ~/src/spdx/license-list-XML/gulpfile.js
[10:02:19] Task 'src/Apache-2.0.xml' is not in your gulpfile
…
```

You can [teach Gulp tasks about arguments][1], but the recommended approach requires even more external dependencies.  Instead, this pull request just drops Gulp in favor raw Node.js.

It also adds the ability to validate individual files from the command line:

```console
$ ./validate-schema.js src/Apache-2.0.xml
validation complete 1 passed, 0 failed
```

which is useful for testing schema changes, since you can edit the XSD and a single license while iterating on the schema and not be distracted by test failures from all the other licenses.

[1]: https://gulpjs.org/recipes/pass-arguments-from-cli.html


goneall:
Looks reasonable, I'll go ahead and merge it in.

@wking I'm working on a major change to the continuous integration to perform more extensive tests and automatically deploy to using the Java application which will replace the Gulp and Node applications.  This is in PR #513

wking:
On Fri, Dec 29, 2017 at 10:58:15AM -0800, goneall wrote:
> … automatically deploy to using the Java application which will
> replace the Gulp and Node applications.

It looks like #513 is still using Node in parallel [1].  I'd rather
keep that in Node, because Gentoo's Node support is *way* better than
it's Java support, and I'd like to be able to continue to run XSD
validation locally without a heavy Java dependency.  But if you end up
covering XSD validation in Java, I can always continue to use the Node
tooling locally.  So I'm a weak -1 (maybe -0.3?) on dropping the Node
XSD validation.

[1]: https://github.com/spdx/license-list-XML/pull/513/files#diff-354f30a63fb0907d4ad57269548329e3R7

goneall:
> It looks like #513 is still using Node in parallel [1].  I'd rather keep that in Node

Java also does XSD validation, but it doesn't hurt to include the Node as well.  I'll keep it in.

Issue #579, opened 2017-12-29
GPL-*-only: Drop standard license headers?


wking:
We currently provide these (e.g. [here][1]), but the license text only recommends the or-later header (e.g. [here][2]).  That means the `-only` headers do not satisfy our:

> [Leave this field blank if there is no standard header as specifically defined in the license][3]

condition.  And I'm not even sure if they satisfy the [potential][4] weaker condition of:

> Recommended by the license steward.

You might have to weaken to:

> Frequently found in the wild.

and that's a big weakening.  Although #489 is going that far for the license text, so maybe we're ok going that far for the headers.  Either way, I think we need to either soften our [header claims][3] or remove the `GPL-*-only` headers.

[1]: https://github.com/spdx/license-list-XML/blob/e176dedba036533f045dc926429afaab7c38e16d/src/GPL-3.0-only.xml#L9
[2]: https://github.com/spdx/license-list-XML/blob/e176dedba036533f045dc926429afaab7c38e16d/src/GPL-3.0-only.xml#L796
[3]: https://spdx.org/spdx-license-list/license-list-overview
[4]: https://github.com/spdx/license-list/issues/5#issuecomment-284526732

jlovejoy:
sorry @wking but this is taking the explanation of the standard header field a bit too rigidly. The standard headers for GPL are very well known and used; the difference of the "any later version" language being present or removed in the license notice is the recommendation from the FSF (reiterated recently by RMS) and codified by the clause about later versions. No one has ever had an issue here before, notably when that definition was made at a time when the two variations were listed individually on the license list (as they now are again). This is a bit of a solution in search of a problem :)

wking:
On Thu, Jan 11, 2018 at 08:50:55PM +0000, Jilayne Lovejoy wrote:
> … the difference of the "any later version" language being present
> or removed in the license notice is the recommendation from the FSF
> (reiterated recently by RMS) …

Can you link to FSF-recommended wording for the only forms?  Then we
could weaken the policy to just “Recommended by the license steward”.

PR #492, opened 2017-11-25
Disclaimer templating of PostgreSQL license

mlinksva:
As seen at https://opensource.org/licenses/PostgreSQL

Issue #435, opened 2017-09-22
Add upstream header suggestions to exceptions?


wking:
#434 seems to pivot on [whether or not to include header text suggested by the license steward, but which are not contained in the license text itself][1].  Assuming we want that, I think we may also want to include steward-recommendations for exception grants.  For example, [GCC is using][2]:

> Under Section 7 of GPL version 3, you are granted additional permissions described in the GCC Runtime Library Exception, version 3.1, as published by the Free Software Foundation.

Assuming we can get the FSF to make that recommended wording explicit (maybe [here][3]?  Or as a template in a GPL 3.0 §7 FAQ?), it would be good to record it in this repository to help with [exception matching][4] (CC @goneall, in case he had other plans for detecting exception grants).

So there are a number of assumptions that need to get worked out before we need an answer for this, but I thought I'd file it in case folks want to discuss before those settle.

[1]: https://github.com/spdx/license-list/issues/5#issuecomment-307292702
[2]: https://gcc.gnu.org/git/?p=gcc.git;a=blob;f=libgcc/libgcov.h;h=eb5a9147bef4fd8ef35e58ecbb8cf8de12aa3295;hb=1bd23ca8c30f4827c4bea23deedf7ca33a86ffb5#l16
[3]: https://www.gnu.org/licenses/gcc-exception-3.1-faq.html
[4]: https://github.com/spdx/license-test-files/issues/2#issuecomment-331510002

wking:
This issue is currently the only significant entry separating our [exception schema][1] from our [license schema][2].  If we decide to markup exception header grants, we could drop `ExceptionType` and use `LicenseType` for both `<license>` and `<exception>`.  I think that's a useful cleanup even if we don't have any exception header grants that we want to file immediately.

[1]: https://github.com/spdx/license-list-XML/blob/5497f5fa183cf42c67a7d59cf52a53b664187ba3/schema/ListedLicense.xsd#L19-L37
[2]: https://github.com/spdx/license-list-XML/blob/5497f5fa183cf42c67a7d59cf52a53b664187ba3/schema/ListedLicense.xsd#L38-L59

jlovejoy:
oof, this is old and pretty sure "Assuming we can get the FSF to make that recommended wording explicit" is not going to happen

PR #581, opened 2017-12-29
Allow standardLicenseHeader under text

wking:
This PR builds on #452; review that first.

Two changes here:

* Rename `<standardLicenseHeader>` → `<fileGrant>`.  “License” is redundant as long as we only include these for licenses, and would be incorrect if/when we start supplying these for exceptions.  “standard” is also redundant as long as we are requiring these to come from the license text itself (more on that below).  “header” alone might be confused with `<h1>`, etc., and is about where the content shows up.  `<fileGrant>`, on the other hand, is about what the content *is*.  In fact, the GPL family grants are generic enough to be applied at the program level (as well as per-file), but the file-specific `<fileGrant>` is more specific for licenses like Apache-2.0 whose text only makes sense in a per-file context.

* Make the `<fileGrant>` element a descendant of the text element (it used to be a sibling).  We may restore the possibility of siblings later, but for now, the ancestor requirement makes it easy to [enforce][1]:

    ```
    F) Standard License Header

    * Should only include text intended to be put in the header of
      source files or other files as specified in the license or
      license appendix when specifically delineated
    …
    * Leave this field blank if there is no standard header as
      specifically defined in the license
    ```

This also makes it easy to cover licenses where the license text and header are identical (e.g. the BSDs and the OFL, #451), because you can use:

```xml
<text>
  <fileGrant>
    …
  </fileGrant>
<text>
```

and it makes it easy to stay DRY and avoid header/license divergence (#578).

I've mocked up the changes to the `Apache-2.0`, so folks can see what this will look like.  If we think this approach has legs (and #452 lands), I'll go through and update the rest of our licenses.

[1]: https://spdx.org/spdx-license-list/license-list-overview#fields
[2]: https://github.com/spdx/license-list-XML/issues/451

### DIFF
--- a/schema/ListedLicense.xsd
+++ b/schema/ListedLicense.xsd
@@ -32,8 +32,7 @@
 		<all>
 			<element name="crossRefs" type="tns:crossRefsType" minOccurs="0" maxOccurs="1"/>
 			<element name="notes" type="tns:formattedFixedTextType" minOccurs="0" maxOccurs="1"/>
-			<element name="standardLicenseHeader" type="tns:formattedAltTextType" minOccurs="0" maxOccurs="1"/>
-			<element name="text" type="tns:formattedAltTextType" minOccurs="0" maxOccurs="1"/>
+			<element name="text" type="tns:formattedAltFileGrantTextType" minOccurs="0" maxOccurs="1"/>
 		</all>
 		<attribute name="licenseId" type="string" use="required" />
 		<attribute name="isOsiApproved" type="boolean"/>
@@ -55,10 +54,19 @@
 	<complexType name="optionalType" mixed="true">
 		<group ref="tns:formattedAltTextGroup" minOccurs="0" maxOccurs="unbounded"/>
 	</complexType>
+	<complexType name="optionalFileGrantType" mixed="true">
+		<group ref="tns:formattedAltFileGrantTextGroup" minOccurs="0" maxOccurs="unbounded"/>
+	</complexType>
 	<complexType name="listType">
 		<choice minOccurs="1" maxOccurs="unbounded">
 			<element name="item" type="tns:formattedAltTextType"/>
 			<element name="list" type="tns:listType"/>
+		</choice>
+	</complexType>
+	<complexType name="listFileGrantType">
+		<choice minOccurs="1" maxOccurs="unbounded">
+			<element name="item" type="tns:formattedAltFileGrantTextType"/>
+			<element name="list" type="tns:listFileGrantType"/>
 		</choice>
 	</complexType>
 	<complexType name="emptyType"/>
@@ -67,6 +75,9 @@
 	</complexType>
 	<complexType name="formattedAltTextType" mixed="true">
 		<group ref="tns:formattedAltTextGroup" minOccurs="0" maxOccurs="unbounded"/>
+	</complexType>
+	<complexType name="formattedAltFileGrantTextType" mixed="true">
+		<group ref="tns:formattedAltFileGrantTextGroup" minOccurs="0" maxOccurs="unbounded"/>
 	</complexType>
 	<group name="formattedFixedTextGroup">
 		<choice>
@@ -88,6 +99,19 @@
 			<element name="br" type="tns:emptyType" minOccurs="0" maxOccurs="unbounded"/>
 			<element name="titleText" type="tns:formattedAltTextType" minOccurs="0" maxOccurs="1"/>
 			<element name="copyrightText" type="tns:formattedAltTextType" minOccurs="0" maxOccurs="1"/>
+		</choice>
+	</group>
+	<group name="formattedAltFileGrantTextGroup">
+		<choice>
+			<element name="p" type="tns:formattedAltFileGrantTextType" minOccurs="0" maxOccurs="unbounded"/>
+			<element name="bullet" type="string" minOccurs="0" maxOccurs="unbounded"/>
+			<element name="list" type="tns:listFileGrantType" minOccurs="0" maxOccurs="unbounded"/>
+			<element name="optional" type="tns:optionalFileGrantType" minOccurs="0" maxOccurs="unbounded"/>
+			<element name="alt" type="tns:altType" minOccurs="0" maxOccurs="unbounded"/>
+			<element name="br" type="tns:emptyType" minOccurs="0" maxOccurs="unbounded"/>
+			<element name="titleText" type="tns:formattedAltTextType" minOccurs="0" maxOccurs="1"/>
+			<element name="copyrightText" type="tns:formattedAltTextType" minOccurs="0" maxOccurs="1"/>
+			<element name="fileGrant" type="tns:formattedAltTextType" minOccurs="0" maxOccurs="1"/>
 		</choice>
 	</group>
 </schema>
--- a/schema/ListedLicense.xsd
+++ b/schema/ListedLicense.xsd
@@ -32,7 +32,7 @@
 		<all>
 			<element name="crossRefs" type="tns:crossRefsType" minOccurs="0" maxOccurs="1"/>
 			<element name="notes" type="tns:formattedFixedTextType" minOccurs="0" maxOccurs="1"/>
-			<element name="text" type="tns:formattedAltFileGrantTextType" minOccurs="0" maxOccurs="1"/>
+			<element name="text" type="tns:formattedAltStandardLicenseHeaderTextType" minOccurs="0" maxOccurs="1"/>
 		</all>
 		<attribute name="licenseId" type="string" use="required" />
 		<attribute name="isOsiApproved" type="boolean"/>
@@ -54,8 +54,8 @@
 	<complexType name="optionalType" mixed="true">
 		<group ref="tns:formattedAltTextGroup" minOccurs="0" maxOccurs="unbounded"/>
 	</complexType>
-	<complexType name="optionalFileGrantType" mixed="true">
-		<group ref="tns:formattedAltFileGrantTextGroup" minOccurs="0" maxOccurs="unbounded"/>
+	<complexType name="optionalStandardLicenseHeaderType" mixed="true">
+		<group ref="tns:formattedAltStandardLicenseHeaderTextGroup" minOccurs="0" maxOccurs="unbounded"/>
 	</complexType>
 	<complexType name="listType">
 		<choice minOccurs="1" maxOccurs="unbounded">
@@ -63,10 +63,10 @@
 			<element name="list" type="tns:listType"/>
 		</choice>
 	</complexType>
-	<complexType name="listFileGrantType">
+	<complexType name="listStandardLicenseHeaderType">
 		<choice minOccurs="1" maxOccurs="unbounded">
-			<element name="item" type="tns:formattedAltFileGrantTextType"/>
-			<element name="list" type="tns:listFileGrantType"/>
+			<element name="item" type="tns:formattedAltStandardLicenseHeaderTextType"/>
+			<element name="list" type="tns:listStandardLicenseHeaderType"/>
 		</choice>
 	</complexType>
 	<complexType name="emptyType"/>
@@ -76,8 +76,8 @@
 	<complexType name="formattedAltTextType" mixed="true">
 		<group ref="tns:formattedAltTextGroup" minOccurs="0" maxOccurs="unbounded"/>
 	</complexType>
-	<complexType name="formattedAltFileGrantTextType" mixed="true">
-		<group ref="tns:formattedAltFileGrantTextGroup" minOccurs="0" maxOccurs="unbounded"/>
+	<complexType name="formattedAltStandardLicenseHeaderTextType" mixed="true">
+		<group ref="tns:formattedAltStandardLicenseHeaderTextGroup" minOccurs="0" maxOccurs="unbounded"/>
 	</complexType>
 	<group name="formattedFixedTextGroup">
 		<choice>
@@ -101,17 +101,17 @@
 			<element name="copyrightText" type="tns:formattedAltTextType" minOccurs="0" maxOccurs="1"/>
 		</choice>
 	</group>
-	<group name="formattedAltFileGrantTextGroup">
+	<group name="formattedAltStandardLicenseHeaderTextGroup">
 		<choice>
-			<element name="p" type="tns:formattedAltFileGrantTextType" minOccurs="0" maxOccurs="unbounded"/>
+			<element name="p" type="tns:formattedAltStandardLicenseHeaderTextType" minOccurs="0" maxOccurs="unbounded"/>
 			<element name="bullet" type="string" minOccurs="0" maxOccurs="unbounded"/>
-			<element name="list" type="tns:listFileGrantType" minOccurs="0" maxOccurs="unbounded"/>
-			<element name="optional" type="tns:optionalFileGrantType" minOccurs="0" maxOccurs="unbounded"/>
+			<element name="list" type="tns:listStandardLicenseHeaderType" minOccurs="0" maxOccurs="unbounded"/>
+			<element name="optional" type="tns:optionalStandardLicenseHeaderType" minOccurs="0" maxOccurs="unbounded"/>
 			<element name="alt" type="tns:altType" minOccurs="0" maxOccurs="unbounded"/>
 			<element name="br" type="tns:emptyType" minOccurs="0" maxOccurs="unbounded"/>
 			<element name="titleText" type="tns:formattedAltTextType" minOccurs="0" maxOccurs="1"/>
 			<element name="copyrightText" type="tns:formattedAltTextType" minOccurs="0" maxOccurs="1"/>
-			<element name="fileGrant" type="tns:formattedAltTextType" minOccurs="0" maxOccurs="1"/>
+			<element name="standardLicenseHeader" type="tns:formattedAltTextType" minOccurs="0" maxOccurs="1"/>
 		</choice>
 	</group>
 </schema>
--- a/schema/ListedLicense.xsd
+++ b/schema/ListedLicense.xsd
@@ -32,6 +32,7 @@
 		<all>
 			<element name="crossRefs" type="tns:crossRefsType" minOccurs="0" maxOccurs="1"/>
 			<element name="notes" type="tns:formattedFixedTextType" minOccurs="0" maxOccurs="1"/>
+			<element name="standardLicenseHeader" type="tns:formattedAltTextType" minOccurs="0" maxOccurs="1"/>
 			<element name="text" type="tns:formattedAltStandardLicenseHeaderTextType" minOccurs="0" maxOccurs="1"/>
 		</all>
 		<attribute name="licenseId" type="string" use="required" />
--- a/src/ECL-1.0.xml
+++ b/src/ECL-1.0.xml
@@ -5,10 +5,6 @@
       <crossRefs>
          <crossRef>http://opensource.org/licenses/ECL-1.0</crossRef>
       </crossRefs>
-      <standardLicenseHeader> Copyright (c) 
-    <alt match=".+" name="copyright">&lt;year&gt; &lt;copyright holders&gt;</alt> Licensed under the Educational
-    Community License version 1.0 
-  </standardLicenseHeader>
     <text>
       <titleText>
          <p>The Educational Community License</p>
@@ -16,10 +12,12 @@
       <p>This Educational Community License (the "License") applies to any original work of authorship (the
          "Original Work") whose owner (the "Licensor") has placed the following notice immediately following
          the copyright notice for the Original Work:</p>
+      <standardLicenseHeader>
       <p>Copyright (c) 
         <alt match=".+" name="copyright">&lt;year&gt; &lt;copyright holders&gt;</alt>
       </p>
       <p>Licensed under the Educational Community License version 1.0</p>
+      </standardLicenseHeader>
       <p>This Original Work, including software, source code, documents, or other related items, is being provided
          by the copyright holder(s) subject to the terms of the Educational Community License. By obtaining,
          using and/or copying this Original Work, you agree that you have read, understand, and will comply
--- a/src/GFDL-1.1-or-later.xml
+++ b/src/GFDL-1.1-or-later.xml
@@ -5,16 +5,6 @@
     <crossRefs>
       <crossRef>http://www.gnu.org/licenses/old-licenses/fdl-1.1.txt</crossRef>
     </crossRefs>
-    <standardLicenseHeader>
-      Copyright (c) <alt name="copyright" match=".+">YEAR YOUR NAME</alt>
-      Permission is granted to copy, distribute and/or modify this
-      document under the terms of the GNU Free Documentation License,
-      Version 1.1 or any later version published by the Free Software
-      Foundation; with the Invariant Sections being LIST THEIR
-      TITLES, with the Front-Cover Texts being LIST, and with the
-      Back-Cover Texts being LIST. A copy of the license is included
-      in the section entitled "GNU Free Documentation License".
-    </standardLicenseHeader>
     <notes>
       This license was released March 2000
     </notes>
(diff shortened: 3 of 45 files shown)


goneall:
Since standardLicenseHeader has been part of the SPDX spec for a significant period of time, I would not change this unless there is a very compelling issue that needs to be solved.  From the provided information, I do not see a compelling enough issue.

wking:
On Fri, Dec 29, 2017 at 07:43:23PM +0000, goneall wrote:
> Since standardLicenseHeader has been part of the SPDX spec for a
> significant period of time, I would not change this…

What are you objecting to?  The standardLicenseHeader → fileGrant
rename?  Making that element a descendent of text?  Both?

> … unless there is a very compelling issue that needs to be solved.

The rename issue is less important to me; I'm fine punting on that
until we want to mark-up exception headers (#435).  Renaming to
fileGrant in this PR would mean less churn if/when we decide to
address #435, but the rename is just going to be a sed one-liner
whenever we decide to do it.

Making the grant element an descendant of text would make policy
violations like #579 impossible, and it would also make grant/license
divergence like #578 impossible.  Both of those make life easier for
reviewers, who can now leave those checks to the XSD instead of
attempting to apply them by hand.  I think that's a major win.

There is an initial downside as we migrate the source (as with any
schema change), but review should be fairly straightforward, because
it's just removing the previous standardLicenseHeader element and
adding new fileGrant (or whatever) tags inside the license itself.
I've pushed f232814 → 0865d15 to drop my license pretty-printing; see
0865d15 for what review will look like (currently just Apache-2.0).

goneall:
> What are you objecting to?  The standardLicenseHeader → fileGrant
rename?  Making that element a descendent of text?  Both?

primarily standardLicenseHeader → fileGrant

I need to think about the structural change making it a descendent of text.  It seems that the license headers will not always be included in the text for all licenses - but I haven't really thought about it enough to form a definite opinion.

wking:
On Fri, Dec 29, 2017 at 04:40:31PM -0800, goneall wrote:
> > What are you objecting to?  The standardLicenseHeader → fileGrant
> > rename?  Making that element a descendent of text?  Both?
> 
> primarily standardLicenseHeader → fileGrant

Dropped (for now [1]) with 0865d15 → 3a11532.

> It seems that the license headers will not always be included in the
> text for all licenses…

They aren't, but that's currently a policy violation (#579).

And if/when we relax that policy, I think we want to allow
standardLicenseHeader both as a descendant *and* as a sibling of text.
That way we can still DRY up to avoid issues like #578.

[1]: https://github.com/spdx/license-list-XML/pull/581#issuecomment-354495573

jlovejoy:
We are not changing the name, so this Issue is now only about whether standardLicenseHeader is a descendant (this could be a change) of text; or a sibling of text (I think this is what we'd consider it now); or can be both.
I don't see what the point of changing this is? What benefit would this gain?

wking:
On Thu, Jan 11, 2018 at 12:58:10PM -0800, Jilayne Lovejoy wrote:
> I don't see what the point of changing this is? What benefit would
> this gain?

I seem to be having a hard time expressing the benefit [1,2].  Here's
a third take :p.

By allowing `standardLicenseHeader` to live inside the license text,
we can drop the mostly-copy/pasted content [3] in favor of two tiny
tags [4,5].  This makes the header markup easier to maintain, avoiding
divergence like that fixed in [6].  It also gives us a way to resolve
#451 by adding two tags (which is the easiest way to resolve it that
I've seen yet).

[1]: https://github.com/spdx/license-list-XML/pull/581#issue-285137419
[2]: https://github.com/spdx/license-list-XML/pull/581#issuecomment-354495573
[3]: https://github.com/spdx/license-list-XML/pull/581/commits/3a115324651802b59afe5e6d072efeea1483f74b#diff-2f085a88ec62a7900e173b55326b94c8L8
[4]: https://github.com/spdx/license-list-XML/pull/581/commits/3a115324651802b59afe5e6d072efeea1483f74b#diff-2f085a88ec62a7900e173b55326b94c8R206
[5]: https://github.com/spdx/license-list-XML/pull/581/commits/3a115324651802b59afe5e6d072efeea1483f74b#diff-2f085a88ec62a7900e173b55326b94c8R219
[6]: https://github.com/spdx/license-list-XML/pull/578#discussion_r159027914

jlovejoy:
thanks @wking  - doh! that would make sense and does have the benefit of not having to maintain two sets of text, but I think we still need a way to have it separate due the GNU licenses. I think this is a good one to discuss.

bradleeedmondson:
I didn't get this at all when it first crossed my inbox. But I *might* understand now. 

I think what I hear @wking saying is that (1) right now, there's only one way to define standard license header, and that's outside the license text; (2) many license texts themselves include their standard license header; (3) putting together (1) and (2), in the current system you have to repeat header text for licenses that include their own standard header (once in the header, once in the license itself); and finally (4) what he recommends doing is allowing us to choose, per license, to have _either_ the current standard license tag (outside license text) _or_ a "standard license header start" tag with a "standard license header end" tag within the license text, so that there's no need to repeat identical license text, but we can still diverge if we need to (e.g. GPL).

wking:
On Thu, Jan 11, 2018 at 10:04:06PM +0000, Bradlee H. Edmndson wrote:
> (4) what he recommends doing is allowing us to choose, per license,
> to have _either_ the current standard license tag (outside license
> text) _or_ a "standard license header start" tag with a "standard
> license header end" tag within the license text, so that there's no
> need to repeat identical license text, but we can still diverge if
> we need to (e.g. GPL).

Right.  Although I think the “can still diverge” option is a separate
design decision.  Based on [1], it sounds like we want to allow that
option (and relax our current nominal policy), although I'm not sure
how far (do we [still require the license steward to recommend the
header *somewhere*][2]?).  Anyhow, I've pushed 3a11532 → 57b1735
restoring the ability to have sibling standardLicenseHeader, and
57b1735 shows GPL-3.0-only which is compliant with the new schema.

Don't take the current commits too seriously.  All of this will need
to be rebased onto master to pick up recent changes.  I'll do that
if/when #452 lands.

[1]: https://github.com/spdx/license-list-XML/issues/579#issuecomment-357057204
[2]: https://github.com/spdx/license-list-XML/issues/579#issue-285091214

goneall:
@wking Would it be OK to change the title of this issue to `Allow standardLicenseHeader under text`?

We could open a separate issue to rename the standardLicenseHeader if needed - from your previous comment and commit it sounds like you're OK going with standardLicenseHeader for now.

wking:
> … I suggest we allow the `standardLicenseHeader` inside or outside the license text…

Already [done][1].

[1]: https://github.com/spdx/license-list-XML/pull/581#issuecomment-357082239

schema/ListedLicense.xsd
```diff
@@ -17,38 +17,24 @@
...
-			<element name="titleText" type="tns:formattedAltTextType" minOccurs="0" maxOccurs="1"/>
-			<element name="copyrightText" type="tns:formattedAltTextType" minOccurs="0" maxOccurs="1"/>
-			<element name="bullet" type="string"/>
-			<element name="list" type="tns:listType"/>
-			<element name="p" type="tns:formattedAltTextType"/>
-			<element name="optional" type="tns:optionalType"/>
-			<element name="alt" type="tns:altType"/>
-			<element name="br" type="tns:emptyType"/>
-		</choice>
+			<element name="text" type="tns:formattedAltStandardLicenseHeaderTextType" minOccurs="0" maxOccurs="1"/>
```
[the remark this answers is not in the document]
wking: > The name `tns:formattedAltStandardLicenseHeaderTextType` was confusing…

The intention is that it is formatted text (allows `<p>`, etc.) that also allows `<alt>` and `<standardLicenseHeader>`.  I think it reads more easily as [`tns:formattedAltFileGrantTextType`][1] ;).  If we feel the need to stick with `<standardLicenseHeader>`, then I'd rather stick with the current name and not abbreviate to `Header` for that portion.  See:

> … “header” alone might be confused with `<h1>`…

from [my initial comment][2].

[1]: https://github.com/spdx/license-list-XML/pull/581/commits/983694a5f5c2e5096691e5d8dd110aa08c233e75#diff-270bad636868434f3b79b80d26c1bdbbR80
[2]: https://github.com/spdx/license-list-XML/pull/581#issue-285137419

goneall:
> If the plan is to use this approach to resolve the GPL-2.0-or-later header typo(s), can we get this and #452 moved from the “later release” milestone to the 3.1milestone?

I'm good with this - I'd really like to get #452 in since it really helps clean up the tool implementation.

@bradleeedmondson @jlovejoy any objections?

schema/ListedLicense.xsd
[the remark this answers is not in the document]
goneall: OK - Can't think of anything better.  Perhaps we could add some `<annotation><description>...` elements to the XSD to describe the types as the schema is starting to get a bit more complex.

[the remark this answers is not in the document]
wking: > Perhaps we could add some `<annotation><description>...` elements to the XSD to describe the types as the schema is starting to get a bit more complex.

Like #586?  ;)

bradleeedmondson:
I am not xml-fluent enough to understand the changes to the xsd, but as described in the conversation here I'm comfortable with the change and with putting it in 3.1.

Sounds like we might have three legal cases though (in this proposal): separate header text (as currently specced), inline in license text, and inline in appendix text. One: is that right? And two: do we correctly enforce that in the xsd?

wking:
On Fri, Jan 12, 2018 at 10:08:33PM +0000, Bradlee H. Edmndson wrote:
> Sounds like we might have three legal cases though (in this
> proposal): separate header text (as currently specced), inline in
> license text, and inline in appendix text.

That's right, although there's currently no markup to distinguish
between “license text” and “appendix text”.  For example, here's the
GPL-3.0 smoothly transitioning between legal text and the appendix
[1], and there's no special markup there save for an <optional> block
around the appendix content.

> … do we correctly enforce that in the xsd?

Yes.  The XSD currently in flight here allows for separate header text
with [2] (unchanged by this PR) and for header text inside the
license/appendix with [3,4,5].

[1]: https://github.com/spdx/license-list-XML/blob/v3.0/src/GPL-3.0-or-later.xml#L762-L770
[2]: https://github.com/spdx/license-list-XML/pull/581/files#diff-270bad636868434f3b79b80d26c1bdbbR35
[3]: https://github.com/spdx/license-list-XML/pull/581/files#diff-270bad636868434f3b79b80d26c1bdbbR36
[4]: https://github.com/spdx/license-list-XML/pull/581/files#diff-270bad636868434f3b79b80d26c1bdbbR81
[5]: https://github.com/spdx/license-list-XML/pull/581/files#diff-270bad636868434f3b79b80d26c1bdbbR116

bradleeedmondson:
Ah, my mistake -- I was remembering the &lt;optional&gt; tags we wrap appendices in.

I'm good with this structural change landing for 3.1 then (with the understanding that we've set aside the 'fileGrant' field renaming for a later time). Changing milestone on this and #452 to [3.1](https://github.com/spdx/license-list-XML/milestone/3)

wking:
I'll rebase tonight now that #492 has landed.

wking:
I've rebased this around #452 with 57b1735 → dfb0fd7, which also completes the transition for the remaining licenses.

One issue that hasn't been discussed much with this change is that having `<standardLicenseHeader>` inside `<text>` gives no way to distinguish between “variable for the purpose of license matching” and “variable for the purpose of grant matching”.  For example:

```xml
      <license>
        <text>
          <p>Use this grant:</p>
          <standardLicenseHeader>
            <p>Copyright <alt name="copyright" match=".+">[year] [name]</alt>.
            All rights reserved.</p>
          </standardLicenseHeader>
        <text>
      </license>
```

allows variation in both the license and the grant, while:

```xml
      <license>
        <standardLicenseHeader>
          <p>Copyright <alt name="copyright" match=".+">[year] [name]</alt>.
          All rights reserved.</p>
        </standardLicenseHeader>
        <text>
          <p>Use this grant:</p>
          <p>Copyright [year] [name].
          All rights reserved.</p>
        <text>
      </license>
```

only allows variation in the grant.  In cases where we feel that distinction is significant, we'll want to go back to sibling `standardLicenseHeader` elements or add another way to distinguish variable contexts.  For dfb0fd7 I've assumed the distinction is not significant enough to be worth mostly-duplicating the text.

dfb0fd7 also addresses a number of issues with our previous grant recommendations, which I'll point out in inline comments.

src/BitTorrent-1.1.xml
```diff
@@ -502,15 +494,16 @@
          <p>The Notice below must appear in each file of the Source Code of any copy you distribute of the Licensed
          Product or any hereto. Contributors to any Modifications may add their own copyright notices to
          identify their own contributions.</p>
-         <p>License: 
-        <br/>The contents of this file are subject to the BitTorrent Open Source License Version 1.0 (the
+         <p>License:</p>
+         <standardLicenseHeader>
+           <p>The contents of this file are subject to the BitTorrent Open Source License Version <alt name="version" match="1\.[01]">1.1</alt> (the
```
wking: Instead of rolling forward with the typo'd “1.0”, I've changed our canonical version to “1.1” (while continuing to allow the typo match).

src/ECL-1.0.xml
```diff
@@ -5,21 +5,19 @@
       <crossRefs>
          <crossRef>http://opensource.org/licenses/ECL-1.0</crossRef>
       </crossRefs>
-      <standardLicenseHeader> Copyright (c) 
-    <alt match=".+" name="copyright">&lt;year&gt; &lt;copyright holders&gt;</alt> Licensed under the Educational
```
wking: This markup was missing `<p>`, so copyright holders was running into the “Licensed under…” sentence without intervening punctuation.  The markup inside `<text>` is already using separate paragraphs.

src/GFDL-1.1-or-later.xml
```diff
@@ -5,16 +5,6 @@
     <crossRefs>
       <crossRef>http://www.gnu.org/licenses/old-licenses/fdl-1.1.txt</crossRef>
     </crossRefs>
-    <standardLicenseHeader>
-      Copyright (c) <alt name="copyright" match=".+">YEAR YOUR NAME</alt>
-      Permission is granted to copy, distribute and/or modify this
-      document under the terms of the GNU Free Documentation License,
-      Version 1.1 or any later version published by the Free Software
-      Foundation; with the Invariant Sections being LIST THEIR
-      TITLES, with the Front-Cover Texts being LIST, and with the
```
wking: “LIST THEIR TITLES”, etc. are markers for variable text.  In the new `<standardLicenseHeader>` below, I've marked those up with `<alt>.  This same issue applies to the other GFDL versions too.

src/GPL-2.0+.xml
```diff
@@ -280,17 +268,19 @@
...
+         Copyright (C) <alt name="copyright" match=".+">&lt;yyyy&gt; &lt;name of author&gt;</alt></p>
          <p>This program is free software; you can redistribute it and/or modify it under the terms of the GNU
          General Public License as published by the Free Software Foundation; either version 2 of the License,
          or (at your option) any later version.</p>
          <p>This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even
          the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General
          Public License for more details.</p>
          <p>You should have received a copy of the GNU General Public License along with this program; if not, write
-         to the Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301<optional>,</optional> USA.
-         Also add information on how to contact you by electronic and paper mail.</p>
```
wking: This was missing a `<p>` marker to split “Also add information…” into its own paragraph.  This also affected `GPL-2.0-or-later`, `LGPL-2.1+`, and `LGPL-2.1-or-later`.

src/Interbase-1.0.xml
```diff
@@ -5,18 +5,6 @@
       <crossRefs>
          <crossRef>https://web.archive.org/web/20060319014854/http://info.borland.com/devsupport/interbase/opensource/IPL.html</crossRef>
       </crossRefs>
-      <standardLicenseHeader> 
-         <p>&gt;The contents of this file are subject to the Interbase Public License Version 1.0 (the "License");
```
wking: `&gt;` was a typo which did not exist in the `<text>` version below.

src/Interbase-1.0.xml
```diff
@@ -531,15 +519,17 @@
              of the Covered Code under Your choice of the NPL or the alternative licenses, if any,
              specified by the Initial Developer in the file described in Exhibit A.</p>
             <p>EXHIBIT A - InterBase Public License.</p>
-            <p>``The contents of this file are subject to the Interbase Public License Version 1.0 (the
+            <standardLicenseHeader>
+            <p><optional>"</optional>The contents of this file are subject to the Interbase Public License Version 1.0 (the
```
wking: I standardized the LaTeX-ish double-backtick to `"`.  This licence remains wildly inconsistent about quotation characters, but that was the only instance of double backticks.

src/LGPL-2.0+.xml
```diff
@@ -6,15 +6,6 @@
       <crossRefs>
          <crossRef>http://www.gnu.org/licenses/old-licenses/lgpl-2.0-standalone.html</crossRef>
       </crossRefs>
-      <standardLicenseHeader> Copyright (C) 
-    <alt match=".+" name="copyright">year name of author</alt> This library is free software; you can redistribute
-    it and/or modify it under the terms of the GNU Library General Public License as published by the Free Software
-    Foundation; version 2. This library is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
```
wking: This was missing the critical “any later version”.  This also affected `LGPL-2.1+`.

src/LGPL-2.0-or-later.xml
```diff
@@ -5,21 +5,6 @@
     <crossRefs>
       <crossRef>http://www.gnu.org/licenses/old-licenses/lgpl-2.0-standalone.html</crossRef>
     </crossRefs>
-    <standardLicenseHeader>
-      Copyright (C)<alt name="copyright" match=".+">year name of author</alt>
-      <p>This library is free software; you can redistribute it and/or modify it
-      under the terms of the GNU Library General Public License as published
-        by the Free Software Foundation; version 2 or any later version.</p>
```
wking: This differs from the FSFs recommended “any later version” wording from `<text>`.  `LGPL-2.1-or-later` had the same issue.

src/LPPL-1.1.xml
```diff
@@ -240,17 +233,21 @@
...
          file is constrained by the conditions in this license.</p>
       <p>Here is an example of such a notice and statement:</p>
-      <p>%% pig.dtx 
-        <br/>%% Copyright 2001 M. Y. Name 
-        <br/>% 
-        <br/>% This program may be distributed and/or modified under the 
-        <br/>% conditions of the LaTeX Project Public License, either version 1.1 
-        <br/>% of this license or (at your option) any later version. 
-        <br/>% The latest version of this license is in % http://www.latex-project.org/lppl.txt 
-        <br/>% and version 1.1 or later is part of all distributions of LaTeX % version 1999/06/01 or later.
```
wking: This line and the one above had internal `%` which I've wrapped.

src/LPPL-1.3a.xml
```diff
@@ -5,15 +5,6 @@
       <crossRefs>
          <crossRef>http://www.latex-project.org/lppl/lppl-1-3a.txt</crossRef>
       </crossRefs>
-      <standardLicenseHeader> Copyright 
-    <alt match=".+" name="copyright">2003 M. Y. Name</alt> This work may be distributed and/or modified under the
-    conditions of the LaTeX Project Public License, either version 1.3 of this license or (at your option) any later
-    version. The latest version of this license is in http://www.latex-project.org/lppl.txt and version 1.3 or later
-    is part of all distributions of LaTeX version 2003/12/01 or later. This work has the LPPL maintenance status
-    "maintained". This Current Maintainer of this work is
```
wking: I've marked the maintenance status with `<alt>` to accept the other statuses mentioned in the license text.  This also affected `LPPL-1.3c`.

jlovejoy:
discussed on Jan 25 call: decision = allow <standardLicenseHeader> inside text tag and ok/should have same matching tags included within <standardLicenseHeader> (and text)
@goneall to review/sanity check the 48 files changed before merging (if any issues, tag legal person)

src/GFDL-1.1-or-later.xml
goneall: Check the formatting on your comment above - missing some text

wking: > Check the formatting on your comment above - missing some text

Thanks, fixed.

src/BitTorrent-1.1.xml
wking: > Another request for moving this to a separate PR since so I can merge in the technical only changes.

Where do we draw the line for technical changes?  Does the [`>` typo fix][1] count?  Does the [variable GFDL markup][2] count?  I'm not clear on this particular instance being any different, and think it would be easiest to just get legal to review the changes (which I've listed out in 4dcd50b1f8).  If you'd rather me file those fixups *before* this PR, I can do that, although in that case we'll have to update/review both locations.

[1]: https://github.com/spdx/license-list-XML/pull/581#discussion_r161873752
[2]: https://github.com/spdx/license-list-XML/pull/581#discussion_r161873308

wking: >  If you'd rather me file those fixups before this PR, I can do that, although in that case we'll have to update/review both locations.

Nevermind.  I can just drop `BitTorrent-1.1` from this PR, which I've done with 4dcd50b → c526576.  I've filed #603 on top of this PR with the `BitTorrent-1.1` change.

wking:
[Test failures][1] look like they need #602.

src/BitTorrent-1.1.xml
goneall: @wking Thanks for splitting out the separate PR.  I considered flagging the typos as well, but decided that it was OK since the text body didn't change.  In general, I was just trying to review structural changes and any necessary changes to the alt/optional text to allow matching both for the headerText and license text.

goneall:
I think the SPDX tools are not picking up the local schema file.  It does pick it up from the repository, but only the top of the master branch.  I'll need to add an option to refer to a difference schema file.  This will require spinning another version of the tools.  I'll start working on this, but it may take a day or so to get it pushed out.

goneall:
Got the schema update working but ran into another problem with the tool parsing the standardLicenseHeader inside the text block.  Should be an easy fix.

goneall:
I fixed the tool, but when testing against this PR, I ran into an issue with the LPPL-1... licenses.

The following XML snippet:

```
...   This Current Maintainer of this work is <alt name="maintainer" match=".+">M. Y. Name</alt>.
        <br/><optional>%</optional>
        <br/><optional>%</optional> This work consists of the ...
```

does not match the text:
```
... This Current Maintainer of this work is M. Y. Name. 
     % 
     % This work consists of the files pig.dtx ...
```

This is due to the algorithm to which selects which alt text to match searches for the next matching text that isn't variable or optional.  In this case, it is finding the "." after the letter M which then leaves the remaining text after the M. causing the text not to match.

I can not think of a solution that would cause other issues in other matching scenarios.  I'll think about it some more and see if I can come up with a solution.

The basic design issue is that the greedy regexes can consume the entire license which would cause almost all matches to fail.  The algorithm to limit the reach of the greedy regexes can be rather complex.

The code that matches variable text can be found at https://github.com/spdx/tools/blob/fcdf7d243b7a66288de3630c56324e6dde4ff74c/src/org/spdx/compare/CompareTemplateOutputHandler.java#L379

wking:
> I fixed the tool, but when testing against this PR, I ran into an issue with the LPPL-1... licenses.

I've broken the `LPPL-1.3*` changes out into #605 so we can land this PR without waiting for the tool fix.

wking:
@goneall, looks like the tools are still [having issues with the schema change][1].  Thoughts?

[1]: https://travis-ci.org/spdx/license-list-XML/jobs/333844219#L471

goneall:
@wking I have the schema fixed on my local machine.  A little too much defensive coding ;)

I was just working on a fix for the LPPL issue before pushing out another version.

I see you've moved the LPPL changes to a separate PR, so if I don't get a solution in the next hour or so, I'll just go with the schema change fix.

PR #583, opened 2017-12-29
schema/ListedLicense: Drop isDeprecated

wking:
`deprecatedVersion` should be set if and only if `isDeprecated` was true.  So instead of recording that information twice, tooling which consumes this format should just check `deprecatedVersion`.  If `deprecatedVersion` is non-empty, the license is deprecated.  If `deprecatedVersion` is empty, the license is not deprecated.

This protects us from desynchronization like that fixed by cc1e05e8 (added deprecated version, 2017-12-28), because it is now impossible to deprecate a license without recording the deprecated version.

The schema change is a two-liner, and the `src/*` change was generated with:

```console
$ sed -i 's/isDeprecated="true" //' $(git grep -l isDeprecated)
```

which should make review straightforward.

Downstream, this will need spdx/tools changes to get the license/exception-loading code looking at `deprecatedVersion` to generate `isDeprecatedLicenseId` output [here][1] and similar.  This change is easy enough here that, if folks think this change is promising, we should probably block this PR on landing the spdx/tools update.  I propose the following order:

1. Review this PR for the general thrust.  If, for some reason, the license list maintainers feel it is not productive, we can close this without further changes.
2. Update spdx/tools to only consider `deprecatedVersion` when generating `isDeprecatedLicenseId` and similar.
3. Rebase this PR to pick up whatever changes have landed in this repository in the meantime.
4. Merge this PR.

[1]: https://github.com/spdx/license-list-data/blob/c4f1648c4931db3ce071ff4b15e52ef749d68076/json/details/GPL-1.0%2B.json#L2

### DIFF
--- a/schema/ListedLicense.xsd
+++ b/schema/ListedLicense.xsd
@@ -47,7 +47,6 @@
 			<element name="text" type="tns:formattedAltTextType" minOccurs="0" maxOccurs="1"/>
 		</all>
 		<attribute name="licenseId" type="string" use="required" />
-		<attribute name="isDeprecated" type="boolean"/>
 		<attribute name="name" type="string" use="required" />
 		<attribute name="listVersionAdded" type="string"/>
 		<attribute name="deprecatedVersion" type="string"/>
@@ -104,15 +103,6 @@
 				<documentation xml:lang="en">
 					<xhtml:p>
 						<xhtml:code>isOsiApproved</xhtml:code> is true if and only if the license is approved by Open Source Initiative as listed <xhtml:a href="https://opensource.org/licenses">here</xhtml:a>.
-					</xhtml:p>
-				</documentation>
-			</annotation>
-		</attribute>
-		<attribute name="isDeprecated" type="boolean">
-			<annotation>
-				<documentation xml:lang="en">
-					<xhtml:p>
-						<xhtml:code>isDeprecated</xhtml:code> is true if and only if the short identifier has been superseded by another.  See the <xhtml:code>deprecatedVersion</xhtml:code> documentation for more details.
 					</xhtml:p>
 				</documentation>
 			</annotation>
--- a/src/AGPL-1.0.xml
+++ b/src/AGPL-1.0.xml
@@ -2,7 +2,7 @@
 <SPDXLicenseCollection xmlns="http://www.spdx.org/license">
    <license isOsiApproved="false" licenseId="AGPL-1.0"
             name="Affero General Public License v1.0"
-            isDeprecated="true" deprecatedVersion="3.1">
+            deprecatedVersion="3.1">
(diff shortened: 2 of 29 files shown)


wking:
Rebased around #452 with 1a74f3e → 4e1a22b.

wking:
Rebased around #586 and #599 with 4e1a22b → a1a3e68.

goneall:
Update to the tools in progress: https://github.com/spdx/LicenseListPublisher/pull/3

wking:
> Update to the tools in progress: spdx/LicenseListPublisher#3

With that pull request landed, can we move this forward again?  I think we'll need to cut a LicenseListPublisher release, update the `Makefile` here to use that version, and then rebase this PR to resolve the conflicts.

goneall:
@wking I'll cut a new release of LicenseListPublish once we have the FSF Free/Libre issues worked out (which should be relatively soon).

lenora2468B (changes requested):
I am still learning all the steps to this and hope it's ok to copy the path. Thank you.

jlovejoy:
@goneall - can we close this, not sure what status is at this point?

goneall:
@jlovejoy  The tooling has been updated, so we can now apply the change.  It does, however, need to be rebased before merging.

I think the change is a good idea since it would make it easier to maintain the license list - you only need to update the deprecated version and it would prevent inconsistencies such as having a deprecated version and isDeprecated set to false.

@wking added good documentation on how to create the pull request, so if @wking isn't available to rebase, I can generate a new PR sometime over the next couple of weeks.

Let's leave it open for now.

swinslow:
@goneall For this old PR -- if I'm following, it looks like this is (1) deleting the definition of `isDeprecated` in the schema, and then (2) deleting `isDeprecated=` attributes from the handful of licenses that have them.

Does that sound correct? If so, I can create a new PR to handle this. I _think_ there are some now-deprecated licenses that weren't picked up here at the time, so it would likely need a refresh anyway.

goneall:
> @goneall For this old PR -- if I'm following, it looks like this is (1) deleting the definition of isDeprecated in the schema, and then (2) deleting isDeprecated= attributes from the handful of licenses that have them.

> Does that sound correct? If so, I can create a new PR to handle this. I think there are some now-deprecated licenses that weren't picked up here at the time, so it would likely need a refresh anyway.

@swinslow Yes - that is correct - a new PR would be appreciated.

chatchai-hub (approved):

jlovejoy:
@swinslow @goneall - can I close this?

swinslow:
Yup -- I just added #1545 to track the yet-to-be-done task that came out of the discussion here, so this PR can be closed (which I'm doing now).

PR #586, opened 2017-12-29
schema/ListedLicense: Document our XML semantics

### LINKED ISSUE #495: Document the semantics of our XML schema
Spun off from #391, which has some links to current external-to-this-repo docs.  I think we want to bring those docs into this repo so we only have one location to handle as we iterate on the schema.  One thing that could use more detailed docs are the semantics of `<alt>` attributes, like the regexp dialect used by `match` and whether or not `name` needs to be unique within a license.

wking:
Using [annotations][1].  The xhtml namespace handling is from [here][2].  The reworded intro is based on our new `README` text.  The ERE match semantics are from [here][3].  The rest of the text is mine, and in some cases (e.g. the `crossRefs` ordering) may be aspirational and not current.

I still have a number of types to document, but thought I'd put up the WIP PR so folks could see where I was planning to go.

Fixes #495.

[1]: https://www.w3.org/TR/2012/REC-xmlschema11-1-20120405/structures.html#cAnnotations
[2]: https://www.w3.org/TR/2012/REC-xmlschema11-1-20120405/structures.html#eg.import.html
[3]: https://spdx.org/spdx-specification-21-web-version#h.2mjng0vqrghe

### DIFF
--- a/schema/ListedLicense.xsd
+++ b/schema/ListedLicense.xsd
@@ -344,7 +344,7 @@
 				<annotation>
 					<documentation xml:lang="en">
 						<xhtml:p>
-							<xhtml:code>titleText</xhtml:code> is the title of the license/exception.
+							<xhtml:code>titleText</xhtml:code> is the title of the license/exception.  This that the content should be formatted as a header and that it optional (as described in <xhtml:content>optionalType</xhtml:content>).
 						</xhtml:p>
 					</documentation>
 				</annotation>
--- a/schema/ListedLicense.xsd
+++ b/schema/ListedLicense.xsd
@@ -1,22 +1,46 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<schema xmlns="http://www.w3.org/2001/XMLSchema" xmlns:tns="http://www.spdx.org/license" targetNamespace="http://www.spdx.org/license" elementFormDefault="qualified">
+<schema xmlns="http://www.w3.org/2001/XMLSchema" xmlns:xhtml="http://www.w3.org/1999/xhtml" xmlns:tns="http://www.spdx.org/license" targetNamespace="http://www.spdx.org/license" elementFormDefault="qualified">
 	<annotation>
 		<documentation xml:lang="en">
-            XML Schema for the SPDX Listed License input format.  This format is intended for internal use
-            of the SPDX legal team for representing licenses in the github source repository
-            github.com/spdx/listed-license-XML.  A separate format of the same licenses are available
-            for application usage at github.com/spdx/license-list-data.
-            Licensed under CC-BY-4.0.
-        </documentation>
+			<xhtml:h1>SPDX License List XML Schema</xhtml:h1>
+			<xhtml:p>
+				This format is internal to the SPDX legal team and is subject to change.  For stable output formats, see <xhtml:a href="https://github.com/spdx/license-list-data">license-list-data</xhtml:a>.
+			</xhtml:p>
+			<xhtml:p>
+				Licensed under CC-BY-4.0.
+			</xhtml:p>
+		</documentation>
 	</annotation>
-	<element name="SPDXLicenseCollection" type="tns:SPDXLicenseCollectionType"/>
+	<element name="SPDXLicenseCollection" type="tns:SPDXLicenseCollectionType">
+		<annotation>
+			<documentation xml:lang="en">
+				<xhtml:p>
+					<xhtml:code>SPDXLicenseCollection</xhtml:code> is the root element of SPDX license lists.
+				</xhtml:p>
+			</documentation>
+		</annotation>
+	</element>
 	<complexType name="SPDXLicenseCollectionType">
+		<annotation>
+			<documentation xml:lang="en">
+				<xhtml:p>
+					<xhtml:code>SPDXLicenseCollectionType</xhtml:code> collects a set of licenses and/or exceptions.  The order in which child elements are listed is not significant.
+				</xhtml:p>
+			</documentation>
+		</annotation>
 		<choice minOccurs="1" maxOccurs="unbounded">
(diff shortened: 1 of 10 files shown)


wking:
I've pushed be1af7e → 86fe5d8 filling in documentation for the remaining entries.  Now that I think the coverage is complete, I've removed the “WIP” from the PR title.

schema/ListedLicense.xsd
```diff
@@ -29,46 +53,214 @@
...
+							<xhtml:code>notes</xhtml:code> contains unstructured notes about the license.
+						</xhtml:p>
+					</documentation>
+				</annotation>
+			</element>
+			<element name="standardLicenseHeader" type="tns:formattedAltTextType" minOccurs="0" maxOccurs="1">
+				<annotation>
+					<documentation xml:lang="en">
+						<xhtml:p>
+							<xhtml:code>standardLicenseHeader</xhtml:code> is the license steward's recommended text for per-file license grants.
```
goneall: IANAL, but from what I understand the term "grant" has specific legal meaning which may not be applicable to all situations where the license header is used.

Suggest changing "grant" to "notice" or "text".

schema/ListedLicense.xsd
```diff
@@ -29,46 +53,214 @@
...
+					</xhtml:p>
+				</documentation>
+			</annotation>
+		</attribute>
 	</complexType>
 	<complexType name="crossRefsType">
+		<annotation>
+			<documentation xml:lang="en">
+				<xhtml:p>
+					<xhtml:code>crossRefsType</xhtml:code> is a set of URIs for the license or exception.  When the license steward has a URI devoted to the license, it must be the first entry in <xhtml:code>crossRefsType</xhtml:code>.  The order in which the other child elements are listed is not significant.
```
goneall: Is the order of cross references agreed to and consistent?  If not, I would just leave it as a set of URI's for the license or exceptions.

schema/ListedLicense.xsd
```diff
@@ -29,46 +53,214 @@
...
 		</sequence>
 	</complexType>
 	<complexType name="altType" mixed="true">
-		<group ref="tns:formattedFixedTextGroup" minOccurs="0" maxOccurs="unbounded"/>
-		<attribute name="name" type="string" use="required" />
-		<attribute name="match" type="string" use="required" />
+		<annotation>
+			<documentation xml:lang="en">
+				<xhtml:p>
+					<xhtml:code>altType</xhtml:code> represents alternative content, where the SPDX considers several alternatives legally equivalent.
```
goneall: This may be OK, but in practice we have a lot of `.+` matches which really could include text not legally equivalent.  Should we include something with a little more "wiggle room" in the language?  For example - "represents alternative content commonly found in the license text, where the SPDX considers typical usage of the alternative text legally equivalent."

@jlovejoy thoughts?

schema/ListedLicense.xsd
```diff
@@ -80,14 +272,54 @@
...
+							<xhtml:code>br</xhtml:code> is a line break with the same semantics as in <xhtml:a href="https://www.w3.org/TR/html52/textlevel-semantics.html#the-br-element">HTML</xhtml:a>.
+						</xhtml:p>
+					</documentation>
+				</annotation>
+			</element>
+			<element name="titleText" type="tns:formattedAltTextType" minOccurs="0" maxOccurs="1">
+				<annotation>
+					<documentation xml:lang="en">
+						<xhtml:p>
+							<xhtml:code>titleText</xhtml:code> is the title of the license/exception.
```
goneall: Do we want to add the titleText is always considered optional per the license matching guidelines?

schema/ListedLicense.xsd
wking: > … from what I understand the term "grant" has specific legal meaning which may not be applicable to all situations where the license header is used.

What's the distinction between “grant” and “notice”?  Maybe “Copyright 2018 me.  All rights reserved” would be a notice but not a grant?  That's fine, but since all of these are licenses, I expect all of our suggested headers will be license grants (although those license grants may also include copyright notices).

schema/ListedLicense.xsd
wking: > Is the order of cross references agreed to and consistent?

It seems to be pretty close.  I've rebased this PR onto master and added a new commit fixing the cases that looked out of order with 5e3aa52 → 7d44198.

schema/ListedLicense.xsd
wking: > This may be OK, but in practice we have a lot of `.+` matches which really could include text not legally equivalent.

I've added a warning about this in 7d44198 → deec602.

schema/ListedLicense.xsd
wking: > Do we want to add the `titleText` is always considered optional per the license matching guidelines?

I don't think we need to call that out, since I'd rather move towards having distinct markup for optionality (`<alt>` and `<optional>`) and semantic meaning (e.g. `<h1>`).  It's my understanding that spdx/tools doesn't treat `<titleText>` as optional at the moment, is that the case?

wking:
> One other general recommendation would be to reference the license matching guidelines that the schema is based on somewhere in the schema annotation.

This license is the canonical source of the schema, and the canonical source of the XML which consumes that schema.  I'd rather have it be the canonical source of the schema semantics as well, in which case there is no reason to reference external matching guidelines.  I'd rather have the external matching guidelines reference *this* documentation as the canonical source.

schema/ListedLicense.xsd
goneall: I just checked and the tools to convert the title to optional

wking: Ok, I've pushed deec602 → 48cf583 which mentions both the title-ness and optional-ness.  I've filed #607 to track my proposed change.

PR #587, opened 2017-12-30
u-boot-exception diverges from original text

sam-ellis:
u-boot-exception diverges from original text. License list text says:
"The U-Boot License Exception:"

Where as original (http://git.denx.de/?p=u-boot.git;a=blob;f=Licenses/Exceptions) says:
"GPL License Exception:"

Google shows no matches for the text that is on the license list, except for the license list itself. Propose to change license list to match original.

### DIFF
--- a/src/exceptions/u-boot-exception-2.0.xml
+++ b/src/exceptions/u-boot-exception-2.0.xml
@@ -6,7 +6,7 @@
       </crossRefs>
       <notes>Typically used with GPL-2.0+</notes>
       <titleText>
-         <p>The U-Boot License Exception:</p>
+         <p>GPL License Exception:</p>
       </titleText>
       <p>Even though U-Boot in general is covered by the GPL-2.0/GPL-2.0+,
          this does *not* cover the so-called "standalone" applications
--- a/src/exceptions/u-boot-exception-2.0.xml
+++ b/src/exceptions/u-boot-exception-2.0.xml
@@ -6,7 +6,7 @@
       </crossRefs>
       <notes>Typically used with GPL-2.0+</notes>
       <titleText>
-         <p>GPL License Exception:</p>
+         <p><alt name="title" match="GPL|The U-Boot">GPL</alt> License Exception:</p>
       </titleText>
       <p>Even though U-Boot in general is covered by the GPL-2.0/GPL-2.0+,
          this does *not* cover the so-called "standalone" applications


src/exceptions/u-boot-exception-2.0.xml
```diff
@@ -6,7 +6,7 @@
       </crossRefs>
       <notes>Typically used with GPL-2.0+</notes>
       <titleText>
-         <p>The U-Boot License Exception:</p>
+         <p>GPL License Exception:</p>
```
wking: I think we need to support the old form for compatibility with our 3.0 release, [2.6 release][1], and possibly earlier releases.  But I agree that our canonical text should match upstream.  That would mean markup like:

```xml
<p><alt name="title" match="GPL|The U-Boot">GPL</alt> License Exception:</p>
```

[1]: https://github.com/spdx/license-list-data/blob/v2.6/rdfa/u-boot-exception-2.0.html#L160

jlovejoy:
I am not as picky about titles as they don't count for matching. In this case, "GPL License Exception" is not a good name, as many of the exceptions are for GPL - this would not be differentiating. Considering that this is specifically for U-boot as per the exception text, that's probably why that title was used. 
I'd be curious to hear from scanning folks if this causes problems?  @pombredanne @goneall @sam-ellis

wking:
On Thu, Jan 11, 2018 at 08:26:21PM +0000, Jilayne Lovejoy wrote:
> I am not as picky about titles as they don't count for matching.

But what's a title?  If your scanner sees “GPL License Exception” at
the top of a file, I think we need matching markup so the scanner can
decide between “this is an insignificant (title) change” and “this is
a significant change to the legal content”.  For example, you don't
want your scanner to ignore:

  The give me $1 every time you compile this code license.  In
  addition to the terms discussed in the GPL, by compiling this code
  you agree to give me $1 for every compilation.

  Even though U-Boot in general is covered by the GPL-2.0/GPL-2.0+…

by considering that whole first paragraph as an insignificant title
change.

goneall:
@jlovejoy As long as the title variant is still considered a title and can be ignored, I would recommend merging this change to allow the older versions to be matched.

Here's a bit more detail:

The way the current SPDX Tools license matching works is the title text is treated as an `<optional>` block.  If the text inside the optional block matches the text of the title found in the license text, it is ignored.  If it doesn't match, it will be flagged as a difference.  Adding the additional `<alt...>` in this PR would allow for matching against the previous license texts.

The reason we took this approach as opposed to a more general `<alt...>` approach is the reasons @wking mentions above - we don't want to accidentally match significant text.

jlovejoy:
discussed on Jan 25 legal call: decided to add <alt> tag to enable matching on either title option.
@wking or @sam-ellis - could one of you update this PR with that and then I'll merge?

wking:
On Thu, Jan 25, 2018 at 05:36:07PM +0000, Jilayne Lovejoy wrote:
> @wking or @sam-ellis - could one of you update this PR with that and
> then I'll merge?

I can't push to this repo, but my recommended markup is in [1].

[1]: https://github.com/spdx/license-list-XML/pull/587#discussion_r159124542

bradleeedmondson:
although the SPDX title "The U-Boot License Exception" wasn't found on the interwebs, we decided on the call to continue to match to it because it was released that way on the 3.0 license list (and folks may therefore be using that text on non-public repos).

This updates the title to match either "The U-Boot License Exception:" or the upstream's "GPL License Exception:"

PR #588, opened 2017-12-30
Add missing close quote to Apache-1.0.xml

sam-ellis:
Added missing close quote character relative to original text at http://www.apache.org/licenses/LICENSE-1.0.

### DIFF
--- a/src/Apache-1.0.xml
+++ b/src/Apache-1.0.xml
@@ -52,7 +52,7 @@
             <bullet>6.</bullet>
           Redistributions of any form whatsoever must retain the following acknowledgment:
           <p>"This product includes software developed by <alt match=".+" name="nameClause6">the 
-          Apache Group for use in the Apache HTTP server project (http://www.apache.org/)</alt>.</p>
+          Apache Group for use in the Apache HTTP server project (http://www.apache.org/)</alt>."</p>
         </item>
       </list>
       <p>THIS SOFTWARE IS PROVIDED BY 


src/Apache-1.0.xml
```diff
@@ -52,7 +52,7 @@
             <bullet>6.</bullet>
           Redistributions of any form whatsoever must retain the following acknowledgment:
           <p>"This product includes software developed by <alt match=".+" name="nameClause6">the 
-          Apache Group for use in the Apache HTTP server project (http://www.apache.org/)</alt>.</p>
+          Apache Group for use in the Apache HTTP server project (http://www.apache.org/)</alt>."</p>
```
wking: Confirming this closing quote upstream:

```console
$ curl -s http://www.apache.org/licenses/LICENSE-1.0 | grep 'apache.org/)'
 *    for use in the Apache HTTP server project (http://www.apache.org/)."
 *    for use in the Apache HTTP server project (http://www.apache.org/)."
```

But in order to maintain backward compat with our 3.0 release, [2.6 release][1], and possibly earlier releases, I think we need to use:

```xml
…</alt>.<optional>"</optional></p>
```
[1]: https://github.com/spdx/license-list-data/blame/v2.6/rdfa/Apache-1.0.html#L174-L175

jlovejoy: considering no one has had an issue with the missing quote until now, we have good representation of tool/scanner makers in this workgroup, and it's in the original license - I don't think the optional tag is necessary.

PR #289, opened 2016-03-30
Automated PR for BitTorrent-1.0

myndzi:

PR #589, opened 2017-12-30
Remove section 4.g. from BitTorrent-1.0 license.

sam-ellis:
BitTorrent-1.0 on license-list has a section 4.g. that does not appear in original license. This section did not appear in pre-XML license-list, so possibly is an error introduced during conversion. See link here http://sources.gentoo.org/cgi-bin/viewvc.cgi/gentoo-x86/licenses/BitTorrent?r1=1.1&amp;r2=1.1.1.1&amp;diff_format=s and I have also downloaded an old BitTorrent source archive and confirmed included LICENSE.txt does not include this clause (http://web.archive.org/web/20090309024144/http://download.bittorrent.com:80/dl/archive/BitTorrent-4.0.0.tar.gz). Propose to remove clause 4.g.

### DIFF
--- a/src/BitTorrent-1.0.xml
+++ b/src/BitTorrent-1.0.xml
@@ -260,28 +260,6 @@
                this License are fulfilled for those portions of the Derivative Works that consist of the
                Licensed Product or any Modifications thereto.
           </item>
-               <item>
-                  <bullet>g.</bullet>
-            Compensation for Distribution of Executable Versions of Licensed Products, Modifications 
-            or Derivative Works. Notwithstanding any provision of this License to the contrary, by distributing, 
-            selling, licensing, sublicensing or otherwise making available any Licensed Product, or Modification 
-            or Derivative Work thereof, you and Licensor hereby acknowledge and agree that you may sell, license 
-            or sublicense for a fee, accept donations or otherwise receive compensation for executable versions 
-            of a Licensed Product, without paying a royalty or other fee to the Licensor or any other Contributor, 
-            provided that such executable versions (i) contain your or another Contributor's material Modifications, 
-            or (ii) are otherwise material Derivative Works. For purposes of this License, an executable version of 
-            the Licensed Product will be deemed to contain a material Modification, or will otherwise be deemed a 
-            material Derivative Work, if (a) the Licensed Product is modified with your own or a third party's software 
-            programs or other code, and/or the Licensed Product is combined with a number of your own or a third party's 
-            software programs or code, respectively, and (b) such software programs or code add or contribute material 
-            value, functionality or features to the License Product. For the avoidance of doubt, to the extent your 
-            executable version of a Licensed Product does not contain your or another Contributor's material Modifications 
-            or is otherwise not a material Derivative Work, in each case as contemplated herein, you may not sell, license 
-            or sublicense for a fee, accept donations or otherwise receive compensation for such executable. Additionally, 
-            without limitation of the foregoing and notwithstanding any provision of this License to the contrary, you 
-            cannot charge for, sell, license or sublicense for a fee, accept donations or otherwise receive compensation 
-            for the Source Code.
-          </item>
             </list>
         </item>
         <item>


wking:
This is rolling back content explicitly added in adffdb0a (Update BitTorrent-1.0.xml, 2016-05-06, #289), but I haven't had time to dig up the motivation for that change.

wking:
Here's [*v1.1*][1] from 2007-03-10 with the 4.g.  I'm still trying to find the time range for 1.0.

[1]: http://web.archive.org/web/20070310221212/http://www.bittorrent.com/bittorrent-open-source-license

wking:
[Here][1] is v1.0 from 2005-02-09 with no 4.g.  Still no 4.g by [2005-11-24][2], and by [2005-12-12 they were showing v1.1][3].  So my current impression is that adffdb0 accidentally injected v1.1 text into v1.0, and that we want to remove it here.  It would be good to get @jlovejoy's oppinion though.

If removal is accepted, we need to decide how to recover from including the extra text in our v3.  Because nobody is likely to have copied our text into a new project in the last few days and because the extra text looks legally significant, I like removing it as you do, withough a backwards-compat `<alt>`.  That will fix v3.1.  I'm not sure what we can do about v3.0.  Do we publish errata/corrigenda for our releases?  It would be good to have a channel so tools could discover issues like this.  But even then, I can't think of a good way for a downstream consumer of a 3.0 `Bittorrent-1.0` to tell if the original license-expression author was aware of the error or not.

[1]: http://web.archive.org/web/20050209224501/http://www.bittorrent.com/license/
[2]: http://web.archive.org/web/20051124014027/http://www.bittorrent.com:80/license/
[3]: http://web.archive.org/web/20051212035130/http://www.bittorrent.com:80/license/

wking:
[Here][1] is v1.0 from 2005-02-09 with no 4.g.  Still no 4.g by [2005-11-24][2], and by [2005-12-12 they were showing v1.1][3].  So my current impression is that adffdb0 accidentally injected v1.1 text into v1.0, and that we want to remove it here.  It would be good to get @jlovejoy's oppinion though.

If removal is accepted, we need to decide how to recover from including the extra text in our v3.  Because nobody is likely to have copied our text into a new project in the last few days and because the extra text looks legally significant, I like removing it as you do, withough a backwards-compat `<alt>`.  That will fix v3.1.  I'm not sure what we can do about v3.0.  Do we publish errata/corrigenda for our releases?  It would be good to have a channel so tools could discover issues like this.  But even then, I can't think of a good way for a downstream consumer of a 3.0 `Bittorrent-1.0` to tell if the original license-expression author was aware of the error or not.

[1]: http://web.archive.org/web/20050209224501/http://www.bittorrent.com/license/
[2]: http://web.archive.org/web/20051124014027/http://www.bittorrent.com:80/license/
[3]: http://web.archive.org/web/20051212035130/http://www.bittorrent.com:80/license/

bradleeedmondson:
> So my current impression is that adffdb0 accidentally injected v1.1 text into v1.0, and that we want to remove it here.

+1 for this as immediate next step (fixing for 3.1)

For discoverably documenting errata e.g. for 3.0, I would say we should open a separate issue.

jlovejoy:
ugh, yes, agreed this was a mistake on my part. In the past, if we ever had something like this, I think we simply fixed it, noted it in the comments of the old spreadsheet and got on with it. No one ever questioned that ;) but I don't really recall a similar example.  
I think this can simply be merged (and thanks @sam-ellis  for catching!) but not sure about the errata reporting aspect. (or if anyone will actually notice and it's worth worrying about too much?)

wking:
On Thu, Jan 11, 2018 at 08:39:28PM +0000, Jilayne Lovejoy wrote:
> (or if anyone will actually notice and it's worth worrying about too
> much?)

I'm not sure if folks will notice or not, especially since
BitTorrent-1.0 is a fairly peripheral license without a large active
userbase (or any active userbase?).  In that respect, I think this a
pretty lucky thing to break, so we have time to setup a plan before we
break something important (like a modern GPL), which will certainly
happen at some point.

I think docs that say “these identifiers are from the 3.0 license
list” (e.g. like Cargo's recent 2.4 claim, rust-lang/cargo#4898, but
for 3.0) will get pretty hairy if we have this sort of issue with a
widely-used license.  An machine-readable errata channel would make it
as easy as possible for consumers to hear about these issues and
change their docs (e.g. to a 3.1) as quickly as possible.

jlovejoy:
as this PR has to do with the fix and that clearly needs to be fixed, I'm merging. but will open another issue and tag "for discussion" as to how to track errors like this that impact the actual text of the license (we shouldn't have many...)

PR #590, opened 2017-12-30
Modify gSOAP-1.3b to cope with "shall" versus "will"

sam-ellis:
Cross referenced license at http://www.cs.fsu.edu/~engelen/license.html uses the word "shall" in section 3.2 instead of "will". Propose to amend XML to cope with either.


bradleeedmondson:
Looks good to me; merging

PR #591, opened 2018-01-01
Test pull request in continuous integration process

goneall:
This should only run tests and not push any results back to license-list-data


goneall:
Temporarily closing PR (again)

wking:
> Temporarily closing PR (again)

Why are you closing these and then opening new ones?  Maybe we can find a way to keep this in a single PR.

goneall:
This was opened in error - is was supposed to be a PR against the goneall clone.  See PR #592 for the reasons they were closed.

PR #593, opened 2018-01-02
Makefile: Use Travis to run canonical-matching tests

wking:
Based on Gary's repository for now, because spdx/license-test-files#6 is still in flight.  Ideally the test data would have signed releases, so we could use `--branch v1.2.3` in the clone call.  But we don't have that yet, so just chase master.

This is my preferred alternative to #592, with an external test repository and the publication logic removed for now.


wking:
The [test failure][1] will be fixed by #589.

[1]: https://travis-ci.org/spdx/license-list-XML/jobs/324411868#L482

wking:
I've pushed e15a8e2 → 992d62e, temporarily rolling back the test data to get the tests passing here.  With that change, this PR can land without waiting for #589.  If this lands first, we'll have to update #589 to bump the test-data hash in `.travis.yml`.  If #589 lands first, I'll update this PR to bump the test-data hash in `.travis.yml`.

goneall:
We should commit both the update to the license list XML file and the test in the same commit to resolve.  I'll revert the copy in https://github.com/goneall/license-list-XML to resolve the test failures in the script.

wking:
I've pushed 992d62e → 8354d14 updating this to use `gpg` to validate the JAR.

> I'll revert the copy in https://github.com/goneall/license-list-XML to resolve the test failures in the script.

No need to revert, because I'm now [pinning the commit][1] in `.travis.yml`.  More discussion in the 992d62efd82e47ee3528971f1d84cfda21de1763 commit message.

[1]: https://github.com/spdx/license-list-XML/pull/593/files#diff-354f30a63fb0907d4ad57269548329e3R6

wking:
I've been unable to convince @goneall that keeping the test data in an external repository will be easier (e.g. see [here][1]), so I've just pushed 268f54b → 1853d30 to demonstrate the submodule approach I'd floated [here][2].  This would give us a similar feel to having local files (for folks who want to `git submodule update --init test`) but would allow others to manipulate this repository without bothering to checkout the test data.  It also gives us a single place to review changes to the test data, avoiding the double-review of #592 and spdx/license-test-files#6 adding many similar files with manual synchronization.  Instead, license-list-XML reviewers will have that condensed down to submodule commit bumps.

If this is *still* not acceptable, let me know, and I'll file another commit to go completely over to the local-file approach.  In that case, I expect we'll revisit the external repo approaches once we get tired of the doubled review.

[1]: https://github.com/spdx/license-list-data/pull/22#issuecomment-355066522
[2]: https://github.com/spdx/license-list-data/pull/22#issuecomment-355073832

wking:
I've been unable to convince @goneall to use either the `.travis.yml` clone/checkout from 992d62ef or the Git submodule approach from 1853d30e6 (some recent discussion in spdx/license-list-data#22), so I've just pushed 1853d30e6 → 9a8648f08 dropping the submodule in favor of a `read-tree` checkout.  The only remaining difference between this PR and #592 are:

* The `Makefile` (here) vs. script (#592) approaches.
* Using `gpg` (here) vs. `sha256sum` (#592) for verifying the downloaded tool JAR.

wking:
The current tool failure is because spdx/tools was broken by wking/fsf-api#9 (discussion on protecting from future breakage in wking/fsf-api#10).  It would be nice if the validation didn't depend on the presence of the FSF API, since as far as I can see that's orthogonal to whether we match the canonical text or not.  But I realize that separating the verification from the generation would take spdx/tools dev-time.

Oh, another difference between here and #592 is that I'm [still][1] punting on auto-publishing to space out the changes.  This PR is just about unit-testing this repo.

[1]: https://github.com/spdx/license-list-XML/pull/593#issue-285565813

goneall:
>  But I realize that separating the verification from the generation would take spdx/tools dev-time.

Also, once we implement publishing we would want to include the FSF-Free generation.

Once versioning is implemented in the FSF Free API code, I'll update the tools and push a new version to remove the error.  I'll also add a property to use a cached version of the FSF Free JSON file so we don't need to update the tool if we run into a similar issue in the future.

goneall:
Makefile and gpg are improvements

We need to resolve the FSF-Free failures before merging.

I would also like to have support for publishing, but I can hold off on that if we can add publishing support within the next couple of weeks.

wking:
On Fri, Jan 05, 2018 at 07:37:16PM +0000, goneall wrote:
> > But I realize that separating the verification from the generation
> > would take spdx/tools dev-time.
>
> Also, once we implement publishing we would want to include the
> FSF-Free generation.

Agreed.  But I'd rather have the option to verify without publishing.

> Once versioning is implemented in the FSF Free API code, I'll update
> the tools and push a new version to remove the error.

Until then, since we don't need working FSF data for verification,
I've switched in temporary dummy data with 845b02c, but it doesn't
work (as described in [1]).

> I'll also add a property to use a cached version of the FSF Free
> JSON file so we don't need to update the tool if we run into a
> similar issue in the future.

Can we get that now?  Then we can move forward with this and buy me
some more time to work out [2].

[1]: https://github.com/wking/fsf-api/pull/9#issuecomment-355389227
[2]: https://github.com/wking/fsf-api/issues/10#issuecomment-355431644

goneall:
@wking The master branch of the tools are updated with the property to force using the cached JSON file.  It has also been updated to comply with the schema changes for PR #452 

Do you think you will have wking/fsf-api#9 resolved soon or should I spin a release with these 2 changes?

goneall:
Note - test file updates are required for Apache-1.0.txt and LLVM-exception.txt to pass tests due to changes made since the 3.0 release of the license list.

goneall:
I just pushed version 2.1.9 of the tools.  To use the file in the resources directory, add -DLocalFsfFreeJson=true to the Java command line. See the update to the goneall/license-list-XML repo for a reference: https://github.com/goneall/license-list-XML/commit/a4dc1f13ddaeb8b209e8051e53cdd2c950b7a0a4

wking:
I've pushed 845b02cd → 2453fb38, rebasing onto master, pulling in v2.1.9 of the tools, and bumping the tests (including goneall/license-test-files#2).  Now Travis is happy with this PR :).

PR #596, opened 2018-01-10
Add crossref for Zimbra-1.3

mlinksva:
The only license without a crossref.

Note wayback machine copies used for crossrefs for 4 other licenses.


bradleeedmondson:
Looks good, thanks.

PR #597, opened 2018-01-11
AGPL-3.0: Drop redundant release-date <notes>

wking:
The XSD suggests (buggily) that we only want a single <notes> element (more details on this in #577).  And the information contained in the `<notes>` section I'm removing is redundant, because the `titleText` contains the release date.  I don't see a need to include that release date as unstrucutured information in two places, and the license title is clearly the more important location.

This is the second per-license change spun out of #566 (based on @jlovejoy's request for [individual review][1]).

[1]: https://github.com/spdx/license-list-XML/pull/566#issuecomment-354372728


wking:
This change happened in #566, so we no longer need a per-license PR.

PR #600, opened 2018-01-15
OSL-2.0: Fix “Licensev.” → “License v.”

wking:
The typo came in with the license in 0162ea9a (Automated, 2016-03-30, #180).

Also add an optional “The ” to match the archived original:

```console
$ curl -s https://web.archive.org/web/20041020171434/http://www.rosenlaw.com/osl2.0.html | grep 'Open Software License' | head -n2
<!-- End Wayback Rewrite JS Include --><h1>The Open Software License<br/> v. 2.0</h1>
<p>This Open Software License (the "License") …
```

### DIFF
--- a/src/OSL-2.0.xml
+++ b/src/OSL-2.0.xml
@@ -7,7 +7,7 @@
       <standardLicenseHeader>Licensed under the Open Software License version 2.0</standardLicenseHeader>
     <text>
       <titleText>
-         <p>Open Software Licensev. 2.0</p>
+         <p><optional>The </optional>Open Software License v. 2.0</p>
       </titleText>
     
       <p>This Open Software License (the "License") applies to any original work of authorship (the
--- a/test/simpleTestForGenerator/OSL-2.0.txt
+++ b/test/simpleTestForGenerator/OSL-2.0.txt
@@ -1,4 +1,4 @@
-Open Software Licensev. 2.0
+Open Software License v. 2.0
 
 This Open Software License (the "License") applies to any original work of authorship (the "Original Work") whose owner (the "Licensor") has placed the following notice immediately following the copyright notice for the Original Work:
 


wking:
Rebased around #593 with 7f59223 → 46070f1.

jlovejoy:
the title - in terms of no "the" and using "v" instead of "version" follows the guidelines we've used since day 1 for all license titles: https://spdx.org/spdx-license-list/license-list-overview#fields
while I understand now (as per discussion on today's legal call with @goneall ) that this could cause issues for tools/matching - I'm not sure it has so far?
we should fix the missing space as that is purely a typo introduced recently, but I don't think we need to add "the" - if we do that here, then we have a bunch more we may need to do that for. To what value does this add? (again a question for the tool vendors)

wking:
> while I understand now (as per discussion on today's legal call with @goneall ) that this could cause issues for tools/matching - I'm not sure it has so far?

In master, running the matching tests which landed with #593:

```console
$ git describe
v3.0-38-gaaf585c
$ sed -i 's/Licensev/License v/g' test/simpleTestForGenerator/OSL-2.0.txt
$ make TEST_DATA=test/simpleTestForGenerator validate-canonical-match
…
The following warning(s) were identified:
        Test for license OSL-2.0 failed: Test for license ID OSL-2.0 failed due to difference found Normal text of license does not match starting at line #1 column #5 "software" when comparing to template text "
This Open Software License (the "License") applies to any original work of authorship (the "Origina...".  Last optional text was not found due to the optional difference: 
        Normal text of license does not match starting at line #1 column #22 "v" when comparing to template text " Open Software Licensev. 2.0"
Completed processing licenses
…
```

so unless we add the space to the title, our own test suite will fail to match license text that correctly includes the space.  I think that's enough for “has caused an issue with tools/matching”.

> … I don't think we need to add "the" - if we do that here, then we have a bunch more we may need to do that for. To what value does this add? (again a question for the tool vendors)

This has the same spdx/tools matching issues as the space above.  If we want to match both the upstream text I linked in the commit message (which includes “The”) and our traditional SPDX text (which does not) we need to use either an `<optional>` or `<alt>` around the optional “The”.  I think we *do* want to match both the upstream text and our own previous text.

goneall:
Since the changes will help the matching, I'll go ahead and merge the changes.

PR #603, opened 2018-01-26
BitTorrent-1.1: Move standardLicenseHeader inside text and support 1.1 grants

wking:
Builds on #581; review that first.

Allow matching for folks who fix the referenced license version from “1.0” → “1.1”.  Make “1.1” our canonical text to reduce future reader confusion.


wking:
Rebased now that #581 has landed.

goneall:
Adding legal review for allowing the different versions to match in the license text and license header.

jlovejoy:
discussed on March 22 legal call.

PR #609, opened 2018-02-05
Sendmail: Drop page-end 139848.1 from the canonical text

wking:
But continue to match it via an `<alt>`, in case anyone had copied our old, previously-canonical text.

In [the upstream PDF][1] (linked with [a crossRef to the archive.org URI][2]) this text is part of the footer on every page.  We've had it in this repo since ef01296f (#212), but I imagine this is a copy/paste issue dating back to the original SPDX version of this license (sometime [around 2015-09-16][3]).

[1]: https://web.archive.org/web/20160322142305/https://www.sendmail.com/pdfs/open_source/sendmail_license.pdf
[2]: https://github.com/spdx/license-list-XML/blob/v3.0/src/Sendmail.xml#L6
[3]: https://lists.spdx.org/pipermail/spdx-legal/2015-September/001497.html

### DIFF
--- a/src/Sendmail.xml
+++ b/src/Sendmail.xml
@@ -81,7 +81,7 @@
         <item>
             <bullet>(i)</bullet>
           Redistributions of source code must retain the above copyright notice, this list of conditions
-             and the following disclaimer. 139848.1
+             and the following disclaimer.
         </item>
         <item>
             <bullet>(ii)</bullet>
--- a/test/simpleTestForGenerator/Sendmail.txt
+++ b/test/simpleTestForGenerator/Sendmail.txt
@@ -25,7 +25,7 @@ Use, Modification and Redistribution (including distribution of any modified or 
 
      (b) Redistribution and use in source and binary forms, with or without modification, are permitted provided that the following conditions are met: 
 
-          (i) Redistributions of source code must retain the above copyright notice, this list of conditions and the following disclaimer. 139848.1 
+          (i) Redistributions of source code must retain the above copyright notice, this list of conditions and the following disclaimer.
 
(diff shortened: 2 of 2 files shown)


jlovejoy:
discussed on legal call, March 22 - we agree that instead of alt tag, the number can be removed as it shouldn't have been included to begin with. Can you revise this pull request?

wking:
> we agree that instead of alt tag, the number can be removed as it shouldn't have been included to begin with.

Just to confirm, this will mean we will no longer match anyone who uses a licenses copied from us since we picked up the license in 2015-09-16.  And that the set of matching license from before and after this change are completely disjoint.  That's what you want?

bradleeedmondson:
I think this is a good point to be explicit about, but I also think that
this is what we want.

The older match will still work against older versions of the SPDX License
List, but we're saying with this change that those older versions had an
error which we're correcting for 3.1 and going forward.

wking:
On Thu, Mar 22, 2018 at 10:45:31AM -0700, Bradlee H. Edmndson wrote:
> The older match will still work against older versions of the SPDX
> License List, but we're saying with this change that those older
> versions had an error which we're correcting for 3.1 and going
> forward.

Both approaches (alt or not) correct the error in our canonical
version.  The difference is whether we will continue to match the
previously-canonical form or not.  The only reasons I can think of for
breaking old-form matches are:

* If we think the old-form had a legally-significant difference from
  the correct form.  But the page-end ID seems like an obvious typo,
  so I doubt it would have legal significance.

* Whether there were any consumers of the old form, either for
  matching or copy/pasting into their own project.  In this case, I
  expect copy/pasting from us to be unlikely, but I'm somewhat
  surprised that nobody has ever used our template for matching.  Do
  the maintainers feel that is really the case?

Anyhow, I've dropped the alt with 3924529 → 356abb2.

jlovejoy:
correct. merging now :)

PR #84, opened 2016-03-30
Automated PR for GPL-2.0

myndzi:

PR #614, opened 2018-02-16
src/GPL-2.0*: Add paragraph between “USA” and “Also add information”

wking:
The missing paragraph dates back to 6045f28a (#84) in this repository.  `GPL-2.0-or-later` had this typo fixed in #581.


wking:
I recommend milestoning this [3.1 release][1].

[1]: https://github.com/spdx/license-list-XML/milestone/3

jlovejoy:
while this doesn't matter for matching purposes, agree we should have new paragraph for aesthetic purpose

Issue #620, opened 2018-03-22
New License/Exception Request: Linux-OpenIB


kestewart:
**Provide a proposed Full Name for the license or exception.**
Linux Kernel Variant of OpenIB.org license

**Provide a proposed Short Identifier.**
Linux-OpenIB

**Provide a functioning url reference to the license or exception text, either from the author or a community recognized source.**

[https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/drivers/infiniband/core/sa.h](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/drivers/infiniband/core/sa.h)

**Create and attach a text file with the license or exception text from the url provided in #3. Please proofread the text file to ensure that:**

- **Information has not been lost or modified.**
- **Formatting is clean and consistent with the license or exception URL.**

>    OpenIB.org BSD license below:
> 
>     Redistribution and use in source and binary forms, with or
>     without modification, are permitted provided that the following
>     conditions are met:
>
>      - Redistributions of source code must retain the above
>        copyright notice, this list of conditions and the following
>        disclaimer.
>
>      - Redistributions in binary form must reproduce the above
>        copyright notice, this list of conditions and the following
>        disclaimer in the documentation and/or other materials
>        provided with the distribution.
> 
>    THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
>    EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
>    MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
>    NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS
>    BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN
>    ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
>    CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
>    SOFTWARE.

**Indicate whether the license is OSI-approved (see: http://www.opensource.org/licenses/alphabetical) or whether it has been submitted for approval to the OSI and is currently under review.**

No

**Provide a short explanation regarding the need for this license or exception to be included on the SPDX License List, including identifying at least one program that uses this license.**

[...]

PR #621, opened 2018-03-22
Linux-OpenIB: Add new license to add #620

swinslow:
Adding XML and test .txt file to add #620 

Two comments / questions: 
1. This is my first time contributing to the license list; someone may want to take a look at these and confirm I'm on-track.
2. The crossRef is to the current header file in Linus's trunk of the Linux kernel. Do we want to point to a specific commit, in case this file changes or is removed in the future?


jlovejoy:
discussed on March 22 legal call - accept

Issue #542, opened 2017-12-21
add GNU licenses with new identifiers (*-only and *-or-later)


jlovejoy:
* GPL-1.0-only or GPL-1.0-or-later 
* GPL-2.0-only or GPL-2.0-or-later
* GPL-3.0-only or GPL-3.0-or-later

* LGPL-2.0-only or LGPL-2.0-or-later
* LGPL-2.1-only or LGPL-2.1-or-later
* LGPL-3.0-only or LGPL-3.0-or-later

* FDL-1.1-only or FDL-1.1-or-later
* FDL-1.2-only or FDL-1.2-or-later
* FDL-1.3-only or FDL-1.3-or-later

* AGPL-3.0-only or AGPL-3.0-or-later

license text is same for each set
ensure standard header reflect correct wording for each one

zvr:
The FDL ones are actually GFDL ;)

zvr:
I assume that this does not affect the `isOsiApproved` attribute...

I mean, if GPL-2.0 was approved, then GPL-2.0-only and GPL-2.0-or-later are also marked as approved.

zvr:
Submitted #553

wking:
Why is AGPL-1.0 not receiving the same treatment?

wking:
I still think this explicit-license-list-entry approach (vs. versioning operators) is going to make it [difficult to GPL proxy declarations in the future][1].  I realize that this is the approach adopted by the legal team after lots of discussion, but thought I'd add a final mention here in case there's a last-minute change of heart or anything ;).

[1]: https://lists.spdx.org/pipermail/spdx-legal/2017-November/002309.html

wking:
With #553 merged, what's left here?  Just [the AGPL-1.0 handling][1]?

[1]: https://github.com/spdx/license-list-XML/issues/542#issuecomment-354163002

zvr:
As far as I can tell, AGPL-1.0 was a license by Affero, not by the Free Software Foundation and they never raised an issue regarding "or later".

wking:
On Thu, Dec 28, 2017 at 11:22:47AM +0000, Alexios Zavras (zvr) wrote:
> As far as I can tell, AGPL-1.0 was a license by Affero, not by the
> Free Software Foundation…

Right.

> … and they never raised an issue regarding "or later".

Is our use of -only and -or-later license names really as narrow as
“we only use these if the steward complains about version operators”?
[...]

PR #624, opened 2018-03-22
Add Generic (1.0-2.5) Unported (3.0) or International (4.0) to CC license names

### LINKED ISSUE #618: Issue with CC BY-SA license identifiers
Issue raised by Bradley Kuhn on the spdx-legal list:

> I'm helping Behan Webster's E-ALE project properly describe and document the
> license of their project in their repository.  (E-ALE is a large tutorial
> project -- with materials pulled together from various sources -- and they
> are primarily licensed under various versions of CC BY-SA.)
> 
> Of course, being a good SPDX citizen, I was going to use SPDX identifiers
> where appropriate to describe the licenses.  But I'd never looked closely
> (before today) at the identifiers available for CC licenses in SPDX.
> 
> I noticed two key problems when I did:
> 
>  * To describe the pre-4.0 versions of CC BY-SA, there are only identifiers
>    for the "Unported" versions.  So, the SPDX identifier cannot actually be
>    used to replace situations where someone, say, has used the "CC BY-SA 3.0
>    United States" license.
> 
>  * The Full Names of the licenses on https://spdx.org/licenses/ are
>    incomplete.   Most say something like:
>       "Creative Commons Attribution Share Alike <VERSION>"
> 
>    But that's not the full name of the license according to CC.
> 
> I therefore suggest two changes to the SPDX License List:
> 
>  * Change existing Full Names to:
>       "Creative Commons Attribution Share Alike 4.0 International"
>       for the 4.0 version and,
>       "Creative Commons Attribution Share Alike <VERSION> Unported"
>       for the older ones.
> 
>    It seems that would be an uncontroversial change -- it just involves
>    adding "International" and "Unported" into the Full Name field.  Does
>    anyone have an argument why that *shouldn't* be done?
> 
>  * It *would* surely be controversial to add *every* version of *every*
>     jurisdiction-specific CC license in the SPDX license list.  Instead of
>     suggesting that, for the moment I suggest that "-Unported" should be
>     added to identifier for the pre-4.0 ones (i.e., "CC-BY-SA-3.0" becomes
>     "CC-BY-SA-3.0-Unported") so that no is confused by this situation.

mlinksva:
See #618

### DIFF
--- a/src/CC-BY-1.0.xml
+++ b/src/CC-BY-1.0.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <SPDXLicenseCollection xmlns="http://www.spdx.org/license">
    <license isOsiApproved="false" licenseId="CC-BY-1.0"
-            name="Creative Commons Attribution 1.0">
+            name="Creative Commons Attribution 1.0 Generic">
       <crossRefs>
          <crossRef>http://creativecommons.org/licenses/by/1.0/legalcode</crossRef>
       </crossRefs>
--- a/src/CC-BY-2.0.xml
+++ b/src/CC-BY-2.0.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <SPDXLicenseCollection xmlns="http://www.spdx.org/license">
    <license isOsiApproved="false" licenseId="CC-BY-2.0"
-            name="Creative Commons Attribution 2.0">
+            name="Creative Commons Attribution 2.0 Generic">
       <crossRefs>
          <crossRef>http://creativecommons.org/licenses/by/2.0/legalcode</crossRef>
       </crossRefs>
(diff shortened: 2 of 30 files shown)


mlinksva:
> Should we add suffixes to these forms (e.g. -unported)?

👎 unnecessary noise in concise identifier, not reflected in CC URLs. Just my opinion.

> Should we add some/all of the ported licenses?

I think that should be considered separately. It was a tangent to #618.

wking:
On Thu, Mar 22, 2018 at 10:50:47PM +0000, Mike Linksvayer wrote:
> > Should we add suffixes to these forms (e.g. -unported)?
> 
> unnecessary noise in concise identifier, not reflected in CC URLs.

It feels sort of like the “maybe folks wouldn't realize GPL-3.0 is
only, we should add a suffix to be extra clear (#542)”.  But the
maintainers may certainly have a differnt preference in this case.
Anyhow, closing #618 with this PR and forcing folks to file new,
specific issues for anything not covered here makes the discussion of
those points a bit more fragmented.  But additional focus would be
nice, so new issues wouldn't be the end of the world ;).

mlinksva:
Knowing which suffix CC chose to use for a particular version has no practical impact. `-only` or `-or-later` does.

I don't really care about how issues are grouped. Changed "Fixes" to "See" so #618 can be closed separately whenever maintainers wish to.

PR #625, opened 2018-03-23
Replacement for #551: New Exception: OpenJDK Assembly exception

swinslow:
This PR is meant to fix the build errors from #551. I'm submitting it as a separate PR solely because my Github skills aren't good enough for me to know how to add on to the existing PR at #551...

The content of the XML file for the OpenJDK assembly exception in this PR is identical to #551.
I've also made the following changes:

- Changed the XML's filename to make "assembly" lowercase and to have version 1.0 in the filename, to mirror the licenseID in the XML.
- Added a test file

### DIFF
--- a/src/exceptions/OpenJDK-assembly-exception-1.0.xml
+++ b/src/exceptions/OpenJDK-assembly-exception-1.0.xml
@@ -1,0 +1,52 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<SPDXLicenseCollection xmlns="http://www.spdx.org/license">
+   <exception licenseId="OpenJDK-assembly-exception-1.0" name="OpenJDK Assembly exception 1.0">
+      <crossRefs>
+         <crossRef>http://openjdk.java.net/legal/assembly-exception.html
+         </crossRef>
+      </crossRefs>
+      <text>
+      <titleText>OpenJDK Assembly Exception</titleText>
+      <p>
+         The OpenJDK source code made available by Oracle America, Inc.
+         (Oracle) at openjdk.java.net ("OpenJDK Code") is distributed
+         under the terms of the GNU General Public License
+         &lt;http://www.gnu.org/copyleft/gpl.html&gt; version 2 only
+         ("GPL2"), with the following clarification and special
+         exception.
+      </p>
+
+      <list>
+         <item>
+            <p>Linking this OpenJDK Code statically or dynamically with
+               other code is making a combined work based on this
+               library. Thus, the terms and conditions of GPL2 cover the
+               whole combination.
+            </p>
+
+            <p>As a special exception, Oracle gives you permission to
+               link this OpenJDK Code with certain code licensed by
+               Oracle as indicated at
+               http://openjdk.java.net/legal/exception-modules-2007-05-08.html
+               ("Designated Exception Modules") to produce an
+               executable, regardless of the license terms of the
+               Designated Exception Modules, and to copy and distribute
+               the resulting executable under GPL2, provided that the
+               Designated Exception Modules continue to be governed by
+               the licenses under which they were offered by Oracle.
+            </p>
+         </item>
+      </list>
+
+      <p>As such, it allows licensees and sublicensees of Oracle's GPL2
+         OpenJDK Code to build an executable that includes those
+         portions of necessary code that Oracle could not provide under
+         GPL2 (or that Oracle has provided under GPL2 with the Classpath
+         exception). If you modify or add to the OpenJDK code, that new
+         GPL2 code may still be combined with Designated Exception
+         Modules if the new code is made subject to this exception by
+         its copyright holder.
+      </p>
+      </text>
+   </exception>
+</SPDXLicenseCollection>
--- a/test/simpleTestForGenerator/OpenJDK-assembly-exception-1.0.txt
+++ b/test/simpleTestForGenerator/OpenJDK-assembly-exception-1.0.txt
@@ -1,0 +1,31 @@
+The OpenJDK source code made available by Oracle America, Inc.
+(Oracle) at openjdk.java.net ("OpenJDK Code") is distributed
+under the terms of the GNU General Public License
+&lt;http://www.gnu.org/copyleft/gpl.html&gt; version 2 only
+("GPL2"), with the following clarification and special
+exception.
+
+Linking this OpenJDK Code statically or dynamically with
+other code is making a combined work based on this
+library. Thus, the terms and conditions of GPL2 cover the
+whole combination.
+
+As a special exception, Oracle gives you permission to
+link this OpenJDK Code with certain code licensed by
+Oracle as indicated at
+http://openjdk.java.net/legal/exception-modules-2007-05-08.html
+("Designated Exception Modules") to produce an
+executable, regardless of the license terms of the
+Designated Exception Modules, and to copy and distribute
+the resulting executable under GPL2, provided that the
+Designated Exception Modules continue to be governed by
+the licenses under which they were offered by Oracle.
+
+As such, it allows licensees and sublicensees of Oracle's GPL2
+OpenJDK Code to build an executable that includes those
+portions of necessary code that Oracle could not provide under
+GPL2 (or that Oracle has provided under GPL2 with the Classpath
+exception). If you modify or add to the OpenJDK code, that new
+GPL2 code may still be combined with Designated Exception
+Modules if the new code is made subject to this exception by
+its copyright holder.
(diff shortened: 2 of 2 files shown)


src/exceptions/OpenJDK-assembly-exception-1.0.xml
```diff
@@ -0,0 +1,52 @@
...
+<SPDXLicenseCollection xmlns="http://www.spdx.org/license">
+   <exception licenseId="OpenJDK-assembly-exception-1.0" name="OpenJDK Assembly exception 1.0">
+      <crossRefs>
+         <crossRef>http://openjdk.java.net/legal/assembly-exception.html
+         </crossRef>
+      </crossRefs>
+      <text>
+      <titleText>OpenJDK Assembly Exception</titleText>
+      <p>
+         The OpenJDK source code made available by Oracle America, Inc.
```
wking: I [still][1] think that this text is closer to a grant than to exception text, with the exception text jusy being the “Linking this OpenJDK Code…” list item below.  It sounded like @jlovejoy [agreed][2], or at least felt that further discussion was warranted.  Has that further discussion happened?  Is the legal team on board with this text somehow being part of the exception and not part of the grant wording?  Or do we have some other plan for resolving this issue?

[1]: https://github.com/spdx/license-list-XML/pull/551#discussion_r158741220
[2]: https://github.com/spdx/license-list-XML/pull/551#discussion_r158868430

bradleeedmondson: If by 'grant' you mean 'recommended license header', i.e. a shorter way for developers to adopt the license (useful prior to adopting SPDX identifiers in source), then I tend to agree. The discussion on the legal call today was that this was an open question, but that we would merge now and resolve whether this text should have to be present for a match or not would be resolved later.

jlovejoy: we discussed whether to mark up the first and last paragraph as optional or not and went with leaving it as is for now - for the reasons of: it's easier to add markup later, rather than take it away and the text of the exception seems to only appear on that webpage (in other words, we did not see examples of just the middle two paragraphs in a license headers in source file, rather the license header refers to the webpage with the exception.
Btw, you keep referring to the "grant" in a way that is confusing: "grant" has legal meaning in that it is referring to the part of a license that grants rights to the licensee. The standard header or license notice (or various other names for the short-hand text that appears in actual source files, especially for longer licenses) is not a "grant" in the legal meaning of the word :)

wking: > Btw, you keep referring to the "grant" in a way that is confusing: "grant" has legal meaning in that it is referring to the part of a license that grants rights to the licensee. The standard header or license notice (or various other names for the short-hand text that appears in actual source files, especially for longer licenses) is not a "grant" in the legal meaning of the word :)

I don't mind “grant” meaning both a particular part of the license and the wording that associates the license with the content.  And neither Wikipedia nor I are lawyers, but they seem to use that meaning as well, [for example][1]:

> A license may be granted by a party…
>
> …
>
> A licensor may grant a license…

I'd rather not use “recommended license header” because that license ↔ content association text does not need to be in the header of the content file, it could be in a sibling file (e.g. `LICENSE.md`) or on the project website (as for this exception), etc.

I've be ok with “[license] notice”, but then I'm not clear on what the verb form should be (“Alice notices Bob this content under the MIT license” doesn't sound right ;).

Whatever we call it, I'm ok with “require this license ↔ content association text for now and possibly make it optional later”.  I think it makes the technical meaning a bit strange, for example if Alice claimed this exception for source that she provided directly and not through openjdk.java.net, the strict interpretation would be a bit confusing (“does the exception apply to Alice's code, or just Oracle's?”).  But even setting aside the license ↔ content association wording, I don't see how that usecase would work because of [the “Oracle gives you permission” exception text][2].  With everything wiggly for non-Oracle copyright holders, I'm fine punting on a license ↔ content association text cleanup until we can address non-Oracle copyright holders for this whole file.  But it would be nice to have a ballpark for when we intend to resolve this by, otherwise my guess is “a few years down the road if/when someone else complains about it” ;).

[1]: https://en.wikipedia.org/wiki/License
[2]: https://github.com/spdx/license-list-XML/pull/625/files#diff-d66b189cacb05b7231154da468563928R27

jlovejoy: people have generally referred to the text such as that suggested to be used at the end of the L/GPL and Apache-2.0 license as "license headers", "license notices" - and for SPDX, "standard headers" (where the license makes such an explicit delineation. We don't need to call it something else. Please simply take this advice - further discussion is not needed here.

bradleeedmondson: Right, I agree with @jlovejoy and find using "grant" for this purpose at best inapt, and at worst irresolvably ambiguous. From a legal perspective, there is a difference between the "license text," by which we mean the language of the contract, and the "license" itself, which is legal permission to do something that would otherwise infringe an intellectual property right.

Casting this situation info programming parlance, you cannot call the `grant` operation on a "license text" object (aka `contract`), only on a `license` itself (legal permission). The legal field has a specific term for when someone says "this is the contract that applies to this situation" without spelling it out word-for-word, and that is `incorporation by reference` (sometimes called `adoption` or `adoption by reference`) of the contract. 

When Alice says "0BSD applies to this software" and releases it to the public, she `adopts` the language of the 0BSD license text, which in turn `grants` the 0BSD-shaped permission (`license`) contained in the 0BSD license text (`contract`). But Alice's act of saying "0BSD applies to this software" is an `incorporation-by-reference` or an `adoption` (of the 0BSD license text/contract), not a `grant`. It's a legally effective statement that _contains_ a license grant, but her act is bigger than a "grant", it's agreement to all the contract terms in the 0BSD license text, including the license grant, the termination provisions (which rescinds the license grant), the right to cure within a certain time frame (as under GPL 3), and whatever other terms are in the license text/`contract`.

In the SPDX framework, we typically only encounter this as license-author-recommended text paragraphs intended to `incorporate` the license by reference, which is why SPDX has traditionally thought of these statements as standard headers (that's how the license authors intended them to be used). While we may or may not think this is not the full story (from a legal point of view), one thing we're pretty sure of is that this kind of statement, saying "0BSD applies," is an `incorporation by reference` (of the license text/`contract`), which _includes_ a `grant`; using/saying the statement "0BSD applies" is not itself a `grant`.




For reference, I'm using the following framework, since I understand this to follow the commonly used definitions in technology licensing and intellectual property law:

Term | Meaning
----- | -----
**Intellectual property** | legal right to prevent others from doing something (for copyright, copying/distributing/modifying/etc.)
**License** | legal permission to perform act that would otherwise be infringement in an intellectual property right
**Grant** | the legal operation of giving a license; the approximate opposite of license revocation or termination
**Contract** | written (usually) agreement that can contain terms and conditions of a legal license
**License text** | ambiguous; most commonly **(a)** the text of a contract granting a license (this is how it's commonly used, although from a legal analysis point of view, the contract is bigger than the license, since it contains the license as well as other contract promises and conditions); but possibly also **(b)** the text of the paragraph legally granting the license (independent of the other clauses in the contract).

wking: So you'd prefer Wikipedia's current:

> A licensor may grant a license…

to be:

> A licensor may adopt a contract…

or:

> A licensor my incorporation by reference a contract…

?  I guess I can live with that ;).

> In the SPDX framework, we typically only encounter this as license-author-recommended text paragraphs intended to incorporate the license by reference, which is why SPDX has traditionally thought of these statements as standard headers (that's how the license authors intended them to be used).

In some cases, yes.  For example, Apache-2.0's “[comment syntax][1]” and “[this file][2]” wording is clearly expecting the header-comment approach.  But the GPL-3.0 lists comment-headers as “[the safest][3]” way to “[attach the … notice to the program][3]”.

The GPL-3.0 *also* provides recommended wording for declaring the… adopted contract… [in an interactive terminal][4].  That information is filling the same role as the header-comment, but is designed for a different audience and channel.  And the FSF's recommended wording is different from their header-comment recommendation.

[1]: https://github.com/spdx/license-list-XML/blob/v3.0/src/Apache-2.0.xml#L219
[2]: https://github.com/spdx/license-list-XML/blob/v3.0/src/Apache-2.0.xml#L225
[3]: https://github.com/spdx/license-list-XML/blob/v3.0/src/GPL-3.0.xml#L786-L787
[4]: https://github.com/spdx/license-list-XML/blob/v3.0/src/GPL-3.0.xml#L816-L827

PR #627, opened 2018-03-23
Qt-LGPL-exception-1.1: Add new ID for Nokia-Qt-exception-1.1

wking:
And deprecate the old ID, which [had two problems][1]:

* Qt no longer belongs to Nokia, and they've changed the title upstream to reflect that.
* Qt has another exception for GPL3, and the old ID didn't mention the target license.

This commit adds the new identifier and deprecates the old one in its favor.

I've used Qt-LGPL-exception-1.1 instead of Kai's suggested Qt-exception-LGPL-1.1 to match the `{name}-{target}-exception-{version}` pattern suggested by the existing i2p-gpl-java-exception.

I've set an `<alt>` in the exception title so it will match the old “Nokia ” prefix, the current “The Qt Company ” prefix, or no prefix at all.  I've used “The Qt Company ” prefix in the test file to match the upstream text (the test is a copy/paste of the upstream text with trailing whitespace removed).

Builds on #626; review that first.  Someone with write access should label this “[new license/exception request][2]”.

CC @kkoehne.

[1]: https://lists.spdx.org/pipermail/spdx-legal/2018-March/002511.html
[2]: https://github.com/spdx/license-list-XML/labels/new%20license%2Fexception%20request

### DIFF
--- a/src/exceptions/Nokia-Qt-exception-1.1.xml
+++ b/src/exceptions/Nokia-Qt-exception-1.1.xml
@@ -10,8 +10,8 @@
          <p>Nokia Qt LGPL Exception version 1.1</p>
       </titleText>
       <p>As an additional permission to the GNU Lesser General Public
-       License version 2.1, the object code form of a ""work that
-       uses the Library"" may incorporate material from a header file
+       License version 2.1, the object code form of a "work that
+       uses the Library" may incorporate material from a header file
        that is part of the Library. You may distribute such object
        code under terms of your choice, provided that:</p>
       <list>
--- a/src/exceptions/Qt-LGPL-exception-1.1.xml
+++ b/src/exceptions/Qt-LGPL-exception-1.1.xml
@@ -1,16 +1,13 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <SPDXLicenseCollection xmlns="http://www.spdx.org/license">
-   <exception licenseId="Nokia-Qt-exception-1.1" name="Nokia Qt LGPL exception 1.1" isDeprecated="true" deprecatedVersion="3.1">
-      <obsoletedBys>
-         <obsoletedBy>Qt-LGPL-exception-1.1</obsoletedBy>
-      </obsoletedBys>
+   <exception licenseId="Qt-LGPL-exception-1.1" name="Qt LGPL exception 1.1" listVersionAdded="3.1">
       <crossRefs>
-         <crossRef>https://www.keepassx.org/dev/projects/keepassx/repository/revisions/b8dfb9cc4d5133e0f09cd7533d15a4f1c19a40f2/entry/LICENSE.NOKIA-LGPL-EXCEPTION</crossRef>
+         <crossRef>http://code.qt.io/cgit/qt/qtbase.git/tree/LGPL_EXCEPTION.txt</crossRef>
       </crossRefs>
-      <notes>DEPRECATED: Use Qt-LGPL-exception-1.1</notes>
+      <notes>Used with the LGPL-2.1, which is mentioned explicitly in the exception text.</notes>
     <text>
       <titleText>
-         <p>Nokia Qt LGPL Exception version 1.1</p>
+         <p><alt name="nokia" match="|Nokia |The Qt Company ">The Qt Company </alt>Qt LGPL Exception version 1.1</p>
       </titleText>
       <p>As an additional permission to the GNU Lesser General Public
        License version 2.1, the object code form of a "work that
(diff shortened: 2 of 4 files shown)


silverhook:
I agree.

And if memory serves me right, (a form of) the [exception predates](https://web.archive.org/web/20080905005722/http://doc.trolltech.com/4.4/license-gpl-exceptions.html) Nokia’s acquisition of Qt as well.

Qt got passed on as such: Trolltech ↦ Nokia ↦ Digia↦ Qt Company – which further strengthens the suggestion that a more generic name should be used.

There is also the [Free Qt Foundation](https://www.kde.org/community/whatiskde/kdefreeqtfoundation.php), but that’s not the primary concern here :)

goneall:
It turns out there are 2 problems causing the tests to fail.

Having a null match in the alt match ```match="|Nokia |The Qt Company "``` is matching the empty string first and ignoring the matches for ```The Qt Company ```, so it doesn't skip those tokens.  To solve this,  optional can be placed around the alt and the match.

The second problem is including the trailing space in the matching text.  Since the matching software tokenizes and removes white space, it doesn't include the space in the match.  To work around this, just don't include any preceding or trailing white space in the match.

Replacing ```<alt name="nokia" match="|Nokia |The Qt Company ">The Qt Company </alt>``` with ```<optional><alt name="nokia" match="Nokia|The Qt Company">The Qt Company </alt></optional>``` should work.

wking:
> Having a null match in the alt match...

Is this something we can fix in the parser?  We already have some of these in master (e.g. [here][1]).

>     <alt name="nokia" match="Nokia|The Qt Company">The Qt Company </alt>

Including a space in the inner text but not in the `match` might break a canonical parser.  I think I'll need to accept an unnecessary leading space and use:

    <alt name="nokia" match="|Nokia|The Qt Company">The Qt Company</alt>

with the trailing space outside the `<alt>`.

[1]: https://github.com/spdx/license-list-XML/blob/7fa29e22ea35652e8fd5717eaf2c6b57cf8d9ff8/src/GPL-2.0-or-later.xml#L20

goneall:
> Is this something we can fix in the parser?

I could not come up with a good general fix (I'll describe the problem in more detail below).

I did find another possible work-around.  If you order the match so that the empty match is at the end (e.g. ```match="Nokia|The Qt Company|"```), it will work.  I'm not sure if this work-around is implementation specific or something that is in the Regular Expression specification.  The pattern matcher seems to try to match the options in order.

The specific issue is the algorithm used to the the alt matching:

1.  Identify the text that may apply to the alt pattern by looking ahead to the next non-alt and non-optional text areas
2.  Create a regular expression matcher based on the ```match=...``` pattern provided
3.  Attempt to match using the regular expression
4.  If the pattern does not match - just continue matching at the current text position
5.  If the pattern does match, move the cursor forward to after the match, then continue matching

The problem is number 5 above - if an empty choice is matched (and it will always match), the cursor is moved forward by zero tokens (e.g. it doesn't move), this leaves the cursor at its current position and doesn't match the variable text we want it to.  The match then fails when it applies the next matching rule.

I can not think of a general design that would avoid this situation.

wking:
Rebased onto master to pick up #392 so we can add an `<obsoletedBy>` entry.

goneall:
I did some more research on the regex matching and confirmed that the order of Regex alternation does matter and is consistent between Regex engine implementations.

Based on this information, the pattern should be changed from ```match="|Nokia|The Qt Company"``` to ```match="Nokia|The Qt Company|"```  This is a solution and not a work-around as the matches are working properly.  

The problem with the current match is that the empty alternation matches first (and always matches) and the rest of the regex pattern isn't even examined.  Since the company name (in this case) is not consumed by the regex matching, the rest of the pattern match will fail.

There is no need to surround the ```<alt> ... </alt>``` by ```<optional>...</optional>```

We should check the order of the other alts where empty alternations are in the match string - the empty alternation should always be last.

wking:
> I did some more research on the regex matching and confirmed that the order of Regex alternation does matter and is consistent between Regex engine implementations.

Interesting.  I did some digging too, and it looks like the POSIX `|` is commutative, preferring the longest overall match.  But it also looks like many (but not all) regexp engines prefer the left-most-branch match instead for performance reasons.  I've pushed b6b27fd here to reorder the matches so that longest-match and left-most-branch coincide, which is a workaround that allows us to support those left-most-branch implementations.  More details in the b6b27fd commit message.

wking:
> We should check the order of the other alts where empty alternations are in the match string - the empty alternation should always be last.

I've filed #629 adding left-most-branch compatibility to the regexps which are already in master.

bradleeedmondson:
Discussed on legal call today; agree to adding new ID and deprecating old ID that references Nokia.

PR #628, opened 2018-03-23
Qt-GPL-exception-1.0: Add this new exception

wking:
Based on @kkoehne's [request][1].  The test is a copy/paste from the upstream source.  I've added an `<alt>` to match assorted company prefixes in the title.

Someone with write access should label this “[new license/exception request][2]”.

[1]: https://lists.spdx.org/pipermail/spdx-legal/2018-March/002512.html
[2]: https://github.com/spdx/license-list-XML/labels/new%20license%2Fexception%20request

### DIFF
--- a/src/exceptions/Qt-GPL-exception-1.0.xml
+++ b/src/exceptions/Qt-GPL-exception-1.0.xml
@@ -11,7 +11,7 @@
       </titleText>
       <p>Exception 1:</p>
       <p>
-        As a special exception you may create a larger work which contains the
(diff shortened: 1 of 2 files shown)


wking:
I've pushed c5842dd with an `<optional>,</optional>` based on [this list message][1].

[1]: https://lists.spdx.org/pipermail/spdx-legal/2018-March/002522.html

bradleeedmondson:
discussed on legal call today; agree to add exception

PR #640, opened 2018-04-25
[Typo]Adding paragraph break for Content def. item

camillem:


goneall:
The failed check is due to a version incompatibility due to a new a new release of NodeJS.  This is already fexied by #644 which specifies a node version which is compatible.  This error should go away when we merge it.

Issue #504, opened 2017-12-02
Automatically update the license-list-data repository on every push or PR


goneall:
Related to Issue #503 in that it could leverage the same Java code and push the generated license list data on every successful update to the master branch.

We would need a way to cache the credentials for github to do the push.

swinslow:
Discussed on 2019-07-30 tech team call, believe this is done but @goneall will confirm and then close.

goneall:
Confirmed - this has already been resolved. Closing.

PR #642, opened 2018-05-04
Automatically push changes to license-list-data

goneall:
Resolves issue #504


goneall:
Hmmm - seems like all the files are marked as changed.  Going to try this again

PR #643, opened 2018-05-04
Automatically push changes to license-list-data

goneall:
Resolves issue #504 

Note: this PR pushes changes to the goneall fork of the license-list-data.  Once reviewed, this should be changed to spdx/license-list-data.


goneall:
@wking Any thoughts before I merge this in?

.travis.yml
```diff
@@ -2,7 +2,7 @@ matrix:
   include:
     - language: java
       script:
-        - make validate-canonical-match
+        - 'if [[ "$TRAVIS_PULL_REQUEST" = "false" && "master" = "$TRAVIS_BRANCH" ]]; then make deploy-license-data; else make validate-canonical-match; fi'
     - language: node_js
       node_js:
-      - "9"
+      - "9"
```
wking: Can we keep the trailing newline?  It makes it easier to read the file when `cat`ing it to the terminal, because the next prompt will start on a new line.

Makefile
```diff
@@ -1,9 +1,58 @@
 TOOL_VERSION = 2.1.12
 TEST_DATA = test/simpleTestForGenerator
-
+GIT_USERNAME=License Publisher (maintained by Gary O'Neall)
+GIT_EMAIL=gary@sourceauditor.com
+LICENSE_DATA_REPO=https://github.com/goneall/license-list-data.git
+LICENSE_DATA_REPO_NO_SCHEME=github.com/goneall/license-list-data.git
```
wking: To stay DRY, how about:

```
LICENSE_DATA_REPO_NO_SCHEME = github.com/goneall/license-list-data.git
LICENSE_DATA_REPO = https://$(LICENSE_DATA_REPO_NO_SCHEME)
```

Also, I think these variable definitions are more readable with whitespace around the `=` (the examples in [the docs][1] use whitespace).

[1]: https://www.gnu.org/software/make/manual/html_node/Setting.html

Makefile
```diff
@@ -1,9 +1,58 @@
...
+LICENSE_DATA_URL=https://$(GITHUB_TOKEN)@$(LICENSE_DATA_REPO_NO_SCHEME)
+LICENSE_OUTPUT_DIR=.tmp
+VERSION=$(shell git describe --always || echo 'UNKNOWN')
+RELEASE_DATE=$(shell date -I)
```
wking: For [POSIX compliance][1], I think we want `date '+%Y-%m-%d'`.

[1]: http://pubs.opengroup.org/onlinepubs/9699919799/utilities/date.html

Makefile
```diff
@@ -1,9 +1,58 @@
...
+COMMIT_MSG="Automatic update of license list data from license XML repository version $(VERSION)"
```
wking: The output repo shouldn't care whether the source repo uses XML or not.  And we don't need to say what's being updated, because it will always be the bulk of the output repo.  I recommend a commit message closer to:

```
COMMIT_MSG = Build $(VERSION) using publisher $(TOOL_VERSION)
```

Makefile
```diff
@@ -1,9 +1,58 @@
...
+#TODO Change the license data repo to license-list-xml before merging pull request
+	
 .PHONY: validate-canonical-match
 validate-canonical-match: licenseListPublisher-$(TOOL_VERSION).jar-valid resources/licenses-full.json $(TEST_DATA) .tmp
-	java -jar -DLocalFsfFreeJson=true -DlistedLicenseSchema="schema/ListedLicense.xsd" licenseListPublisher-$(TOOL_VERSION).jar LicenseRDFAGenerator src .tmp 1.0 2000-01-01 $(TEST_DATA) expected-warnings
+	java -jar -DLocalFsfFreeJson=false -DlistedLicenseSchema="schema/ListedLicense.xsd" licenseListPublisher-$(TOOL_VERSION).jar LicenseRDFAGenerator src .tmp 1.0 2000-01-01 $(TEST_DATA) expected-warnings
```
wking: This seems like an orthogonal change, but I'm fine leaving it in here if you don't want to spin it off into its own PR.  If you keep it, you probably want to drop the`resources/licenses-full.json` dependency above, and possibly drop the `resources/licenses-full.json` target as well.

Makefile
```diff
@@ -1,9 +1,58 @@
...
+
+.PHONY: deploy-license-data
+deploy-license-data: licenseListPublisher-$(TOOL_VERSION).jar-valid resources/licenses-full.json $(TEST_DATA) .tmp
```
wking: No need for the `resources/licenses-full.json` dependency here, since you set `-DLocalFsfFreeJson=false` below.  Or maybe I'm misunderstanding that option?

Makefile
```diff
@@ -1,9 +1,58 @@
...
+#TODO Change the license data repo to license-list-xml before merging pull request
+	
 .PHONY: validate-canonical-match
 validate-canonical-match: licenseListPublisher-$(TOOL_VERSION).jar-valid resources/licenses-full.json $(TEST_DATA) .tmp
-	java -jar -DLocalFsfFreeJson=true -DlistedLicenseSchema="schema/ListedLicense.xsd" licenseListPublisher-$(TOOL_VERSION).jar LicenseRDFAGenerator src .tmp 1.0 2000-01-01 $(TEST_DATA) expected-warnings
+	java -jar -DLocalFsfFreeJson=false -DlistedLicenseSchema="schema/ListedLicense.xsd" licenseListPublisher-$(TOOL_VERSION).jar LicenseRDFAGenerator src .tmp 1.0 2000-01-01 $(TEST_DATA) expected-warnings
+
+.PHONY: deploy-license-data
+deploy-license-data: licenseListPublisher-$(TOOL_VERSION).jar-valid resources/licenses-full.json $(TEST_DATA) .tmp	
+	git clone $(LICENSE_DATA_URL) $(LICENSE_OUTPUT_DIR) --quiet --depth 1
```
wking: I think the idiomatic approach is to put the options before the positional arguments:

```
git clone --quiet --depth 1 $(LICENSE_DATA_URL) $(LICENSE_OUTPUT_DIR)
```

Makefile
```diff
@@ -1,9 +1,58 @@
...
+	# Clean out the old data directories
+	rm -r .tmp/html
```
wking: The clone will fail if `.tmp` already exists:

```console
$ mkdir -p .tmp/html
$ git clone --depth 1 git://github.com/spdx/license-list-data.git .tmp
fatal: destination path '.tmp' already exists and is not an empty directory.
```

If you want to ensure a fresh clone each time, you should either switch on the existence of `LICENSE_OUTPUT_DIR` to choose between `clone` and `pull`, or always clobber any pre-existing clone:

```
rm -rf $(LICENSE_OUTPUT_DIR)
git clone --quiet --depth 1 $(LICENSE_DATA_URL) $(LICENSE_OUTPUT_DIR)
```

Makefile
```diff
@@ -1,9 +1,58 @@
...
+	rm -r .tmp/rdfnt
+	rm -r .tmp/rdfturtle
+	rm -r .tmp/rdfxml
+	rm -r .tmp/template
+	rm -r .tmp/text
+	rm -r .tmp/website
+	rm .tmp/licenses.md
+	java -jar -DLocalFsfFreeJson=false -DlistedLicenseSchema="schema/ListedLicense.xsd" licenseListPublisher-$(TOOL_VERSION).jar LicenseRDFAGenerator src .tmp $(VERSION) $(RELEASE_DATE) $(TEST_DATA) expected-warnings
+	
+	echo $COMMIT_MSG
```
wking: No need for this `echo`, because `make` will show the `git commit -m ...` line as it runs it.

Makefile
```diff
@@ -1,9 +1,58 @@
...
+	cd .tmp && \
```
wking: This needs to be `$(LICENSE_OUTPUT_DIR)` to respect users who change that variable value.

Makefile
```diff
@@ -1,9 +1,58 @@
...
+	pwd && \
```
wking: No need for this; maybe it was for debugging?

Makefile
```diff
@@ -1,9 +1,58 @@
...
+	git config user.email "$(GIT_EMAIL)" && \
+	git config user.name "$(GIT_USERNAME)" && \
```
wking: You can set these in the commit itself with `git commit --author "$(GIT_USERNAME <$(GIT_EMAIL)>" ...`.  With that approach, you may want a single `GIT_AUTHOR` variable instead.

Makefile
```diff
@@ -1,9 +1,58 @@
...
+	git add -A && \
+	git commit -m $(COMMIT_MSG) && \
```
wking: To avoid `&&` chaining, this could be:

```
git -C "$(LICENSE_OUTPUT_DIR)" add -A .
git -C "$(LICENSE_OUTPUT_DIR)" commit --author "$(GIT_AUTHOR)" -m "$(COMMIT_MSG)"
git -C "$(LICENSE_OUTPUT_DIR)" push --quiet origin
```

Makefile
```diff
@@ -1,9 +1,58 @@
...
+	
+.PHONY: release-license-data
+release-license-data: deploy-license-data
+	if [[ $VERSION =~ .+-g[a-f0-9]{7} ]]
+	then
+		echo Can not release license data - license list version '$VERSION' does not match a release pattern
+		exit 1
+	else
+		cd .tmp && \
+		git tag -a $(VERSION) -m "Adding release mathing the license list XML tag $(VERSION)" && \
```
wking: I'd really like to be using signed tags for our data releases.  That means we either need to set an encrypted Travis variable with a secret key, or we need to be signing these tags locally instead of creating them in Travis.  Personally, my preference is for the latter, because establishing trust in a secret key stored in a Travis variable would be difficult.

Also, it seems strange to use a variable for the commit message, but to define the tag message inline here.  If you go ahead with the Travis-tagging approach, can you either make both strings inline, or make both use variables?

.travis.yml
```diff
@@ -2,7 +2,7 @@ matrix:
   include:
     - language: java
       script:
-        - make validate-canonical-match
+        - 'if [[ "$TRAVIS_PULL_REQUEST" = "false" && "master" = "$TRAVIS_BRANCH" ]]; then make deploy-license-data; else make validate-canonical-match; fi'
```
wking: I don't think we need the wrapping single-quote here.  For a similar example that does not use it, see [here][1].

[1]: https://github.com/spdx/spdx-spec/blob/1f7f2a0213eceba89aec010b89f286fc6e0a0a21/.travis.yml#L18

Makefile
```diff
@@ -1,9 +1,53 @@
...
-validate-canonical-match: licenseListPublisher-$(TOOL_VERSION).jar-valid resources/licenses-full.json $(TEST_DATA) .tmp
-	java -jar -DLocalFsfFreeJson=true -DlistedLicenseSchema="schema/ListedLicense.xsd" licenseListPublisher-$(TOOL_VERSION).jar LicenseRDFAGenerator src .tmp 1.0 2000-01-01 $(TEST_DATA) expected-warnings
+validate-canonical-match: licenseListPublisher-$(TOOL_VERSION).jar-valid $(TEST_DATA) $(LICENSE_OUTPUT_DIR)
+	java -jar -DLocalFsfFreeJson=false -DlistedLicenseSchema="schema/ListedLicense.xsd" licenseListPublisher-$(TOOL_VERSION).jar LicenseRDFAGenerator src $(LICENSE_OUTPUT_DIR) 1.0 2000-01-01 $(TEST_DATA) expected-warnings
+
+.PHONY: deploy-license-data
+deploy-license-data: licenseListPublisher-$(TOOL_VERSION).jar-valid $(TEST_DATA) $(LICENSE_OUTPUT_DIR)	
+	git clone $(LICENSE_DATA_URL) $(LICENSE_OUTPUT_DIR) --quiet --depth 1
+	# Clean out the old data directories
+	rm -r $(LICENSE_OUTPUT_DIR)/html
```
wking: I [still][1] think you should replace these with a blanket removal before the `clone`.  Have you run a successful `clone` with a non-empty `LICENSE_OUTPUT_DIR`?

[1]: https://github.com/spdx/license-list-XML/pull/643#discussion_r186794301

wking: Ah, if you're looking for a clean slate *after* the clone (removing files from the data repo unless they are fresh from this run), I think you should replace these with:

```
git -C "$(LICENSE_OUTPUT_DIR)" rm -rf .
```

And I think you want the earlier `rm -rf .tmp/*` before the clone as well.

Makefile
```diff
@@ -1,9 +1,53 @@
...
+	rm -r $(LICENSE_OUTPUT_DIR)/rdfnt
+	rm -r $(LICENSE_OUTPUT_DIR)/rdfturtle
+	rm -r $(LICENSE_OUTPUT_DIR)/rdfxml
+	rm -r $(LICENSE_OUTPUT_DIR)/template
+	rm -r $(LICENSE_OUTPUT_DIR)/text
+	rm -r $(LICENSE_OUTPUT_DIR)/website
+	rm $(LICENSE_OUTPUT_DIR)/licenses.md
+	java -jar -DLocalFsfFreeJson=false -DlistedLicenseSchema="schema/ListedLicense.xsd" licenseListPublisher-$(TOOL_VERSION).jar LicenseRDFAGenerator src $(LICENSE_OUTPUT_DIR) $(VERSION) $(RELEASE_DATE) $(TEST_DATA) expected-warnings
+	
+	echo $(COMMIT_MSG)
```
wking: [Still][1] no need for this.

[1]: https://github.com/spdx/license-list-XML/pull/643#discussion_r186794610

Makefile
```diff
@@ -1,9 +1,53 @@
...
+	git -C "$(LICENSE_OUTPUT_DIR)" add -A .
+	git -C "$(LICENSE_OUTPUT_DIR)" commit --author "$(GIT_AUTHOR)" -m "$(COMMIT_MSG)"
+	echo Pushing updates to the license list data repository.  This could take a while...
```
wking: If you're concerned about folks getting antsy, I'd rather drop this line and also remove the `--quiet` from the following push.

Makefile
```diff
@@ -1,9 +1,53 @@
...
+GIT_AUTHOR = License Publisher (maintained by Gary O'Neall) <gary@sourceauditor.com>
+LICENSE_DATA_REPO_NO_SCHEME = github.com/goneall/license-list-data.git
+LICENSE_DATA_REPO = https://$(LICENSE_DATA_REPO_NO_SCHEME)
+LICENSE_DATA_URL = https://$(GITHUB_TOKEN)@$(LICENSE_DATA_REPO_NO_SCHEME)
+LICENSE_OUTPUT_DIR = .tmp
+VERSION = $(shell git describe --always || echo 'UNKNOWN')
+RELEASE_DATE = $(shell date '+%Y-%m-%d')
+COMMIT_MSG = License list build $(VERSION) using license list publisher $(TOOL_VERSION)
+RELEASE_MSG = Adding release matching the license list XML tag $(VERSION)
+#TODO Change the license data repo to license-list-xml before merging pull request
```
wking: I'm not sure what you intend to do here, but if you're getting close to merging, you may want to address this todo.

Makefile
```diff
@@ -1,9 +1,53 @@
...
+	
 .PHONY: validate-canonical-match
-validate-canonical-match: licenseListPublisher-$(TOOL_VERSION).jar-valid resources/licenses-full.json $(TEST_DATA) .tmp
-	java -jar -DLocalFsfFreeJson=true -DlistedLicenseSchema="schema/ListedLicense.xsd" licenseListPublisher-$(TOOL_VERSION).jar LicenseRDFAGenerator src .tmp 1.0 2000-01-01 $(TEST_DATA) expected-warnings
+validate-canonical-match: licenseListPublisher-$(TOOL_VERSION).jar-valid $(TEST_DATA) $(LICENSE_OUTPUT_DIR)
+	java -jar -DLocalFsfFreeJson=false -DlistedLicenseSchema="schema/ListedLicense.xsd" licenseListPublisher-$(TOOL_VERSION).jar LicenseRDFAGenerator src $(LICENSE_OUTPUT_DIR) 1.0 2000-01-01 $(TEST_DATA) expected-warnings
+
+.PHONY: deploy-license-data
+deploy-license-data: licenseListPublisher-$(TOOL_VERSION).jar-valid $(TEST_DATA) $(LICENSE_OUTPUT_DIR)
```
wking: `clone` will create the target directory if it didn't exist before, so there's no need for `$(LICENSE_OUTPUT_DIR)` in the prereqs here.

Makefile
goneall: I'll add an rm -rf .tmp prior to the clone.  I want to retain the README and other files, so I think I need to remove the content of each directory.

Makefile
goneall: I'll update the repo url after the travis run and right before the merge

Makefile
wking: > I want to retain the README and other files, so I think I need to remove the content of each directory.

Maybe something like:

    find "$(LICENSE_OUTPUT_DIR)" -mindepth 1 -maxdepth 1 -name .git -prune -o -type d -exec rm -rf {} \+

to automatically remove all directories other than `.git`?

Makefile
wking: > I'll update the repo url after the travis run and right before the merge

Looks like you updated the URL in 7d169ff8164, but you still have this comment?

Makefile
```diff
@@ -1,9 +1,43 @@
...
+	
 .PHONY: validate-canonical-match
-validate-canonical-match: licenseListPublisher-$(TOOL_VERSION).jar-valid resources/licenses-full.json $(TEST_DATA) .tmp
-	java -jar -DLocalFsfFreeJson=true -DlistedLicenseSchema="schema/ListedLicense.xsd" licenseListPublisher-$(TOOL_VERSION).jar LicenseRDFAGenerator src .tmp 1.0 2000-01-01 $(TEST_DATA) expected-warnings
+validate-canonical-match: licenseListPublisher-$(TOOL_VERSION).jar-valid $(TEST_DATA) $(LICENSE_OUTPUT_DIR)
+	java -jar -DLocalFsfFreeJson=false -DlistedLicenseSchema="schema/ListedLicense.xsd" licenseListPublisher-$(TOOL_VERSION).jar LicenseRDFAGenerator src $(LICENSE_OUTPUT_DIR) 1.0 2000-01-01 $(TEST_DATA) expected-warnings
+
+.PHONY: deploy-license-data
+deploy-license-data: licenseListPublisher-$(TOOL_VERSION).jar-valid $(TEST_DATA)
+	rm -rf $(LICENSE_OUTPUT_DIR)
```
wking: I doubt anyone would override `LICENSE_OUTPUT_DIR` to include whitespace or other shell-sensitive characters.  But in some of these lines we're quoting the expanded value (`"$(LICENSE_OUTPUT_DIR)"`) to protect against that chance.  If we want that protection (and it's cheap, so why not), I think we want it here too.  If you feel that the protection is unnecessary, then we should remove the quotes from the other locations.

Makefile
```diff
@@ -1,9 +1,43 @@
...
+	git clone $(LICENSE_DATA_URL) $(LICENSE_OUTPUT_DIR) --quiet --depth 1
```
wking: I [still][1] think we want to list the options before the positional arguments.

[1]: https://github.com/spdx/license-list-XML/pull/643#discussion_r186793330

Makefile
bradleeedmondson: +1 for adding quotes on the off chance (reducing theoretical exploitability)

Makefile
```diff
@@ -1,9 +1,43 @@
...
+	# Clean out the old data directories
+	find "$(LICENSE_OUTPUT_DIR)" -mindepth 1 -maxdepth 1 -name .git -prune -o -type d -exec rm -rf {} \+
+	rm -f $(LICENSE_OUTPUT_DIR)/licenses.md
```
wking: Do we need this?  I expect generated output will always create this file, and we only need to remove things that may not be generated by future builds (e.g. if we remove a deprecated license identifier in the v4.0 license list).

Makefile
```diff
@@ -1,9 +1,43 @@
...
+	
+	git -C "$(LICENSE_OUTPUT_DIR)" add -A .
+	git -C "$(LICENSE_OUTPUT_DIR)" commit --author "$(GIT_AUTHOR)" -m "$(COMMIT_MSG)"
+	git -C "$(LICENSE_OUTPUT_DIR)" push origin
+	
+.PHONY: release-license-data
+release-license-data: deploy-license-data
+	if [[ $VERSION =~ .+-g[a-f0-9]{7} ]]
+	then
+		echo Can not release license data - license list version '$VERSION' does not match a release pattern
```
wking: I think you want `$(VERSION)` here and a few lines above.  For example:

```console
$ cat Makefile
VERSION = 5

x:
  echo a '$VERSION' b
$ make x
echo a 'ERSION' b
a ERSION b
```

I also think you can drop the single-quotes, because `$(VERSION)` should not contain shell-sensitive characters.

Makefile
```diff
@@ -1,9 +1,43 @@
...
+		exit 1
+	else
+		git -C "$(LICENSE_OUTPUT_DIR)" tag -a $(VERSION) -m "$(RELEASE_MESSAGE)"
```
wking: This is generating unsigned tags, which I [still][1] don't think we want.

[1]: https://github.com/spdx/license-list-XML/pull/643#discussion_r186797369

Makefile
wking: > +1 for adding quotes on the off chance (reducing theoretical exploitability)

You can still shoot yourself in the foot.  For example:

```console
$ head -n4 Makefile 
VARIABLE = mua"; haha; "

escape:
        python -c 'import sys; print(sys.argv)' "$(VARIABLE)"
$ make escape
python -c 'import sys; print(sys.argv)' "mua"; haha; ""
['-c', 'mua']
/bin/sh: haha: command not found
/bin/sh: : command not found
make: *** [Makefile:4: escape] Error 127
```

But "issues with values that contain `$` or `"`" is narrower than you'd have without the wrapping quotes.  In fact, `'$(LICENSE_OUTPUT_DIR)'`, would be better than the double-quoted approach because it would protect against values that contain `$` and only have issues with values that contain `'`.

Makefile
wking: You probably also want `echo` -> `@echo` to avoid printing the command, because for `echo` the command largely duplicates the standard output.

Makefile
```diff
@@ -1,9 +1,43 @@
...
+	rm -f $(LICENSE_OUTPUT_DIR)/licenses.md
+	java -jar -DLocalFsfFreeJson=false -DlistedLicenseSchema="schema/ListedLicense.xsd" licenseListPublisher-$(TOOL_VERSION).jar LicenseRDFAGenerator src $(LICENSE_OUTPUT_DIR) $(VERSION) $(RELEASE_DATE) $(TEST_DATA) expected-warnings
+	
+	git -C "$(LICENSE_OUTPUT_DIR)" add -A .
+	git -C "$(LICENSE_OUTPUT_DIR)" commit --author "$(GIT_AUTHOR)" -m "$(COMMIT_MSG)"
+	git -C "$(LICENSE_OUTPUT_DIR)" push origin
+	
+.PHONY: release-license-data
+release-license-data: deploy-license-data
+	if [[ $VERSION =~ .+-g[a-f0-9]{7} ]]
```
wking: You'll need semicolons here (and `\` line continuation unless you collapse the whole `if` block onto one line).  Otherwise `make` splits on the newlines and tries to execute them separately:

```console
$ cat Makefile 
test:
        if [ 1 = 1 ]
        then
                echo true
        else
                echo false
        fi
$ make
if [ 1 = 1 ]
/bin/sh: -c: line 1: syntax error: unexpected end of file
make: *** [Makefile:2: test] Error 1
```

Makefile
goneall: leaving this in for the off chance we change the generator to no longer creates this file in the future.  We wouldn't want a stale version to be left in the repo.

Makefile
goneall: BTW - this goal hasn't really been tested (it isn't used in the Travis file), so there are probably some other bugs in here.  We could just comment out this section and save the improvements for later.

goneall: I'll create a separate PR

Issue #399, opened 2017-04-03
SPDX matching guidelines and common variant of BSD-3-Clause


bradleeedmondson:
From Richard:

---------------------------------

I've noticed that a very common variant of what I think of as the 3-clause BSD license has this as its third clause:

3. The name of the author may not be used to endorse or promote
      products derived from this software without specific prior
      written permission.

Am I right that such a license, otherwise matching[1] what SPDX calls 'BSD-3-Clause', would not be considered a match to 'BSD-3-Clause' because clause 3 does not say 'Neither the name of the author nor the name of its contributors...'?

Asking because this seems to be the case from a strict reading of the SPDX matching guidelines, yet the result also seems to violate its spirit.

 Richard

[1]The non-use of 'contributors' results in changes to the stock disclaimer language too. I don't think it is clear whether this too would result in a non-match under the SPDX guidelines.

jlovejoy:
I agree that these two variations are not legally different and should be accommodated in our matching markup. 
 
Text of clause 3 we have for comparison:

3. Neither the name of the copyright holder nor the names of its contributors may be used to endorse or promote products derived from this software without specific prior written permission.

But would need help with someone from the technical side as to how to implement such markup

jlovejoy:
marking as later, as need to make sure other lawyers agree

bradleeedmondson:
discussed on legal call today; we agree that these two wordings are identical and they should be marked up to both match BSD-3 clause.

one thing to note is that we need to be careful not to do a simple match for both "may" and "may not" as their meanings are inverted in the two versions by language earlier in the sentence.

Would @wking or @goneall be able to write and test some &lt;alt&gt; tags that will match both, but not the hybrids "neither ... may _not_" or  "The name of the author _may_ be used..."?

goneall:
I created a pull request #650 which has been unit tested with the 2 variants for clause 3 mentioned above.

jlovejoy:
closing this issue as #650 has now been merged.

PR #650, opened 2018-06-04
Add alt tag to handle BSD 3rd clause alternative language

goneall:
Resolves issue #399

### DIFF
--- a/src/BSD-3-Clause.xml
+++ b/src/BSD-3-Clause.xml
@@ -26,8 +26,9 @@
             </item>
             <item>
                 <bullet>3.</bullet>
-                Neither the name of <alt match=".+" name="organizationClause3">the copyright holder</alt> 
-                 nor the names of its contributors may be used to endorse or promote products derived from this 
+				<alt match="(The\sname\sof\s.+may not)|(Neither\sthe\sname\sof\s.+nor\sthe\snames\sof\sits\scontributors\smay)" name="organizationClause3">
+                Neither the name of the copyright holder nor the names of its contributors may</alt> 
+                 be used to endorse or promote products derived from this 
                  software without specific prior written permission.
             </item>
         </list>
--- a/src/BSD-3-Clause.xml
+++ b/src/BSD-3-Clause.xml
@@ -26,8 +26,8 @@
             </item>
             <item>
                 <bullet>3.</bullet>
-				<alt match="(The\sname\sof\s.+may not)|(Neither\sthe\sname\sof\s.+nor\sthe\snames\sof\sits\scontributors\smay)" name="organizationClause3">
-                Neither the name of the copyright holder nor the names of its contributors may</alt> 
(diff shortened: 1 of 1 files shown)


src/BSD-3-Clause.xml
```diff
@@ -26,8 +26,9 @@
             </item>
             <item>
                 <bullet>3.</bullet>
-                Neither the name of <alt match=".+" name="organizationClause3">the copyright holder</alt> 
-                 nor the names of its contributors may be used to endorse or promote products derived from this 
+				<alt match="(The\sname\sof\s.+may not)|(Neither\sthe\sname\sof\s.+nor\sthe\snames\sof\sits\scontributors\smay)" name="organizationClause3">
```
wking: I think we want literal spaces here instead of `\s`.  [Our docs reference POSIX EREs][1] and [POSIX EREs do not include `\s`][2].  POSIX EREs do include `[:space:]`, but we already rely on the matching engine to normalize whitespace, case, etc. before comparing with our templates.

Also, you seem to be using tabs to indent while the surrounding code uses spaces (although I'm still holding out hope for automatically formatting these, either via #432 or @zvr's [xmlindent][3]).

[1]: https://github.com/spdx/license-list-XML/blob/v3.1/schema/ListedLicense.xsd#L259-L267
[2]: http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap09.html#tag_09_04
[3]: https://github.com/zvr/xmlindent

src/BSD-3-Clause.xml
```diff
@@ -26,8 +26,9 @@
             </item>
             <item>
                 <bullet>3.</bullet>
-                Neither the name of <alt match=".+" name="organizationClause3">the copyright holder</alt> 
-                 nor the names of its contributors may be used to endorse or promote products derived from this 
+                <alt match="(The name of.+may not)|(Neither the name of.+nor the names of its contributors may)" name="organizationClause3">
```
wking: Maybe `.+` -> ` .+ ` (requiring wrapping whitespace) for both `.+`.

goneall: Thought we would leave out the spaces in case there are other white space characters used (e.g. linefeed).

wking: > Thought we would leave out the spaces in case there are other white space characters used (e.g. linefeed).

But that applies to all template whitespace, no?  I think linefeed -> space normalization is preprocessing that tools need to do before comparing with the template.

wking:
Can somebody with push rights remove this branch now that it's merged?  There are a few other merged-but-not-deleted branches as well:

```console
$ git branch -a --merged | grep origin/ | grep -v 'master\|/pr/'
  remotes/origin/bsd-3-varient
  remotes/origin/fixodbl
  remotes/origin/improvemit
  remotes/origin/old-format
  remotes/origin/release-license-data
  remotes/origin/update-tools-version
```

goneall:
> Can somebody with push rights remove this branch now that it's merged? 

Done

PR #651, opened 2018-06-05
README: Make an explicit commitment to case-insensitive uniqueness

wking:
Encouraging case preservation while allowing for case-insentive comparison matches the spec position discussed in spdx/spdx-spec#63.  Note that this is just the list commitment.  It *allows* the spec, tools, and other list consumers to decide to be case-insensitive, but does not require them to be either case-sensitive or case-insensitive.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,12 @@ Please note that the format for this repository is internal to the SPDX legal te
 
 In contrast, the output files in the [SPDX license-list-data repository](https://github.com/spdx/license-list-data) are stable and well-supported, and make the License List available in RDFa, HTML, text, and JSON formats. You can use [SPDX tools](https://github.com/spdx/tools) (or create your own) to consume the supported formats of the license list.
 
+# Case insensitivity
+
+This project commits to never, in any past or future version, contain identifiers which differ only in case but have different semantics.
+For example, the presence of an [`MIT`](src/MIT.xml) identifier in the current list means that there will never be an `Mit` identifier which matches text not matched by the `MIT` identifier.
+List consumers are enouraged to use the canonical identifier casing, but this uniqueness commitment ensures that case-insensitive comparison with listed identifiers will be unambiguous.
+
 # How to contribute/participate
 
 We welcome contribution!  See [our contribution documentation](CONTRIBUTING.md) for details.


jlovejoy:
discussed on call June 14th. Good thing to clarify and really is about over-arching goal of not duplicating identifiers (in any way). But readme is not the right place, this should be part of field description here: https://spdx.org/spdx-license-list/license-list-overview proposed changes to be made via mailing list and will close this PR once changes are made on website.

wking:
> But readme is not the right place, this should be part of field description here: https://spdx.org/spdx-license-list/license-list-overview proposed changes to be made via mailing list and will close this PR once changes are made on website.

If this gets documented somewhere else, I think we need to link that commitment from both this repo and the wording in spdx/spdx-spec#37.  I prefer versioned locations for links like that, because they are more resistant to link rot.  But this particular issue isn't *that* important, so some link-rot exposure is acceptable.

And personally, I prefer our current in-repo docs for this repo (in the schema XSD) to unversioned external docs (#391).  I have no problem with anyone maintaining external docs about the local XML format and semantics, but think those will be much harder to keep current as our schema evolves.

zvr:
This repo does not exist in a vacuum; it is perfectly reasonable to assume that whoever access this has also access to the main SPDX website where things are explained.

wking:
> This repo does not exist in a vacuum; it is perfectly reasonable to assume that whoever access this has also access to the main SPDX website where things are explained.

I'm worried about discoverability, not access.  I've been fairly involved with the license list for several months, and feel like I have a fairly good grasp of this repo.  But while I know the wiki has team pages and or-later design discussion, I don't know what other information is there or how it's organized.  Perhaps some newcomers would guess to look there for this repo's copying statement, but I'd be very surprised if *all* newcomers guessed to look there.

wking:
Ahh, not sure why I thought this would go in the wiki.  I'm more comfortable with using the non-wiki portion of spdx.org, but still think it would be easier for all parties to have exerthing related to maintaining this repo live in or be linked from this repo.

The related list thread is [here](https://lists.spdx.org/g/Spdx-legal/message/2318).

jlovejoy:
closing as already covered in SPDX spec

Issue #652, opened 2018-06-13
Broken link


yoooooooooooooooooooooooo:
`http://opensource.linux-mirror.org/licenses/afl-2.0.txt` at https://spdx.org/licenses/AFL-2.0.html is broken.

jlovejoy:
we can probably just remove that since it was updated with the wayback archive of the same link

jlovejoy:
https://github.com/spdx/license-list-XML/pull/653

PR #653, opened 2018-06-14
remove old broken URL

jlovejoy:
we have wayback archive link now, so no need to have both broken link and wayback link
https://github.com/spdx/license-list-XML/issues/652

### DIFF
--- a/src/AFL-2.0.xml
+++ b/src/AFL-2.0.xml
@@ -2,7 +2,6 @@
 <SPDXLicenseCollection xmlns="http://www.spdx.org/license">
    <license isOsiApproved="true" licenseId="AFL-2.0" name="Academic Free License v2.0">
       <crossRefs>
-         <crossRef>http://opensource.linux-mirror.org/licenses/afl-2.0.txt</crossRef>
          <crossRef>http://wayback.archive.org/web/20060924134533/http://www.opensource.org/licenses/afl-2.0.txt</crossRef>
       </crossRefs>
     <text>


wking:
> we have wayback archive link now, so no need to have both broken link and wayback link

In a number of cases, it was an explicit legal-team decision to keep the old broken link without additional known-dead markup.  More discussion on the issue in general in #433.  In this specific case, the old, dead link is still there at @bradleeedmondson's [request][1].

[1]: https://github.com/spdx/license-list-XML/pull/419#issuecomment-329563093

jlovejoy:
how to deal with URLs that end up broken later is not really a critical issue to spend time on or have a specific policy on (which is probably why we haven't been entirely consistent). We never guaranteed that they'll always work (no one can for a third party link!), they are most useful when the license is first added and if we notice they end up broken, the it's good to fix them - whether be it via a new live link or using a web archive link. If people find this confusing or we get multiple people asking for an explanation, we can add a note on the info about the field here: https://spdx.org/spdx-license-list/license-list-overview

wking:
> how to deal with URLs that end up broken later is not really a critical issue to spend time on or have a specific policy on (which is probably why we haven't been entirely consistent).

I think adopting a generic policy would be less work than deciding (and re-deciding) on a per-link basis.  For instance, in the wake of this PR, I'm not sure if I should file a follow-up PR removing our other known-dead links.  Should I?  Or do we wait until others file per-link issues for each link?  Can I file those issues?  If not, why am I different from @yoooooooooooooooooooooooo?

goneall:
If we were to create a policy, there are actually some use cases where dead links can be of value:

- Historical references to original license websites that no longer exist
- A hint when matching license references (there are occasions where a copyright owner will only use a URL and a name in a source text file)

goneall:
BTW - Is there an efficient way to test for links programmaticly?  We could test for this when generating the license data and produce a warning.  Testing the links by attempting to fetch data would be too slow.

bradleeedmondson:
For the reasons @goneall gives above, I think that it would be worthwhile to make an explicit decision about this and say, at least, what our target is (even if we aren't able to update all existing files to comply right away). My current thinking would be to do the following:

1. Include license author's original URL
2. If current URL goes stale, add best available URL we can find at the time
3. Do not delete stale URLs (that is, don't delete/replace solely because they are stale -- if we made a mistake in the original then we would still delete or replace)

We could later decide whether to stick with the current, implicit, approach that "all crossrefs get equivalent markup, even if we know one has gone stale," or adopt known-dead (boolean) or last-known-good (timestamp) markup to differentiate the stale links.

goneall:
> or adopt known-dead (boolean) or last-known-good (timestamp) markup to differentiate the stale links

I like the idea of marking known dead links in some fashion.  Propose we add an optional attribute `dead` default to false on the `crossRef` element to indicate a dead link.  We could then check this when generating the license list HTML and not hyperlink any dead links.

wking:
Any chance we can move the "what should our policy be" discussion into #433? ;)

jlovejoy:
we will discuss and close this out on the call today - and note the outcome in #433 :)

PR #654, opened 2018-06-15
match philosophy or licenses in path of why-not-lgpl url

mlinksva:
changed by publisher (FSF) between
http://web.archive.org/web/20170908053542/http://www.gnu.org/licenses/gpl-3.0.txt
and
http://web.archive.org/web/20170930072152/http://www.gnu.org/licenses/gpl-3.0.txt

noted by @ale5000-git at
https://github.com/github/choosealicense.com/pull/581#issuecomment-397477772


jlovejoy:
discussed on June 28th call: acknowledged that could just alt tag around whole URL, but this works and is fine, so will merge.

PR #656, opened 2018-06-19
Add bullet around section 4.4

goneall:
Resolves a very visible formatting issue.  Also, replace the test file with the text prescribed form the original source.

### DIFF
--- a/src/ODbL-1.0.xml
+++ b/src/ODbL-1.0.xml
@@ -7,9 +7,11 @@
       </crossRefs>
     <text>
       <titleText>
+	     <optional>##</optional>
          <p>ODC Open Database License (ODbL)</p>
       </titleText>
       <optional>
+	     <optional>###</optional>
          <p>Preamble</p>
          <p>The Open Database License (ODbL) is a license agreement intended to allow users to freely share, modify,
          and use this Database while maintaining this same freedom for others. Many databases are covered by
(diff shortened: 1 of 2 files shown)


test/simpleTestForGenerator/ODbL-1.0.txt
```diff
@@ -1,15 +1,3 @@
-ODC Open Database License (ODbL)
-
-Preamble
```
wking: How did you generate this new content?  When I run:

```console
$ curl -s https://opendatacommons.org/files/2018/02/odbl-10.txt | dos2unix >test/simpleTestForGenerator/ODbL-1.0.txt
```

I still get the preamble.

goneall: I cut out the preamble.  After this is merged, I can create a separate PR with the preamble.

I don't want to confuse this PR too much by adding in optional text for the preamble in addition to the more important fix of the section formatting.  

I included the text since I needed to update the test file anyway and thought I should just go with the original source minus the preamble.

goneall: When I copy the text verbatim, the match fails.  I'll need to investigate why.

src/ODbL-1.0.xml
```diff
@@ -7,9 +7,11 @@
       </crossRefs>
     <text>
       <titleText>
+	     <optional>##</optional>
          <p>ODC Open Database License (ODbL)</p>
       </titleText>
       <optional>
+	     <optional>###</optional>
```
wking: I would prefer `<alt name="header2" match="#*" />` here and, with `header1`, above to keep the Markdown out of our canonical HTML.  More on this approach in #429, #623, etc.

goneall: Good idea, but currently don't have the time to update.  I'll merge in this version, feel free to submit a PR for the next release.

wking: > ... feel free to submit a PR for the next release.

Submitted as #661.  Any reason that couldn't land in this release?

PR #657, opened 2018-06-19
Add alt around the copyright text

goneall:

This would help anyone using the SPDX matching templates for most common MIT uses.

### DIFF
--- a/src/MIT.xml
+++ b/src/MIT.xml
@@ -9,7 +9,9 @@
          <p>MIT License</p>
       </titleText>
       <copyrightText>
-         <p><alt match=".+" name="copyright">Copyright (c) &lt;year&gt; &lt;copyright holders&gt;</alt></p>
+         <p>Copyright (c) 
+		    <alt match=".+" name="copyright">&lt;year&gt; &lt;copyright holders&gt;</alt>
+	     </p>
       </copyrightText>
     
       <p>Permission is hereby granted, free of charge, to any person obtaining a copy of this software and
--- a/src/MIT.xml
+++ b/src/MIT.xml
@@ -9,7 +9,7 @@
          <p>MIT License</p>
       </titleText>
       <copyrightText>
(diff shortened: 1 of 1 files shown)


src/MIT.xml
```diff
@@ -9,7 +9,7 @@
          <p>MIT License</p>
       </titleText>
       <copyrightText>
-         <p>Copyright (c) &lt;year&gt; &lt;copyright holders&gt;</p>
+         <p><alt match=".+" name="copyright">Copyright (c) &lt;year&gt; &lt;copyright holders&gt;</alt></p>
```
wking: This seems overly broad.  Can we follow [the BSD-2-Clause example][1] and keep "Copyright (c) " required?

[1]: https://github.com/spdx/license-list-XML/blob/v3.1/src/BSD-2-Clause.xml#L10-L11

goneall:
@wking Good point - I'll update the PR

src/MIT.xml
```diff
@@ -9,7 +9,9 @@
          <p>MIT License</p>
       </titleText>
       <copyrightText>
-         <p>Copyright (c) &lt;year&gt; &lt;copyright holders&gt;</p>
+         <p>Copyright (c) 
+		    <alt match=".+" name="copyright">&lt;year&gt; &lt;copyright holders&gt;</alt>
+	     </p>
```
wking: Your indent here, and on the previous line, is a strange mix of tabs and spaces.  Since the old `MIT.xml` contained no tabs, you may want to stick with just spaces here.

goneall: I was matching the style of the BSD license.  Feel free to create a PR against this branch, otherwise it is good enough from my perspective.

wking: > Feel free to create a PR against this branch...

Ok, will do.

wking: > > Feel free to create a PR against this branch...
>
> Ok, will do.

Missed this branch (sorry for the delay), so I've filed the fixup as #662.

Issue #646, opened 2018-05-17
New Exception Request: Google patent grant


jlovejoy:
from email: 

Google uses a standard "Additional IP Rights Grant (Patents)" text in a lot of projects they've been publishing - {} sections are mine:

'''
Additional IP Rights Grant (Patents)

"These implementation{s}" means the copyrightable works {that implement the Product} distributed by Google as part of the {Project}.

Google hereby grants to you a perpetual, worldwide, non-exclusive, no-charge, royalty-free, irrevocable (except as stated in this section) patent license to make, have made, use, offer to sell, sell, import, transfer, and otherwise run, modify and propagate the contents of these implementations of {Product}, where such license applies only to those patent claims, both currently owned by Google and acquired in the future, licensable by Google that are necessarily infringed by these implementations of {Product}. This grant does not include claims that would be infringed only as a consequence of further modification of these implementations. If you or your agent or exclusive licensee institute or order or agree to the institution of patent litigation or any other patent enforcement activity against any entity (including a cross-claim or counterclaim in a lawsuit) alleging that any of these implementations of {Product} or any code incorporated within any of these implementations of {Product} constitutes direct or contributory patent infringement, or inducement of patent infringement, then any patent rights granted to you under this License for these implementations of {Product} shall terminate as of the date such litigation is filed.
'''

Sources: https://webrtc.org/license/additional-ip-grant/,  https://www.webmproject.org/license/additional/, https://github.com/ImageMagick/webp/blob/master/PATENTS, https://golang.org/PATENTS

Discussed on May 17th call. need to determine what markup is needed and title (then make PR)

bradleeedmondson:
To clarify, discussed today and accepted for inclusion on "Exceptions" list.

swinslow:
I'll prepare the files for this one for 3.2.

swinslow:
Jilayne and Bradlee - I've taken a closer look at the Google patent exception. It turns out that they've actually used several slightly different versions, with substantive differences. E.g., WebRTC's doesn't include "royalty-free"; WebM has a broader scope of conditions that trigger termination; etc.

I'll prepare separate versions for each different version that I've found. If this warrants further discussion on a future call in light of the differences, we might consider pushing it out past 3.2.

jlovejoy:
[...]

PR #658, opened 2018-06-26
Add XML and tests for three Google Additional IP Rights Grant statements

swinslow:
This adds XML and test files for three "Additional IP Rights Grant" statements from Google, intended to address #646. 

As discussed in that thread, I've included three separate files because each one has slightly different wording, in ways that I believe are substantively different and not suitable to handle just via matching guidelines and templating. 

I have added templating to the project names within each XML file, so that these could be reused with other Google projects that have the same base text.


swinslow:
Following from the conversation on yesterday's legal team call, I believe it was agreed that these unilateral patent grants from Google are not "exceptions" for purposes of the SPDX Exception List. So I am closing this PR. I believe this should also likely close #646.

Kongpcmail:
> This adds XML and test files for three "Additional IP Rights Grant" statements from Google, intended to address #646.
> 
> As discussed in that thread, I've included three separate files because each one has slightly different wording, in ways that I believe are substantively different and not suitable to handle just via matching guidelines and templating.
> 
> I have added templating to the project names within each XML file, so that these could be reused with other Google projects that have the same base text.
> 
> Signed-off-by: Steve Winslow [swinslow@gmail.com](mailto:swinslow@gmail.com)

@othree

PR #659, opened 2018-06-26
re-add old (stale) URL for Imlib2

bradleeedmondson:
re-add old (stale) URL for imlib2, per #573 (implicit policy of retaining stale URLs)

### DIFF
--- a/src/Imlib2.xml
+++ b/src/Imlib2.xml
@@ -2,13 +2,14 @@
 <SPDXLicenseCollection xmlns="http://www.spdx.org/license">
    <license isOsiApproved="false" licenseId="Imlib2" name="Imlib2 License">
       <crossRefs>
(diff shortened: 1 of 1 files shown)


wking:
I think this case is a bit trickier.  Things like wayback links have a different authority (the wayback folks seem reliable, but still), so I understand keeping known-dead author links around when adding a working wayback link.  In this case, #573 replaced a link into the imlib2 source repo with a new link into the imlib2 source repo, both of which are under enlightenment.org.  I think that sort of same-content, same-controlling-authority link change can happen without retaining the known-dead link.  If that makes sense, I think we should close this PR without merging it.

I'll comment in #433 about addressing cases like this in a general policy.

wking:
> I'll comment in #433 about addressing cases like this in a general policy.

Heh, turns out I'd [already done this][1] :p.

[1]: https://github.com/spdx/license-list-XML/issues/433#issuecomment-354332391

Issue #649, opened 2018-05-31
New exception: OCaml LGPL linking exception


bradleeedmondson:
OCaml linking exception reported by Andy Li on the SPDX mailing list:

> Full Name:
> OCaml linking exception
> 
> Short Identifier:
> OCaml-linking-exception
> 
> Reference to exception text:
> https://github.com/ocaml/opam/blob/master/LICENSE
> 
> Exception text is attached as "OCaml-linking-exception.txt". Notice that individual usages in libraries have slightly difference mainly due to differences in software names and author names.
> 
> I haven't noticed any information on whether this is OSI-approved.
> 
> The exception is used by OCaml itself and is very popular among OCaml libraries:
> 
>  * OCaml: https://github.com/ocaml/ocaml/blob/trunk/LICENSE
>  * opam: https://github.com/ocaml/opam/blob/master/LICENSE
>  * OCaml batteries-included: https://github.com/ocaml-batteries-team/batteries-included/blob/master/LICENSE
> 
> Best regards,
> Andy

bradleeedmondson:
Text of the attached document:

> As a special exception to the GNU Lesser General Public License, you
> may link, statically or dynamically, a "work that uses the Library"
> with a publicly distributed version of the Library to produce an
> executable file containing portions of the Library, and distribute
> that executable file under terms of your choice, without any of the
> additional requirements listed in clause 6 of the GNU Lesser General
> Public License. By "a publicly distributed version of the Library", we
> mean either the unmodified Library as distributed, or a modified
> version of the Library that is distributed under the conditions
> defined in clause 2 of the GNU Lesser General Public License. This
> exception does not however invalidate any other reasons why the
> executable file might be covered by the GNU Lesser General Public
> License.

bradleeedmondson:
I can write this up as XML

bradleeedmondson:
Wrote up XML to add this in PR #660, but we still have questions for the license authors/submitters, e.g. here:
https://lists.spdx.org/g/Spdx-legal/topic/22080605#1338


Removed from milestone 3.2 since this is not ready for decision-making

jlovejoy:
[...]

PR #660, opened 2018-06-28
add OCaml linking exception to LGPL

bradleeedmondson:
add OCaml linking exception to LGPL to address #649 

Note: Legal team still need to officially decide whether to add this to the exceptions list.

### DIFF
--- a/src/exceptions/OCaml-LGPL-linking-exception.xml
+++ b/src/exceptions/OCaml-LGPL-linking-exception.xml
@@ -16,24 +16,25 @@
               As a special exception to the GNU <alt name="libraryTypo"
               match="Library|Lesser">Lesser</alt> General Public License, you
               may link, statically or dynamically, a "work that uses <alt
-              name="linkedSoftwareName1" match="the OCaml Core System|opam">
-              the OCaml Core System</alt>with a publicly distributed version of
-              <alt name="linkedSoftwareName2"
-              match="the OCaml Core System|opam">the OCaml Core System</alt> to
-              produce an executable file containing portions of
+              name="linkedSoftwareName1"
+              match="the OCaml Core System|opam|the Library">the OCaml Core
+              System</alt>with a publicly distributed version of <alt
+              name="linkedSoftwareName2"
+              match="the OCaml Core System|opam|the Library">the OCaml Core
+              System</alt> to produce an executable file containing portions of
               <alt name="linkedSoftwareName3"
-              match="the OCaml Core System|opam">the OCaml Core System</alt>,
-              and distribute that executable file under terms of your choice,
-              without any of the additional requirements listed in clause 6 of
-              the GNU Lesser General Public License. By "a publicly distributed
-              version of <alt name="linkedSoftwareName4"
-              match="the OCaml Core System|opam">
-              the OCaml Core System</alt>", we mean either the unmodified <alt
-              name="linkedSoftwareName5" match="OCaml Core System|opam">
+              match="the OCaml Core System|opam|the Library">the OCaml Core
+              System</alt>, and distribute that executable file under terms of
+              your choice, without any of the additional requirements listed in
+              clause 6 of the GNU Lesser General Public License. By "a publicly
+              distributed version of <alt name="linkedSoftwareName4"
+              match="the OCaml Core System|opam|the Library">the OCaml Core
+              System</alt>", we mean either the unmodified <alt
+              name="linkedSoftwareName5" match="OCaml Core System|opam|Library">
               OCaml Core System</alt> as distributed by <alt name="distributor"
               match="INRIA|OCamlPro">INRIA</alt>, or a modified version of <alt
               name="linkedSoftwareName6"
-              match="the OCaml Core System|opam|the opam">
+              match="the OCaml Core System|opam|the opam|the Library">
               the OCaml Core System</alt> that is distributed under the
               conditions defined in clause 2 of the GNU Lesser General Public
               License. This exception does not however invalidate any other
--- a/src/exceptions/OCaml-LGPL-linking-exception.xml
+++ b/src/exceptions/OCaml-LGPL-linking-exception.xml
@@ -12,8 +12,8 @@
          <p>OCaml LGPL Linking Exception</p>
       </titleText>
       <p>
-        As a special exception to the GNU <alt name="lesser" match="Library|Lesser">Lesser</alt> 
-        General Public License, you may
+        As a special exception to the GNU <alt name="lesser" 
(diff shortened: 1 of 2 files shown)


src/exceptions/OCaml-LGPL-linking-exception.xml
```diff
@@ -0,0 +1,37 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<SPDXLicenseCollection xmlns="http://www.spdx.org/license">
+   <exception licenseId="OCaml-LGPL-Linking-Exception" name="OCaml LGPL Linking Exception">
```
wking: You should probably set [`listVersionAdded`][1] (more docs [here][2]) with a FIXME value or some such, so we don't forget to set that before merging this PR.  The attribute is not required, but I expect that's so we can grandfather in some older entries (where we don't remember when they were added).

[1]: https://github.com/spdx/license-list-XML/blob/v3.1/schema/ListedLicense.xsd#L53
[2]: https://github.com/spdx/license-list-XML/blob/v3.1/schema/ListedLicense.xsd#L135

src/exceptions/OCaml-LGPL-linking-exception.xml
```diff
@@ -0,0 +1,37 @@
...
+      <crossRefs>
+         <crossRef>https://caml.inria.fr/ocaml/license.en.html</crossRef>
+      </crossRefs>
+      <notes>
+        Adopted by OCaml core in 2001 here: https://github.com/ocaml/ocaml/commit/02ef950033b81fe371759f024faa55f361ba83a6#diff-9879d6db96fd29134fc802214163b95a (git-svn-id: http://caml.inria.fr/svn/ocaml/trunk@4146 f963ae5c-01c2-4b8c-9fe0-0dff7051ff02)
```
wking: `<notes>` [takes `formattedFixedTextType`][1].  Do we want to add a link element (`a`) [here][2] to make this sort of thing more readable?

[1]: https://github.com/spdx/license-list-XML/blob/v3.1/schema/ListedLicense.xsd#L47
[2]: https://github.com/spdx/license-list-XML/blob/v3.1/schema/ListedLicense.xsd#L348-L355

src/exceptions/OCaml-LGPL-linking-exception.xml
```diff
@@ -0,0 +1,37 @@
...
+    <text>
+      <titleText>
+         <p>OCaml LGPL Linking Exception</p>
+      </titleText>
+      <p>
+        In the following, "the OCaml Core System" refers to all files marked
+        "Copyright INRIA" in this distribution.
+
+        The OCaml Core System is distributed under the terms of the GNU Lesser
+        General Public License (LGPL) version 2.1 (included below).
```
wking: Neither this paragraph nor the previous paragraph match the text [here][1].  I think the exception is just the following paragraph, and that we could match this exception across versions with suitable `<alt>` blocks in that paragraph.

[1]: https://github.com/ocaml/ocaml/commit/02ef950033b81fe371759f024faa55f361ba83a6#diff-9879d6db96fd29134fc802214163b95aR1

src/exceptions/OCaml-LGPL-linking-exception.xml
```diff
@@ -0,0 +1,37 @@
...
+
+        As a special exception to the GNU Lesser General Public License, you may
```
wking: The original text for this exception [has "Library"][1] instead of "Lesser".  I recommend:

```xml
As a special exception to the GNU <alt name="lesser" match="Library|Lesser">Lesser</alt> General Public License, you may
```

[1]: https://github.com/ocaml/ocaml/commit/02ef950033b81fe371759f024faa55f361ba83a6#diff-9879d6db96fd29134fc802214163b95aR15

src/exceptions/OCaml-LGPL-linking-exception.xml
```diff
@@ -0,0 +1,37 @@
...
+        link, statically or dynamically, a "work that uses the OCaml Core System"
```
wking: To support both the [current][1] "the OCaml Core System" and the [original][2] "the Library", I recommend:

```xml
link, statically or dynamically, a "work that uses <alt name="work1" match=".*">the OCaml Core System</alt>"
```

and similarly below (with `work2`, `work3`, etc.) for all locations of "the OCaml Core System".

[1]: https://caml.inria.fr/ocaml/license.en.html
[2]: https://github.com/ocaml/ocaml/commit/02ef950033b81fe371759f024faa55f361ba83a6#diff-9879d6db96fd29134fc802214163b95aR16

src/exceptions/OCaml-LGPL-linking-exception.xml
```diff
@@ -0,0 +1,37 @@
...
+        with a publicly distributed version of the OCaml Core System to produce
+        an executable file containing portions of the OCaml Core System, and
+        distribute that executable file under terms of your choice, without any
+        of the additional requirements listed in clause 6 of the GNU Lesser
+        General Public License. By "a publicly distributed version of the OCaml
+        Core System", we mean either the unmodified OCaml Core System as
+        distributed by INRIA, or a modified version of the OCaml Core System
+        that is distributed under the conditions defined in clause 2 of the GNU
```
wking: This was originally [clause 3 of the LGPL-2.0][1].  That seems like a typo, since the LGPL-2.0's [section 2][2] is about modification and distribution, while its [section 3][3] is about transferring to the GPL-2.0-only.  I'm not sure how long that typo'ed version was in circulation, or whether supporting any typo'ed versions is worth the legal ambiguity of allowing "clause 3" here.

[1]: https://github.com/ocaml/ocaml/commit/02ef950033b81fe371759f024faa55f361ba83a6#diff-9879d6db96fd29134fc802214163b95aR24
[2]: https://github.com/spdx/license-list-XML/blob/v3.1/src/LGPL-2.0.xml#L210-L273
[3]: https://github.com/spdx/license-list-XML/blob/v3.1/src/LGPL-2.0.xml#L274-L293

src/exceptions/OCaml-LGPL-linking-exception.xml
bradleeedmondson: Maybe a good idea, but this is a bigger issue for the schema. <wontfix> as far as this PR is concerned.

bradleeedmondson:
Build test failed, but I can't tell whether that's my commit or the environment. I'm assuming 3.3 is on hold until we figure this out.

cc: @goneall @jlovejoy @wking

goneall:
@bradleeedmondson @jlovejoy @wking Just pushed a fix - you should be able to merge all PR's.

To fix the errors in your own branch, you need to merge in the fix from PR #704

PR #661, opened 2018-06-29
ODbL-1.0: Remove ## header markup from canonical text

wking:
This markup makes sense in plain-text renderings to denote headers.  But HTML has other ways to represent headers (`<h1>`, etc.), as seen [in the canonical upstream source][1].  This comment removes those entries from the canonical text (which we will eventually render as HTML at [2]), while still allowing us to match [upstream's canonical plain-text version][3].

Previous discussion [here][4].  There's also a precedent for this sort of thing in #429 and #623.

[1]: https://opendatacommons.org/licenses/odbl/1.0/
[2]: https://spdx.org/licenses/ODbL-1.0.html
[3]: https://opendatacommons.org/files/2018/02/odbl-10.txt
[4]: https://github.com/spdx/license-list-XML/pull/656#discussion_r197598522

### DIFF
--- a/src/ODbL-1.0.xml
+++ b/src/ODbL-1.0.xml
@@ -7,11 +7,11 @@
       </crossRefs>
     <text>
       <titleText>
-	     <optional>##</optional>
+        <alt name="header1" match="#*"/>
          <p>ODC Open Database License (ODbL)</p>
       </titleText>
       <optional>
-	     <optional>###</optional>
+        <alt name="header2" match="#*"/>
          <p>Preamble</p>
          <p>The Open Database License (ODbL) is a license agreement intended to allow users to freely share, modify,
          and use this Database while maintaining this same freedom for others. Many databases are covered by


src/ODbL-1.0.xml
```diff
@@ -7,11 +7,11 @@
       </crossRefs>
     <text>
       <titleText>
-	     <optional>##</optional>
+        <alt match="header1" match="#*"/>
```
bradleeedmondson: Probably intended name="header1".

Two elements called match is illegal XML, causing travis build fail:
https://travis-ci.org/spdx/license-list-XML/builds/398415897

wking: > Probably intended name="header1".

Oops, thanks.  Fixed with 2729eed -> 23d2671.

goneall:
@bradleeedmondson You show up as leaving review comments, but for some reason I can't view them from github.

@wking The changes look good to me, but I just want to check with @bradleeedmondson on the review before merging

goneall:
@bradleeedmondson @wking Found the comments in email.  Looks like all the issues are resolved.  I'll go ahead and merge this in for the next release.

PR #663, opened 2018-07-05
New license: ODC-By-1.0

jlovejoy:
text test file and xml file

### DIFF
--- a/src/ODC-By-1.0.xml
+++ b/src/ODC-By-1.0.xml
@@ -1,0 +1,504 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<SPDXLicenseCollection xmlns="http://www.spdx.org/license">
+   <license isOsiApproved="false" licenseId="ODC-By-1.0"
+            name="Open Data Commons Attribution License v1.0">
+      <crossRefs>
+         <crossRef>https://opendatacommons.org/licenses/by/1.0/</crossRef>
+      </crossRefs>
+    <text>
+      <titleText>
+	     <optional>##</optional>
+         <p>ODC Attribution License (ODC-By)</p>
+      </titleText>
+      <optional>###</optional>
+      <p>Preamble</p>
+      <p>The Open Data Commons Attribution License is a license agreement intended to allow users to freely
+      share, modify, and use this Database subject only to the attribution requirements set out in Section 4.</p>
+      <p>Databases can contain a wide variety of types of content (images, audiovisual material, and sounds
+      all in the same database, for example), and so this license only governs the rights over the Database,
+      and not the contents of the Database individually. Licensors may therefore wish to use this license
+      together with another license for the contents.</p>
+      <p>Sometimes the contents of a database, or the database itself, can be covered by other rights not
+      addressed here (such as private contracts, trademark over the name, or privacy rights / data protection
+      rights over information in the contents), and so you are advised that you may have to consult other
+      documents or clear other rights before doing activities not covered by this License.</p>
+      <p>------</p>
+    
+      <p>The Licensor (as defined below)</p>
+      <p>and</p>
+      <p>You (as defined below)</p>
+      <p>agree as follows:</p>
+      <list>
+        <item>
+            <bullet>1.0</bullet>
+          Definitions of Capitalised Words
+          <p>"Collective Database" - Means this Database in unmodified form as part of a
+             collection of independent databases in themselves that together are assembled into a
+             collective whole. A work that constitutes a Collective Database will not be considered a
+             Derivative Database.</p>
+            <p>"Convey" - As a verb, means Using the Database, a Derivative Database, or the
+             Database as part of a Collective Database in any way that enables a Person to make or receive
+             copies of the Database or a Derivative Database. Conveying does not include interaction with a
+             user through a computer network, or creating and Using a Produced Work, where no transfer of a
+             copy of the Database or a Derivative Database occurs.</p> 
+             <p>  "Contents" - The
+             contents of this Database, which includes the information, independent works, or other
+             material collected into the Database. For example, the contents of the Database could be
+             factual data or works such as images, audiovisual material, text, or sounds.</p>
+            <p>"Database" - A collection of material (the Contents) arranged in a systematic or
+             methodical way and individually accessible by electronic or other means offered under the
+             terms of this License.</p>
+            <p>"Database Directive" - Means Directive 96/9/EC of the European Parliament and of
+             the Council of 11 March 1996 on the legal protection of databases, as amended or
+             succeeded.</p>
+            <p>"Database Right" - Means rights resulting from the Chapter III ("sui
+             generis") rights in the Database Directive (as amended and as transposed by member
+             states), which includes the Extraction and Re-utilisation of the whole or a Substantial part
+             of the Contents, as well as any similar rights available in the relevant jurisdiction under
+             Section 10.4.</p>
+            <p>"Derivative Database" - Means a database based upon the Database, and includes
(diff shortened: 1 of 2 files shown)


src/ODC-By-1.0.xml
```diff
@@ -0,0 +1,504 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<SPDXLicenseCollection xmlns="http://www.spdx.org/license">
+   <license isOsiApproved="false" licenseId="ODC-By-1.0"
+            name="Open Data Commons Attribution License v1.0">
```
wking: I think you want [`listVersionAdded="3.2"`][1] in as well.

[1]: https://github.com/spdx/license-list-XML/blob/v3.1/schema/ListedLicense.xsd#L131-L139

src/ODC-By-1.0.xml
```diff
@@ -0,0 +1,504 @@
...
+      <crossRefs>
+         <crossRef>https://opendatacommons.org/licenses/by/1.0/</crossRef>
+      </crossRefs>
+    <text>
+      <titleText>
+	     <optional>##</optional>
```
wking: I think we should use `<alt>` for these.  For an example of a similar change, see #661.

src/ODC-By-1.0.xml
```diff
@@ -0,0 +1,504 @@
...
+         <p>ODC Attribution License (ODC-By)</p>
+      </titleText>
+      <optional>###</optional>
+      <p>Preamble</p>
```
wking: Is the preamble optional?

src/ODC-By-1.0.xml
```diff
@@ -0,0 +1,504 @@
...
+      <p>Sometimes the contents of a database, or the database itself, can be covered by other rights not
+      addressed here (such as private contracts, trademark over the name, or privacy rights / data protection
+      rights over information in the contents), and so you are advised that you may have to consult other
+      documents or clear other rights before doing activities not covered by this License.</p>
+      <p>------</p>
+    
+      <p>The Licensor (as defined below)</p>
+      <p>and</p>
+      <p>You (as defined below)</p>
+      <p>agree as follows:</p>
```
wking: I think this is all one paragraph.  I'd rather not force newlines, but if you want to, I think you should use [`<br/>`][1].

Also, should this text be optional?  It looks like it's "I'm about to start in on the license text", not like actual license text itself.

[1]: https://github.com/spdx/license-list-XML/blob/v3.1/schema/ListedLicense.xsd#L399-L407

src/ODC-By-1.0.xml
```diff
@@ -0,0 +1,504 @@
...
+          <p>"Collective Database" - Means this Database in unmodified form as part of a
+             collection of independent databases in themselves that together are assembled into a
+             collective whole. A work that constitutes a Collective Database will not be considered a
+             Derivative Database.</p>
+            <p>"Convey" - As a verb, means Using the Database, a Derivative Database, or the
+             Database as part of a Collective Database in any way that enables a Person to make or receive
+             copies of the Database or a Derivative Database. Conveying does not include interaction with a
+             user through a computer network, or creating and Using a Produced Work, where no transfer of a
+             copy of the Database or a Derivative Database occurs.</p> 
+             <p>  "Contents" - The
```
wking: Doesn't really matter, but you may want to remove the space from `<p> "Contents"` for consistency with the other entries.

test/simpleTestForGenerator/ODC-By-1.0.txt
```diff
@@ -0,0 +1,195 @@
+# ODC Attribution License (ODC-By)
+
+### Preamble
+
+The Open Data Commons Attribution License is a license agreement intended to allow users to freely share, modify, and use this Database subject only to the attribution requirements set out in Section 4.
```
wking: This text seems to lack some line wrapping from [their canonical text/plain form][1].  To make it easier to verify that this is in fact their upstream text, can we generate this file programatically, with something like:

```console
$ curl -s https://opendatacommons.org/files/2018/02/odc_by_1.0_public_text.txt | dos2unix | sed 's/[[:space:]]*$//' >test/simpleTestForGenerator/ODC-By-1.0.txt
```

[1]: https://opendatacommons.org/files/2018/02/odc_by_1.0_public_text.txt

src/ODC-By-1.0.xml
jlovejoy: done (and just realized looking back at 3.1 release licenses that I don't think we remembered to do this... )

src/ODC-By-1.0.xml
jlovejoy: discussed on call as could go either way - this was what Gary submitted and it's works, so leaving as is (and copied same method here) unless it's so annoying or causes problems that someone wants to update :)

wking: > ... unless it's so annoying or causes problems that someone wants to update :)

I'm fine filing a follow-up update like #661.

src/ODC-By-1.0.xml
jlovejoy: we generally don't mark the preamble as optional unless there is a good reason or definitive indication from the license steward/author to do so.

src/ODC-By-1.0.xml
jlovejoy: it's essentially how it appears in the original: https://opendatacommons.org/licenses/by/1.0/ and https://opendatacommons.org/files/2018/02/odc_by_1.0_public_text.txt

test/simpleTestForGenerator/ODC-By-1.0.txt
jlovejoy: I'm not sure I know what you mean - but line wrapping doesn't matter/need to match the canonical in any case.

src/ODC-By-1.0.xml
wking: > it's essentially how it appears in the original...

Using `<br/>` allows you to reproduce that appearance without claiming that this is four separate paragraphs.

jlovejoy: it doesn't matter, so let's not waste time on it - it's fine as is :)

test/simpleTestForGenerator/ODC-By-1.0.txt
wking: > I'm not sure I know what you mean...

You can probably get a similar effect by copy/pasting [their canonical text/plain form][1] from your browser.

> ... but line wrapping doesn't matter/need to match the canonical in any case.

Right, so why would we have line-wrapping preferences of our own, instead of just dropping in the text as provided by upstream (with the unix LF line endings for consistency with the rest of our samples and removing their trailing whitespace to make `git show --check` happy)?  By sticking with their wrapping, we make it easier for folks to check our text against text asserted by opendatacommons.org.  Assuming upstream doesn't drastically re-wrap, this also makes for less churn if/when we update to keep up with their changes (e.g. #600, #608).

[1]: https://opendatacommons.org/files/2018/02/odc_by_1.0_public_text.txt

src/ODC-By-1.0.xml
wking: > it doesn't matter, so let's not waste time on it - it's fine as is :)

It matters to me, but I'm fine filing a follow-up PR ;).

test/simpleTestForGenerator/ODC-By-1.0.txt
jlovejoy: I believe the text .txt files have no wrapping, and I used that to then create the xml file (roughly, it is a bit onerous, the whole process) and then wrapped by hand where needed. considering white space does not matter for matching, I don't think this matters either. let's leave it be and focus on any substantive problems

Issue #616, opened 2018-03-06
New Exception: URW Base 35 AGPL-3.0-only font exception


wking:
Adding a tracking issue for [the URW Base 35 license exception][1]:

> The font and related files in this directory are distributed under the GNU AFFERO GENERAL PUBLIC LICENSE Version 3 (see the file COPYING), with the following exemption:
>
> As a special exception, permission is granted to include these font programs in a Postscript or PDF file that consists of a document that contains text to be displayed or printed using this font, regardless of the conditions or license applying to the document itself.

which was [raised on the list in a 2018-02-28 mail from Stefan Brüns][2].

Can somebody with push access label this “[new license/exception request][3]”?

[1]: https://github.com/ArtifexSoftware/urw-base35-fonts/blob/65962e27febc3883a17e651cdb23e783668c996f/LICENSE
[2]: https://lists.spdx.org/pipermail/spdx-legal/2018-February/002478.html
[3]: https://github.com/spdx/license-list-XML/labels/new%20license%2Fexception%20request

bradleeedmondson:
Discussed on legal call today.

Appears to be used only in URW Base 35 Fonts package (google search for matching text shows only ~5 hits); however that package is included in Debian, so we will approve and add this exception to the SPDX License List. Approved.

Proposed name: AGPL-3.0 font exception PS-PDF

Proposed short identifier: AGPL-3.0-font-exception-ps-pdf

Note that the PostScript and PDF document format appears to be a condition for triggering this exception.

bradleeedmondson:
@wking Would you be able to file a PRs for the XML definition of this license?

wking:
> Proposed short identifier: AGPL-3.0-font-exception-ps-pdf

Do we want to stick a version number on that?  What happens if they update the text in the future in a way the SPDX considers legally distinct from the current text?

bradleeedmondson:
Ah, yes, good point. Now that you raise it, I think we should version just about everything we add to the list. (Perhaps we should make this an explicit directive going forward?) 

We always list version number when the license author specifies versions, but it's not completely clear what to do when the author doesn't version their text. Rather than specify 1.0 for them, perhaps we should version by date, as we do with [some](https://github.com/spdx/license-list-XML/blob/master/src/Adobe-2006.xml) [license](https://github.com/spdx/license-list-XML/blob/master/src/W3C-19980720.xml) [files](https://github.com/spdx/license-list-XML/blob/master/src/W3C-20150513.xml)? 

[...]

PR #664, opened 2018-07-06
PS/PDF font exception (2017-08-17)

bradleeedmondson:
Add URW Base 35 AGPL-3.0-only font exception to address #616 for milestone 3.2


src/exceptions/PS-or-PDF-font-exception-20170817.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<SPDXLicenseCollection xmlns="http://www.spdx.org/license">
+   <exception licenseId="PS-or-PDF-font-exception-20170817" name="PS/PDF font exception (2017-08-17)" listVersionAdded="3.2">
```
wking: If we're hyphenating the date in the name, do we want to hyphenate the date in the ID as well?  I'd hyphenate both for consistency, but don't feel strongly about it if you have your own preferences.

src/exceptions/PS-or-PDF-font-exception-20170817.xml
```diff
@@ -0,0 +1,21 @@
...
+      <crossRefs>
+        <crossRef>https://github.com/ArtifexSoftware/urw-base35-fonts/blob/master/LICENSE</crossRef>
+        <crossRef>https://github.com/ArtifexSoftware/urw-base35-fonts/blob/65962e27febc3883a17e651cdb23e783668c996f/LICENSE</crossRef>
```
wking: We don't have a policy on pinned vs. floating links, but there is some previous discussion [here][1].  As I said there, I prefer pinning the link when that's as easy as GitHub makes it.  But I don't think we need to have *both* pinned and floating links.  Folks can easily switch from the pinned link to a the floating version using GitHub's UI, without us having to link to the floating version directly.  I recommend keeping this line and dropping the previous `master` line.

[1]: https://github.com/spdx/license-list-XML/pull/512#discussion_r156771168

src/exceptions/PS-or-PDF-font-exception-20170817.xml
```diff
@@ -0,0 +1,21 @@
...
+      </crossRefs>
+      <notes>Specified to be used with APGL-3.0</notes>
```
wking: Now that we've deprecated `AGPL-3.0` do we want to spell out `AGPL-3.0-or-later or AGPL-2.0-only` here?  And do we want to soften this, based on [your][1]:

> I also agree with your markup approach, but I think we can take it a step further and use <alt> for the license name (rather than leaving the happenstance of AGPL-3.0 per this request), and knock AGPL-3.0 out of the exception name and identifier. This may not be technically necessary now, but seems to be the best way to plan for future use in conjunction with other licenses, which it's my understanding the exceptions are intended to support.

Is there anything AGPL-3.0-specific about the wording of the second paragraph?

[1]: https://github.com/spdx/license-list-XML/issues/616#issuecomment-382951343

src/exceptions/PS-or-PDF-font-exception-20170817.xml
```diff
@@ -0,0 +1,21 @@
...
+    <text>
+      <optional>
+        <p>The font and related files in this directory are distributed under the
+        GNU AFFERO GENERAL PUBLIC LICENSE Version 3 (see the file COPYING), with
+        the following exemption:</p>
```
wking: I think this paragraph is part of the standard license header, and not part of the exception text itself.  Did the legal team have time to discuss [my proposal to allow `<text>` inside `<standardLicenseHeader>`][1]?  If not, do we want to take the less-DRY alternative I floated there?

[1]: https://github.com/spdx/license-list-XML/issues/616#issuecomment-382835783

jlovejoy:
hey guys - we need to get 3.2 out, seems like there are a few nits here, but nothing earth shattering, can we merge this so it doesn't hold up the release?

src/exceptions/PS-or-PDF-font-exception-20170817.xml
jlovejoy: I think date is fine as is - the short identifier is already kind of long and the other ids we used dates in (W3C) don't have hyphens

src/exceptions/PS-or-PDF-font-exception-20170817.xml
jlovejoy: I think this is fine as is and consistent with the way we've been doing it. As for your proposal - we probably ought to slot that in for a discussion on an upcoming call

src/exceptions/PS-or-PDF-font-exception-20170817.xml
bradleeedmondson: The difference was to match previous IDs and titles, most of which had the ID as "YYYYMMDD" and the title as "(YYYY-MM-DD)" or "(YYYY)"

PR #672, opened 2018-07-18
Remove UPL-1.0 duplicate titleText

mlinksva:
I realize titleText is optional, so it doesn't hurt matching to have arbitrary titleTexts, but it looks silly eg at https://spdx.org/licenses/UPL-1.0.html and doesn't reflect actual use eg 

  - https://github.com/oracle/weblogic-kubernetes-operator/blob/master/LICENSE
  - https://github.com/oracle/docker-images/blob/master/LICENSE
  - https://github.com/oracle/vagrant-boxes/blob/master/LICENSE

### DIFF
--- a/src/UPL-1.0.xml
+++ b/src/UPL-1.0.xml
@@ -6,9 +6,6 @@
          <crossRef>http://opensource.org/licenses/UPL</crossRef>
       </crossRefs>
     <text>
-      <titleText>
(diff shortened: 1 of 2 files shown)


goneall:
There is a test file in test/simpleTestForGenerator/UPL-1.0.txt which also needs to be updated.  It includes the extra title text as well.

PR #679, opened 2018-08-08
(test) Sendmail: add alt tags

jlovejoy:
adding alt tags to Sendmail license (and testing XML Editor)


jlovejoy:
this was a test of the XML editor, but don't delete yet, as it may be useful... (but will need someone to check my alt tags potentially)

jlovejoy:
this was a test, so I think we can simply close it?

jlovejoy:
closing as this was only a test

PR #681, opened 2018-08-09
Test of SPDX License XML editor: mod BSD-2-Clause-FreeBSD

swinslow:
Just testing out use of the XML editor, to see what happens if I try to modify an existing license.


swinslow:
This was just a test of the XML editor -- I'm closing it.

PR #688, opened 2018-09-06
NASA-1.3: Add template tags for blank lines

### LINKED ISSUE #421: Decide whether to further templatize NASA-1.3 license
Recently merged the NASA-1.3 license in #137, but it has several sections that arguably should be templatized further, e.g. lines 18-20, line 65, etc. in 034386eb7380f49af58910504b3ec126ec8b4b34

We should decide here whether we want to do this, or not (as we might not, if we consider this more of a one-off or vanity license that is only used verbatim).

swinslow:
Fixes #421 


jlovejoy:
this looks good to me, but could @zvr do a check before merging? (@goneall  and @bradleeedmondson are both busy or out this week :)

PR #697, opened 2018-10-04
Add Sendmail-8.23 and test

### LINKED ISSUE #678: New License: Sendmail v8.23
FYI, the SPDX version of the Sendmail Liense is 8.16.  The current version is 8.23 which contains additional language.  I have attached both (see email)

Here is the original URL for the Sendmail License 8.23 text that Tom Incorvia provided: 

https://www.proofpoint.com/sites/default/files/sendmail-license.pdf

swinslow:
Fixes #678 

### DIFF
--- a/src/Sendmail-8.23.xml
+++ b/src/Sendmail-8.23.xml
@@ -1,0 +1,116 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<SPDXLicenseCollection xmlns="http://www.spdx.org/license">
+   <license isOsiApproved="false" licenseId="Sendmail-8.23" name="Sendmail License 8.23">
+      <crossRefs>
+         <crossRef>https://www.proofpoint.com/sites/default/files/sendmail-license.pdf</crossRef>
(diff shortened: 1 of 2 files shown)


swinslow:
The CI build failed, but it looks like the license XML file + test file were fine. 
Looks like the failure was due to something with the deployment code: see https://travis-ci.org/spdx/license-list-XML/jobs/437154527

@goneall Do you have any thoughts what's going on here?

jlovejoy:
@wking can you have a look at why this is failing?

swinslow:
@wking FYI, I am getting the same Error 64 when I run `make validate-canonical-match` on the mainline master branch.

goneall:
I found the issue - some of the warnings were not being ignored.  We need to add more ignored warnings to the list of ignored warning files.  I'll do a pull request later today which should address this.

goneall:
Should be good to merge - build failures are fixed in the master branch.

swinslow:
Hi Gary, do you know if there's a way for me to re-trigger the build check? It's still showing as failed here.

goneall:
@swinslow It will still fail since this branch does not have the updated expected-warnings from the fix.  I would suggest going ahead and merging this with master.  The merge will trigger a new build which should work.

PR #708, opened 2018-10-10
New License/Exception Request: OGL-UK-2.0

jlovejoy:
xml file for OGL-UK-2.0 to resolve issue # 547 (submitted via XML editor)

### DIFF
--- a/src/OGL-UK-2.0.xml
+++ b/src/OGL-UK-2.0.xml
@@ -1,0 +1,65 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<SPDXLicenseCollection xmlns="http://www.spdx.org/license">
+    <license isOsiApproved="false" licenseId="OGL-UK-1.0" name="Open Government License (UK) v1.0" listVersionAdded="3.3">
+        <crossRefs>
+            <crossRef>http://www.nationalarchives.gov.uk/doc/open-government-licence/version/1/</crossRef>
+        </crossRefs>
+        <text>
+            <titleText>Open Government Licence v2.0</titleText>
+            <p>You are encouraged to use and re-use the Information that is available under this licence, the Open Government Licence, freely and flexibly, with only a few conditions.</p>
+            <p>Using Information under this licence</p>
+            <p>Use of copyright and database right material expressly made available under this licence (the 'Information') indicates your acceptance of the terms and conditions below.</p>
+            <p>The Licensor grants you a worldwide, royalty-free, perpetual, non-exclusive licence to use the Information subject to the conditions below.</p>
+            <p>This licence does not affect your freedom under fair dealing or fair use or any other copyright or database right exceptions and limitations.</p>
+            <p>You are free to:</p>
+            <list>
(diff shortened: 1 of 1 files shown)


test/simpleTestForGenerator/OGL-UK-1.0.txt
```diff
@@ -0,0 +1,69 @@
...
+
+You are free to:
+		copy, publish, distribute and transmit the Information;
+		adapt the Information;
+		exploit the Information commercially for example, by combining it with other Information, or by including it in your own product or application.
+
+You must, where you do any of the above:
+		acknowledge the source of the Information by including any attribution statement specified by the Information Provider(s) and, where possible, provide a link to this licence;
+		 If the Information Provider does not provide a specific attribution statement, or if you are using Information from several Information Providers and multiple attributions are not practical in your product or application, you may consider using the following:
 Contains public sector information licensed under the Open Government Licence v1.0.
```
swinslow: I think the test is failing because this text actually differs from the XML markup in this line. Note that before the colon, it says "you may **consider using** the following". The XML markup says "you may **use** the following".

I don't know which one would be "correct" according to the license steward... Perhaps this should have an either/or choice markup in the license?

src/OGL-UK-1.0.xml
```diff
@@ -0,0 +1,65 @@
...
+                <item>ensure that you do not use the Information in a way that suggests any official status or that the Information Provider endorses you or your use of the Information;</item>
+                <item>ensure that you do not mislead others or misrepresent the Information or its source;</item>
+                <item>ensure that your use of the Information does not breach the Data Protection Act 1998 or the Privacy and Electronic Communications (EC Directive) Regulations 2003.</item>
+            </list>
+            <p>These are important conditions of this licence and if you fail to comply with them the rights granted to you under this licence, or any similar licence granted by the Licensor, will end automatically.</p>
+            <p>Exemptions</p>
+            <p>This licence does not cover the use of:</p>
+            <list>
+                <item>personal data in the Information;</item>
+                <item>information that has neither been published nor disclosed under information access legislation (including the Freedom of Information Acts for the UK and Scotland) by or with the consent of the Information Provider;</item>
```
swinslow: Argh, I think the test is failing now b/c of different capitalization -- "information" vs. "Information" for the first word...

swinslow: ...which is odd, I thought the matching guidelines said to ignore capitalization. I could be wrong about this one re: why the test is failing.

jlovejoy: we'll see... changed it. it is a defined term here...

swinslow:
@jlovejoy I think I've got it working now. It looks like the bullets have to be actually declared within the item element. So:

```
<item>personal data in the Information;</item>
```

gets changed to:

```
<item><bullet>-</bullet> personal data in the Information;</item>
```

(and similar for the other items / bullets.) I'm testing this out now to confirm. We can chat on the legal call momentarily.

Issue #683, opened 2018-08-09
Add top level license to license list code and files


kestewart:
@jlovejoy I've gotten a question about the top level license for this github repo (for the aggregation), 
and given we have the output listed in the specification https://spdx.github.io/spdx-spec/appendix-I-SPDX-license-list/  which is CC-BY-3.0  and have had since the start,  I believe that CC-BY-3.0 is a reasonable license to denote this repository with.    Thoughts?

jlovejoy:
began to discuss on Aug 23 call - needs more thought and discussion. Should we take similar approach to data and any potential copyrightable text in SPDX docs as per section 2.2 of spec?

jlovejoy:
(discussed on Sept 20 call)
by way of example:
- spec license is CC-BY-3.0
- SPDX tools are Apache-2.0
- SPDX meta-data - CC0-1.0 (if at all)

for license-list repo - different types of files:
- XML schema has CC-BY-4.0 notice -- should we make this software license? would be easier to make this CC0-1.0. need to get okay from SPDX Legal team, particularly contributors to that file
- validate-schema.js has MIT -- also make this CC0-1.0 to be consistent and for same reasons as above. Check with Trevor that no other code was pulled into this file and we can change license
- XML license files - not going to put license notice in XML files, but a notice in README would helpful to clarify that to the extent there is any copyrightable material... it's CC0-1.0
action items: draft something for review (@jlovejoy) also work on explanation of "license of the license" in parallel (@swinslow @pmadick @bradleeedmondson )

goneall:
As one of the contributors to the XML schema file, CC0-1.0 is fine by me.  @wking is the other contributor to the XML schema file.

wking:
> As one of the contributors to the XML schema file, CC0-1.0 is fine by me.

Me too.  Somebody file a PR with a header tag! ;)

jlovejoy:
thanks @goneall and @wking  - I'll sort out wrapping this up later/over next few days. As per discussion on call, also want to put out an email to the list, so everyone is aware generally.

jlovejoy:
PR #710 created for this

swinslow:
I believe this can be closed now? At least w/r/t anything that will be part of the 3.3 release.

swinslow:
Believe this one has been done since Oct. 2018, so I'm going to go ahead and close it.

PR #710, opened 2018-10-12
update license to CC0-1.0

jlovejoy:
update license to CC0-1.0 as per discussion on Sept 20th call and ok'd by contributors

### DIFF
--- a/schema/ListedLicense.xsd
+++ b/schema/ListedLicense.xsd
@@ -7,7 +7,7 @@
 				This format is internal to the SPDX legal team and is subject to change.  For stable output formats, see <xhtml:a href="https://github.com/spdx/license-list-data">license-list-data</xhtml:a>.
 			</xhtml:p>
 			<xhtml:p>
-				SPDX-License-Identifier: CC0-1.0.
+				SPDX-License-Identifier: CC0-1.0
 			</xhtml:p>
 		</documentation>
 	</annotation>
--- a/schema/ListedLicense.xsd
+++ b/schema/ListedLicense.xsd
@@ -7,7 +7,7 @@
 				This format is internal to the SPDX legal team and is subject to change.  For stable output formats, see <xhtml:a href="https://github.com/spdx/license-list-data">license-list-data</xhtml:a>.
 			</xhtml:p>
 			<xhtml:p>
-				Licensed under CC-BY-4.0.
+				Licensed under CC0-1.0.
 			</xhtml:p>
 		</documentation>
 	</annotation>
--- a/schema/ListedLicense.xsd
+++ b/schema/ListedLicense.xsd
@@ -7,7 +7,7 @@
 				This format is internal to the SPDX legal team and is subject to change.  For stable output formats, see <xhtml:a href="https://github.com/spdx/license-list-data">license-list-data</xhtml:a>.
 			</xhtml:p>
 			<xhtml:p>
-				Licensed under CC0-1.0.
+				SPDX-License-Identifier: CC0-1.0.
 			</xhtml:p>
 		</documentation>
 	</annotation>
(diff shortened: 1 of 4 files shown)


jlovejoy:
this is the PR related to #683 
asked for review by copyright holders / contributors to these two file: @goneall  and @wking 
also posted on mailing list for review by Oct 18th before merging for 3.3 release

jlovejoy:
for the xsd file - is there any reason not to use "SPDX-License-Identifier: CC0-1.0" there as well for the notation??

swinslow:
Hi @jlovejoy -- with these updates, I'd also suggest updating line 33 in /package.json to change it from:

`"license": "CC-BY-3.0 AND MIT"`

to:

`"license": "CC-BY-3.0 AND CC0-1.0"`

since I think validate-schema.js is the only file under MIT, and it's changing to CC0-1.0 in this PR. 

I think keeping CC-BY-3.0 in the package-level metadata makes sense, since it's referring to the spec's own license.

wking:
+1 to updating `package.json` in this PR, so it's clear what's feeding that value.  @goneall, do you remember why you included CC-BY-3.0 in 378fe01b48e5?  My current guess is that it was supposed to match `ListedLicense.xsd`'s CC-BY-4.0, in which case this PR should update it to:

```json
"license": "CC0-1.0"
```

> I think keeping CC-BY-3.0 in the package-level metadata makes sense, since it's referring to the spec's own license.

But [the spec][1] isn't part of this repository, so we don't need to worry about its license here.

[1]: https://github.com/spdx/spdx-spec

goneall:
> do you remember why you included CC-BY-3.0 in 378fe01?

I don't recall, but changing it to match the current XSD license makes sense to me since it is part of this repo.

jlovejoy:
thanks @swinslow  @wking  and @goneall - I think we want consistency here with CC0-1.0 and to avoid conjunctive (or disjunctive) licensing unless there is work included under a pre-existing license, which I don't think there is here.
I'll do another pass and add any other files to this PR along those lines to be thorough.

swinslow:
Agreed -- thanks all, given the above I agree with just listing CC0-1.0 as the license in the metadata.

jlovejoy:
done. that leaves the only non-xml files in the repo to not have explicit license info to be:

- https://github.com/spdx/license-list-XML/blob/master/Makefile

- https://github.com/spdx/license-list-XML/blob/master/expected-warnings

- goneall.gpg
(plus CONTRIBUTING and README, of course)

I think we are good to merge, then?

wking:
Why don't we do the `Makefile` too?  @goneall and I are the only current contributors there, so we can probably drop it into the new `CC0-1.0` bucket.

schema/ListedLicense.xsd
```diff
@@ -7,7 +7,7 @@
 				This format is internal to the SPDX legal team and is subject to change.  For stable output formats, see <xhtml:a href="https://github.com/spdx/license-list-data">license-list-data</xhtml:a>.
 			</xhtml:p>
 			<xhtml:p>
-				Licensed under CC-BY-4.0.
+				SPDX-License-Identifier: CC0-1.0.
```
wking: The trailing period here is not part of a valid license expression.  Can we drop it?

jlovejoy: fixed, good catch!

wking:
> ... where should I put "SPDX-License-Identifier: CC0-1.0"

As the first line?  Do we want to include a Copyright declaration up there too, or are we comfortable leaving that implicit in the version-control history?

> ... and is there specific comment notation that should be used here?

`Makefile`s use [`#` comments][1], so just:

```Makefile
# SPDX-License-Identifier: CC0-1.0
```

[1]: http://pubs.opengroup.org/onlinepubs/9699919799/utilities/make.html#tag_20_76_13_01

PR #712, opened 2018-10-12
New License/Exception Request: OGL-UK-3.0

jlovejoy:
new license, will need text file to be added once PR is created

### DIFF
--- a/src/OGL-UK-3.0.xml
+++ b/src/OGL-UK-3.0.xml
@@ -1,0 +1,65 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<SPDXLicenseCollection xmlns="http://www.spdx.org/license">
+    <license isOsiApproved="false" licenseId="OGL-UK-3.0" name="Open Government Licence v3.0" listVersionAdded="3.3">
+        <crossRefs>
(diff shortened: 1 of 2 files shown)


jlovejoy:
@swinslow - I incorporate your suggested improvements from the old PR and then re-uploaded to a new branch and added text file here. can you review?

src/OGL-UK-3.0.xml
```diff
@@ -0,0 +1,65 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<SPDXLicenseCollection xmlns="http://www.spdx.org/license">
+    <license isOsiApproved="false" licenseId="OGL-UK-3.0" name="Open Government Licence v3.0" listVersionAdded="3.3">
```
swinslow: My only question is whether the *name* attribute should be "Open Government Licence v3.0" or "Open Government Licence **(UK)** 3.0", to match the UK in the licenseId. I saw your email to Howard and the SPDX-legal list asking about this.

swinslow: I'm good with this as-is; I expect it's something we can revise in the future if the preference is to insert "(UK)" later.

jlovejoy: yes, we'll wait to hear back from Howard and adjust accordingly - hopefully by Thursday :)

PR #713, opened 2018-10-18
Replacement for #708: New License/Exception Request: OGL-UK-1.0

swinslow:
Submitting as a new PR (because I'm not sure how to merge into #708) -- this is intended to fix the test failures discussed in that thread.


swinslow:
This looks like it failed not because of the OGL-UK-1.0 license, but because of an error in the FreeRTOS exception notes section that was just merged. I'll check and see if I can re-submit to fix.

jlovejoy:
oh no!!

swinslow:
Okay - looks like it's working now!  =)  This should be good to merge, and to close #708 after that.

Issue #714, opened 2018-10-18
New license request: GPLCC-1.0 [SPDX-Online-Tools]


goneall:
**1.** License Name: GPL Cooperation Commitment 1.0
**2.** Short identifier: GPLCC-1.0
**3.** URL: https://github.com/gplcc/gplcc/blob/eee52346dd48bdf985657f082bb847f12f40c464/Project/COMMITMENT
**4.** OSI Approval: no

jlovejoy:
for further info, see original email from @richardfontana  at https://lists.spdx.org/g/Spdx-legal/message/2410

bkuhn:
Just a reminder that this will need to be marked for a later release, given that this issue is intimately tied #655 which is now marked for "Later Release".  I'd suggest keeping this Issue and #655 in the same state in tandem, since obviously if one of them is added to the SPDX license list, the other one should be, for parity since they are so similar, as @jlovejoy [pointed out in the other issue](https://github.com/spdx/license-list-XML/issues/655#issuecomment-397696797).

jlovejoy:
That is not an accurate assessment of where we ended up -  most of the discussion was on the mailing list and so that comment is extremely out of date. Where we got to on #655 was: given that the people who are actually working in implementing SPDX identifiers in the kernel were not asking for or wanting it and there seemed to be a lot of controversy, no change to the current status is the best way to proceed. 

As it turned out, combining the discussion with the GPL Cooperation Commitment was not helpful and I believe I expressed at some point that my idea of combining the discussion was obviously not a good one. Please don't quote an out-dated statement now, as if it's current.

jlovejoy:
updated discussion posted on mailing list at: https://lists.spdx.org/g/Spdx-legal/message/2571

jlovejoy:
#733 merged

PR #733, opened 2018-12-06
added new exception GPL-CC-1.0.xml

zvr:
As per #714, an exception to track the GPL Cooperation Commitment.

A single file with `optional` sections and `alt` matches to handle all variants: Individual, Company, and Project.

**Note**: The matching is simple, accepting for example the originals
> I commit

and
> COMPANY commits

The simplicity means that it also accepts the grammatically incorrect
> I commits

If we really want to make sure this is not OK, the matches have to be extended from single words to phrases.

### DIFF
--- a/src/exceptions/GPL-CC-1.0.xml
+++ b/src/exceptions/GPL-CC-1.0.xml
@@ -2,11 +2,11 @@
 <SPDXLicenseCollection xmlns="http://www.spdx.org/license">
   <exception licenseId="GPL-CC-1.0" name="GPL Cooperation Commitment 1.0">
     <crossRefs>
-      <crossRef>https://gplcc.github.io/</crossRef>
+      <crossRef>https://github.com/gplcc/gplcc/blob/master/Project/COMMITMENT</crossRef>
+	    <crossRef>https://gplcc.github.io/gplcc/Project/README-PROJECT.html</crossRef> 	
     </crossRefs>
     <notes>
-      Single exception covering three variants:
-      Individual, Company, and Project.
+      This is the GPL Cooperation Commitment for projects. It is distinct from the GPL Cooperation Commitment for companies or individuals in that it applies at the project level for all contributions going forward as of the date it is adopted. 
     </notes>
     <text>
       <titleText>
--- a/src/exceptions/GPL-CC-1.0.xml
+++ b/src/exceptions/GPL-CC-1.0.xml
@@ -8,94 +8,96 @@
       Single exception covering three variants:
       Individual, Company, and Project.
     </notes>
-    <titleText>
+    <text>
+      <titleText>
+	<optional>
+	<p>
+	  GPL Cooperation Commitment Version 1.0
+	</p>
+	</optional>
+      </titleText>
+      <p>
+	<optional>Solely for any software for which I personally own
+	copyright that is licensed under a Covered License,</optional>
+	<alt name="before" match="before|Before">before</alt>
+	filing or continuing to prosecute any legal proceeding or claim
+	(other than a Defensive Action) arising from termination of a
+	Covered License,<alt name="subject" match="I|we|.*">we</alt>
+	commit<optional>s</optional>
+	to extend to the person or entity ("you") accused of violating
+	the Covered License the following provisions regarding cure and
+	reinstatement, taken from GPL version 3. As used here, the term
+	"this License" refers to the specific Covered License being enforced.
+      </p>
+      <p>
+	<optional>"</optional>However,
+	if you cease all violation of this License, then your
+	license from a particular copyright holder is reinstated (a)
+	provisionally, unless and until the copyright holder explicitly
+	and finally terminates your license, and (b) permanently, if
+	the copyright holder fails to notify you of the violation by
+	some reasonable means prior to 60 days after the cessation.
+      </p>
+      <p>
+	Moreover, your license from a particular copyright holder is
+	reinstated permanently if the copyright holder notifies you of
+	the violation by some reasonable means, this is the first time you
+	have received notice of violation of this License (for any work)
+	from that copyright holder, and you cure the violation prior to
+	30 days after your receipt of the notice.<optional>"</optional>
+      </p>
+      <p>
+	<alt name="subject" match="I|We|.*">We</alt>
+	intend<optional>s</optional>
+	this Commitment to be irrevocable, and binding and enforceable
+	against <alt name="object" match="me|us|.*">us</alt>
+	and assignees of or successors to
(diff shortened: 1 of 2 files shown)


src/exceptions/GPL-CC-1.0.xml
```diff
@@ -0,0 +1,101 @@
...
+<SPDXLicenseCollection xmlns="http://www.spdx.org/license">
+  <exception licenseId="GPL-CC-1.0" name="GPL Cooperation Commitment 1.0">
+    <crossRefs>
+      <crossRef>https://gplcc.github.io/</crossRef>
+    </crossRefs>
+    <notes>
+      Single exception covering three variants:
+      Individual, Company, and Project.
+    </notes>
+    <titleText>
```
goneall: Need to add `<text>`

src/exceptions/GPL-CC-1.0.xml
```diff
@@ -0,0 +1,101 @@
...
+      "We" means each contributor to this repository as of the date of
+      inclusion of this file, including subsidiaries of a corporate contributor.
+    </p>
+    </optional>
+    <optional>
+    <p>
+      This work is available under a Creative Commons Attribution-ShareAlike 4.0
+      International license (https://creativecommons.org/licenses/by-sa/4.0/).
+    </p>
+    </optional>
```
goneall: Need to add `</text>`

src/exceptions/GPL-CC-1.0.xml
```diff
@@ -0,0 +1,101 @@
...
+    <p>
+      Moreover, your license from a particular copyright holder is
+      reinstated permanently if the copyright holder notifies you of
+      the violation by some reasonable means, this is the first time you
+      have received notice of violation of this License (for any work)
+      from that copyright holder, and you cure the violation prior to
+      30 days after your receipt of the notice.<optional>"</optional>
+    </p>
+    <p>
+      <alt name="subject" match="I|We|.*">We</alt>
```
goneall: Since .* will match any of the words, the other alternatives will be ignored.  I'm thinking we just restrict it to I and We and remove the .*

src/exceptions/GPL-CC-1.0.xml
```diff
@@ -0,0 +1,101 @@
...
+      intend<optional>s</optional>
+      this Commitment to be irrevocable, and binding and enforceable
+      against <alt name="object" match="me|us|.*">us</alt>
```
goneall: Same comment as above - Since .* will match any of the words, the other alternatives will be ignored.  I'm thinking we just restrict it to I and We and remove the .*

src/exceptions/GPL-CC-1.0.xml
```diff
@@ -0,0 +1,101 @@
...
+      and assignees of or successors to
+      <alt name="possessive" match="my|our|.*">our</alt>
```
goneall: Same comment as above -  Since .* will match any of the words, the other alternatives will be ignored.  I'm thinking we just restrict it to I and We and remove the .*

src/exceptions/GPL-CC-1.0.xml
```diff
@@ -0,0 +1,101 @@
...
+    </p>
+    <p>
+      "Covered License" means the GNU General Public License, version
+      2 (GPLv2), the GNU Lesser General Public License, version 2.1
+      (LGPLv2.1), or the GNU Library General Public License, version
+      2 (LGPLv2), all as published by the Free Software Foundation.
+    </p>
+    <p>
+      "Defensive Action" means a legal proceeding or claim
+      that<alt name="subject" match="I|We|.*">We</alt>
```
goneall: Same comment as above - Since .* will match any of the words, the other alternatives will be ignored.  I'm thinking we just restrict it to I and We and remove the .*

goneall:
I lean toward individual words/phrases.  If we want to allow any expression, I would suggest changing the match to just `.*` since the alternate words would will also be matched with the `.*`.

src/exceptions/GPL-CC-1.0.xml
zvr: I know that ".*" matches anything, but I wanted to make it explicit.
We need the `.*` because in the Company variant, this gets the COMPANY_NAME.

src/exceptions/GPL-CC-1.0.xml
zvr: same reply. need COMPANY_NAME

src/exceptions/GPL-CC-1.0.xml
zvr: same reply. need COMPANY_NAME

src/exceptions/GPL-CC-1.0.xml
zvr: same reply. need COMPANY_NAME

zvr:
if we want to forbid the grammatically incorrect, we would have to have something like:
`<alt name="someone_commits" match="(I|we) commit|.* commits">I commit</alt>`

src/exceptions/GPL-CC-1.0.xml
goneall: That makes sense - I agree with the approach.

zvr:
@goneall what can be done with tests?  I see that the usual way is to have the complete text as a file with the same name under tests.
Is there a way of having all three variants?

goneall:
@zvr Right now, there is only one test allowed per license/exception.  The tool has 2 license test implementations.  There is a test implementation that allows for multiple tests based on a naming convention (see https://github.com/spdx/LicenseListPublisher/blob/master/src/org/spdx/licenselistpublisher/licensegenerator/LicenseTester.java), but it isn't in use since we did not get consensus on a naming convention for the tests.  See https://github.com/spdx/license-test-files/issues

jlovejoy:
thanks for all the work on this, @zvr  and @goneall  - we are going to hold off to next release and have a bit more of a think about whether we should try to incorporate all 3 variants or just the project variant

jlovejoy:
pending discussion/feedback based on https://lists.spdx.org/g/Spdx-legal/message/2571 - we will need to update this to reflect just the text of the project commitment. I think that will simply mean straight text with no optional/alt text. Will make appropriate changes once we've had more time on mailing list for comments

jlovejoy:
updated to reflect project variant only. This resulted in no alternative or optional text, although I left the notice about the license of the license at the bottom as optional. @richardfontana  can you have a look/check?

richardfontana:
> updated to reflect project variant only. This resulted in no alternative or optional text, although I left the notice about the license of the license at the bottom as optional. @richardfontana can you have a look/check?

@jlovejoy looks good to me

PR #754, opened 2019-01-16
crossRefs: HTTP -> HTTPS: gnu.org, creativecommons.org, opensource.org

yegortimoshenko:
<http://gnu.org>, <http://creativecommons.org> and <http://opensource.org> all redirect to their respective HTTPS counterparts. This pull request updates these URLs (only in `crossRefs`, never in license text) to redirect targets.


jlovejoy:
this seems like a fine update to me. @goneall agree?

goneall:
@jlovejoy Agree
@yegortimoshenko Thanks for catching this and submitting the PR

Issue #751, opened 2019-01-12
libpng license: review new version, add markup


jlovejoy:
Latest release of libpng, 1.6.36, updated its license: http://www.libpng.org/pub/png/src/libpng-LICENSE.txt . To cite the homepage at http://www.libpng.org/pub/png/libpng.html : "The current public release, libpng 1.6.36, . . . updates the license (identical terms to the zlib license, with the old license appended in the manner of the Python Software Foundation License version 2, and the list of contributing authors moved to a separate AUTHORS file)."

The update license file looks to be two licenses:
- the first part, "PNG Reference Library License version 2" has similar aspects as zlib, but with a longer lead-in and disclaimer in the first part
- the second part, "PNG Reference Library License version 1 (for libpng 0.5 through 1.6.35)" is similar in substance to  Libpng currently on SPDX License List, but would need review and markup to be a match. This needs further review in terms of matching to ensure differences are not legally substantive (my initial impression is that they are not)

Submitter's suggestion is to add first license and update existing Libpng (at which point, someone could use the AND operator to combine the two for identifying the license of this library.

Could someone please pick this one up and prepare a specific proposal and analysis?

jlovejoy:
see initial email here: https://lists.spdx.org/g/Spdx-legal/topic/28914731#2520

bradleeedmondson:
Discussed on legal call 2/7/2019. We view this as a new license and will plan to add it as libpng-2.0. I can take this issue and template it in a PR.

bradleeedmondson:
Oops colliding with @zvr. Mind if I keep this one?

zvr:
Sure, no problem.

bradleeedmondson:
I've added the v2 license in #775 

Can I get a review that this looks correct? In particular we should be sure that we mean to add only the new text, not the ([new text]) + ([new header text for old license] + [old license]) as the v2. Does #775 look right to @jlovejoy @goneall ? Thanks!

swinslow:
#775 was merged, so I'm closing this issue. Thanks @bradleeedmondson!

PR #775, opened 2019-02-21
Add Libpng-2.0

bradleeedmondson:
Add Libpng-2.0 per #751


bradleeedmondson:
Build failed bc what I called "title" in the alt tag should be "name".

src/Libpng-2.0.xml
```diff
@@ -0,0 +1,49 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<SPDXLicenseCollection xmlns="http://www.spdx.org/license">
+   <license isOsiApproved="false" licenseId="Libpng-2.0" name="libpng License 2.0">
```
swinslow: The `license` tag should probably also include a `listVersionAdded="3.5"` value. (@jlovejoy has noted this on some of my PRs, so I'm noting it here) =)

src/Libpng-2.0.xml
```diff
@@ -0,0 +1,49 @@
...
+      <crossRefs>
+         <crossRef>http://www.libpng.org/pub/png/src/libpng-LICENSE.txt</crossRef>
+      </crossRefs>
+
+    <text>
+      <p>PNG Reference Library License version 2</p>
+      <alt match="^[-]*$" name="dashed-line">---------------------------------------</alt>
+      <list>
```
swinslow: Should this list be within a `<copyrightText>` block, so that the copyright notices are tagged as such and are optional?

bradleeedmondson: Good point! I think the title should be in a `<titleText>` block, with the dashed line optional. I think the line is made "optional" by the regex in the match parameter, but should it also be `<optional>` or in `<copyrightText>`?

goneall: I would suggest enclosing the list of copyrights in a `<copyrightText>` block.  I also agree wrapping the title would be a good idea - this would allow the matching algorithm to treat the title as optional.

src/Libpng-2.0.xml
bradleeedmondson: Added #795 to deprecate plain libpng identifier in favor of LibPNG-1.0

swinslow:
Minor nit, is "LibPNG" the project's preferred capitalization? I'm seeing varying capitalizations, but from a (very) quick skim "libpng" seems to be the more common form?

DennisClark:
Many open-source libraries, such as "LibPNG" (libpng, Libpng, LIBPNG, etc.) are very casual and non-enforcing regarding the exact spelling of the name; I don't think they really care.

bradleeedmondson:
Discussed on today's call. Decision to update the name attribute to upstream's "PNG Reference Library version 2" and to move the identifier to lowercase to match upstream's name for the library (as well as libtiff, the next license on the SPDX License List).

jlovejoy:
this looks good to merge to me? agree @bradleeedmondson and @swinslow ?

swinslow:
From the notes in the preceding comment, it looks like we were going to first change the name to align with the upstream source and the identifier to make it lowercase. @bradleeedmondson, can you make that change and then we can merge? Thank you!

bradleeedmondson:
The build check failed, but I can't tell whether that's a problem with my PR or with the build environment. This is toward the end of the [1]build log:

> gpg --verify --no-default-keyring --keyring ./goneall.gpg licenseListPublisher-2.1.17.jar.asc
> gpg: no valid OpenPGP data found.
> gpg: the signature could not be verified.
> Please remember that the signature file (.sig or .asc)
> should be the first file given on the command line.
> make: *** [licenseListPublisher-2.1.17.jar-valid] Error 2

I'm guessing the build environment is using a licenseListPublisher that was mistakenly released as an unsigned jar? But I could be way, way off. @goneall can you check? Thanks!

[1] https://api.travis-ci.org/v3/job/510724767/log.txt

goneall:
@bradleeedmondson I restarted the travis build and this time it passed.  I have no idea why it failed the first time.  There hasn't been any recent changes in the tools jar files.

PR #778, opened 2019-03-01
Fix #777 Use header-less archive.org URLs

### LINKED ISSUE #777: Use header-less archive.org URLs
E.g. compare

http://web.archive.org/web/20021004124254/http://www.opensource.org/licenses/academic.php

used now with

http://web.archive.org/web/20021004124254id_/http://www.opensource.org/licenses/academic.php

The second variant shows the original content without the archive.org header, which should IMO be preferred.

ppalaga:


ppalaga:
Ideally this change should be accompanied by some sort of a test that ensures that newly added archive.org URLs are header-less too.

swinslow:
given resolution of #777, closing this PR

PR #795, opened 2019-03-05
Deprecate libpng in favor of LibPNG-1.0

bradleeedmondson:
Deprecate "libpng" identifier in favor of LibPNG-1.0
* add LibPNG-1.0.xml, with notes about deprecation, and line-wrap at about 80 chars
* update libpng.xml to deprecated status, with deprecation License List version, and deprecation notes
* add test file LibPNG-1.0.txt, leaving old test file Libpng.txt in place per other deprecations (e.g., #627)


jlovejoy:
what is the compelling reason to deprecate libpng for LibPNG-1.0? 
it appears to be just to tidy up the identifier, which is preferable from a tidy standpoint, but I'm not sure that alone warrants deprecation??

bradleeedmondson:
The only rationale I have is to make sure users can quickly disambiguate v1 and v2. I think it's marginally better going forward, but I'm not wedded to this change and happy to do what the group decides.

swinslow:
I don't have a super-strong opinion either, but I'd be in favor of deprecating here. Partly for the reason @bradleeedmondson mentioned, to help with disambiguating; and partly because on the libpng license page they're now explicitly referring to the old license as "version 1".

DennisClark:
I do not think that libpng needs to be deprecated in favor of LibPNG-1.0, but I don't feel strongly about it. I just think it's fine the way it is.

swinslow:
Discussed on legal team call 2019-03-07, decided to postpone making a call on whether to deprecate libpng now, will discuss again on a subsequent call.

bradleeedmondson:
Postponed decision on this deprecation past 3.5 release. 

Note the libpng-2.0 license was added in 3.5, so the "libpng" and "libpng-2.0" license IDs will be out of sync.

To decide later whether to tidy these up as "libpng-1.0" and "libpng-2.0".

swinslow:
Given that this has been sitting unmerged for 3 years, I'm going to go ahead and close this one for now. Can probably re-open as a new issue (rather than PR) for discussion if there's interest in re-visiting this.

Issue #617, opened 2018-03-13
Add notes explaining "or later" clause in *GPL*-only licenses


hraban:
AGPL-3.0-only.xml contains the same text as AGPL-3.0-or-later.xml:

> This program is free software: you can redistribute it and/or modify it under the terms of the GNU Affero General Public License as published by the Free Software Foundation, either version 3 of the License, or (at your option) any later version.

wking:
> AGPL-3.0-only.xml contains the same text as AGPL-3.0-or-later.xml…

The license text is the same, because it's a single license.  The difference is in the recommended grant wording:

```console
$ git describe
v3.0
$ diff -u src/AGPL-3.0-*
--- src/AGPL-3.0-only.xml       2018-03-13 16:19:53.794074010 -0700
+++ src/AGPL-3.0-or-later.xml   2018-03-13 16:19:53.794074010 -0700
@@ -1,26 +1,23 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <SPDXLicenseCollection xmlns="http://www.spdx.org/license">
-  <license licenseId="AGPL-3.0-only" isOsiApproved="true"
-  name="GNU Affero General Public License v3.0 only">
+  <license licenseId="AGPL-3.0-or-later" isOsiApproved="true"
+  name="GNU Affero General Public License v3.0 or later">
     <crossRefs>
       <crossRef>http://www.gnu.org/licenses/agpl.txt</crossRef>
       <crossRef>http://www.opensource.org/licenses/AGPL-3.0</crossRef>
     </crossRefs>
     <standardLicenseHeader>
       Copyright (C)<alt name="copyright" match=".+">[year] [name of author]</alt>
-      <p>
-       This program is free software: you can redistribute it and/or modify it
-        under the terms of the GNU Affero General Public License as published
-       by the Free Software Foundation, version 3.</p>
-      <p>
-        This program is distributed in the hope that it will be useful,
-       but WITHOUT ANY WARRANTY; without even the implied warranty of
-       MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
-       GNU Affero General Public License for more details.</p>
-      
-      <p>
-        You should have received a copy of the GNU Affero General Public License
-        along with this program. If not, see
+      <p>This program is free software: you can redistribute it and/or modify it
+      under the terms of the GNU Affero General Public License as published
+      by the Free Software Foundation, either version 3 of the License, or 
[...]

PR #797, opened 2019-03-08
add explanatory text re: -only and -or-later

jlovejoy:
add explanatory text re: -only and -or-later delineation in standard license header (versus default text in full copy of license). proposed solution to #617

### DIFF
--- a/src/GPL-2.0-only.xml
+++ b/src/GPL-2.0-only.xml
@@ -31,7 +31,7 @@
       to use the code under GPL-2.0-only, as distinguished from use of code under GPL-2.0-or-later
       (i.e., GPL-2.0 or some later version). The license notice (as seen in the Standard License 
       Header field below)  states which of these applies to the code in the file. The example 
-      in the exhibit to the license shows the license notice for the "or later" approach.</p>
+      in the exhibit to the license shows the license notice for the "or later" approach.
     </notes>
     <text>
     <titleText>
--- a/src/GPL-2.0-only.xml
+++ b/src/GPL-2.0-only.xml
@@ -27,8 +27,11 @@
       </p>
     </standardLicenseHeader>
     <notes>
-      This license was released: June 1991 This refers to when this
-      GPL 2.0 only is being used (as opposed to GPLv2 or later).
+      This license was released: June 1991. This license identifier refers to when 
+      GPL-2.0-only is being used as opposed to GPL-2.0-or-later. The option to use either that
+      version of the license only or to make it available under any later version of that license
+      is denoted in the standard license header. (The standard license header in the license 
+      itself includes the default "or (at your option) any later version" language.)
     </notes>
     <text>
     <titleText>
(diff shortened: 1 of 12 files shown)


jlovejoy:
@tieguy @theopenchainproject - want to have a look here?

jlovejoy:
if people think this explains thing well enough, then will add to other GNU licenses

jlovejoy:
have reached out to FSF to get their review as well

src/GPL-2.0-only.xml
```diff
@@ -27,8 +27,11 @@
       </p>
     </standardLicenseHeader>
     <notes>
-      This license was released: June 1991 This refers to when this
-      GPL 2.0 only is being used (as opposed to GPLv2 or later).
+      This license was released: June 1991. This license identifier refers to when 
+      GPL-2.0-only is being used as opposed to GPL-2.0-or-later. The option to use either that
```
tieguy: Can I suggest replacing "The option to use... " with "The license header of individual files will indicate which of 'only' or 'later' applies to that particular file." I _think_ that's the same intent, but if not, might want to clarify in a different way?

shanecoughlan:
> @tieguy @theopenchainproject - want to have a look here?

Looks like a good solution to me.

jlovejoy:
thanks @tieguy and @shanecoughlan - I've gotten some feedback from Donald at FSF as well, so just need to nail down that with him and make some edits here... hopefully soon!

swinslow:
Hi @jlovejoy, this looks good to me! Does this incorporate the changes to the notes that you were incorporating based on your discussion with FSF? If so, I'm glad to go ahead and push the merge button.

src/GPL-2.0-only.xml
```diff
@@ -27,11 +27,12 @@
...
-      GPL-2.0-only is being used as opposed to GPL-2.0-or-later. The option to use either that
-      version of the license only or to make it available under any later version of that license
-      is denoted in the standard license header. (The standard license header in the license 
-      itself includes the default "or (at your option) any later version" language.)
+      This license was released: June 1991. This license identifier refers to the choice 
+      to use the code under GPL-2.0-only, as distinguished from use of code under GPL-2.0-or-later
+      (i.e., GPL-2.0 or some later version such as GPL 3, a possible future GPL 4,
+      and so on). The license notice (usually at the head of the file) states which of these
+      applies to the code in the file.
+      <p>The example in the exhibit to the license shows the license notice for the "or later" approach.</p>
```
tieguy: I assume there is no way for FSF to issue/approve an example that does not show the or-later approach?

swinslow: Probably not before we freeze the release for 3.5... which we're hoping to do today / Monday.  =)  But I haven't spoken with them, so maybe @jlovejoy can weigh in here.

jlovejoy:
@tieguy  - note, on the SPDX license pages, we show the correct standard header (in terms of reflecting the only or or-later option) but the standard license text exhibit has the or-later language. This is what was confusing people.

I've now updated the language again to tighten up a bit and call out the above fact more explicitly. I've also added the rest of the licenses this applies to.  I think we can go with this.

Also note: AGPL-1.0 and LGPL-3.0 don't have this issue, as they don't include the license notice exhibit

PR #802, opened 2019-03-08
Add CERN-OHL-1.1

jlovejoy:


jlovejoy:
@swinslow indeed, you are right - fixed!

PR #824, opened 2019-03-21
Add three new templated OpenSSL exceptions

### LINKED ISSUE #720: New Exception Request: OpenSSL Exception
Full name: OpenSSL Exception
Short identifier: OpenSSL-exception
Existing discussion: https://lists.spdx.org/g/Spdx-legal/topic/24235315#2398

swinslow:
This commit addresses issue #720 by adding templated versions of
three different variants of OpenSSL exceptions found in the wild.
It allows these variants to be used with different licensors and
with different underlying licenses. It does not modify existing
openvpn-openssl-exception, which is specific to OpenVPN and GPLv2.

Fixes #720

### DIFF
--- a/src/exceptions/debian-openssl-exception.xml
+++ b/src/exceptions/debian-openssl-exception.xml
@@ -1,0 +1,22 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<SPDXLicenseCollection xmlns="http://www.spdx.org/license">
+   <exception licenseId="debian-openssl-exception" name="Debian OpenSSL Exception" listVersionAdded="3.5">
+      <crossRefs>
+         <crossRef>https://people.gnome.org/~markmc/openssl-and-the-gpl.html</crossRef>
+      </crossRefs>
+    <text>
+      <p>In addition, as a special exception, the copyright holders give
+         permission to link the code of portions of this program with the
+         OpenSSL library under certain conditions as described in each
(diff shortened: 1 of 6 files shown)


jlovejoy:
just read thread in #720 - I'm a little lost in terms of if this is ready to merge for 3.5
is the naming vetted for these? Also seems like using some capitalization would be helpful and consistent perhaps?
e.g. MongoDB-OpenSSL-exception
(although doesn't look like we did that for the existing OpenSSL exception, sigh)

@swinslow as you and @silverhook have been following this, I'll let you merge when it's ready or say otherwise if you need more input

swinslow:
@jlovejoy Prior legal team discussions had suggested things were aligned on these four exceptions, bu given the comments in #720 it looks like discussion is continuing which may make some parts of this unnecessary. So this may not yet be ready for 3.5. I suggest we move it to 3.6 and I'll aim to close it out early in the release cycle.

jlovejoy:
@swinslow - agree re: discussion continuing and to give that some more time

swinslow:
Per separate conversations on prior legal team call, this one needs some more work. I'm moving it from 3.6 to 3.7 and removing the "Accepted" tag, given the subsequent discussion in the issues thread.

swinslow:
Closing this, in favor of the analysis that will be happening in #1356 -- we can pull bits of this PR back out if it turns out to be useful.

Issue #800, opened 2019-03-08
New license request: BlueOak-1.0.0 [SPDX-Online-Tools]


kemitchell:
**1.** License Name: Blue Oak Model License 1.0.0
**2.** Short identifier: BlueOak-1.0.0
**3.** URL: https://blueoakcouncil.org/license/1.0.0
**4.** OSI Status: Not Submitted

jlovejoy:
(from email Kyle sent to list)
Aside: All other licenses on our published permissive license list,
https://blueoakcouncil.org/list, are currently specified by SPDX ID.

Explanation:

The Blue Oak Model License 1.0.0 advances the state of the art of
permissive public software licensing by unique combination of old and
new techniques. To name but a few highlights, Blue Oak:
- uses plain English, not legalese, written by a committee of open
 licensing lawyer-specialists
- explicitly addresses both copyright and patent
- removes uncertainty about the legal characterization of its terms
- covers all relevant contributor patents, without scope limitation or
 defensive termination
- includes a cure provision for attribution violations
- explicitly states that contributors cannot revoke their grants
- expresses the expectation that all contributors will license on the
 same terms

Both the Blue Oak Model License and its steward, Blue Oak Council, were
announced this week.  We expect adoption in meaningful projects, now
that the license has been made public.  If need by, I will be happy to
follow up with links to a couple examples, as they appear.

For more information on the license's features and history, see:

https://blueoakcouncil.org/2019/03/06/model.html

DennisClark:
Looks good to me. A few comments: 

1. It's always nice to see a Version Number on a new license request, which gives one confidence that the license text is stable. 
2. It's also very nice to see warranty disclaimer text in mixed case rather than all upper case; I hope that is an ongoing trend.
3. It's remarkably easy to read and understand.  

So even though it's apparently not yet an OSI license, and at this point adoption is still pending in actual projects, I give it an early approval.

kemitchell:
@DennisClark thank you!

> It's always nice to see a Version Number on a new license request, which gives one confidence that the license text is stable.

The license was both revised and vetted collaboratively, by a team of experts, and not just once. We are actively pushing for adoption of 1.0.0, as submitted.

That being said, we are very open to new conversations about design and implementation. We versioned from the get-go precisely out of respect for the value of open collaboration, and did our initial work in private only to ensure we'd offer something worthy of close consideration.

[...]

PR #829, opened 2019-03-22
Blue Oak Model License 1.0.0

kemitchell:
This draft PR for #800 was another chance for me to do an XML file.

The interesting part of this one was the Markdown syntax.  I'm not aware of another license on the list that uses that kind of formatting. Fortunately, Blue Oak doesn't do anything fancy, and it was straightforward to mark all of the Markdown syntax characters optional.

As an aside, we _do_ recommend copying the Markdown plain text from blueoakcouncil.org verbatim, even into plain-text files without Markdown extensions.

### DIFF
--- a/src/BlueOak-1.0.0.xml
+++ b/src/BlueOak-1.0.0.xml
@@ -7,25 +7,25 @@
     <notes>This license was released on March 6, 2019.</notes>
     <text>
       <titleText>
-        <p># Blue Oak Model License</p>
+        <p><optional># </optional>Blue Oak Model License</p>
       </titleText>
       <p>Version 1.0.0</p>
-      <p>## Purpose</p>
+      <p><optional>## </optional>Purpose</p>
       <p>This license gives everyone as much permission to work with this software as possible, while protecting contributors from liability.</p>
-      <p>## Acceptance</p>
-      <p>In order to receive this license, you must agree to its rules.  The rules of this license are both obligations under that agreement and conditions to your license. You must not do anything with this software that triggers a rule that you cannot or will not follow.</p>
-      <p>## Copyright</p>
+      <p><optional>## </optional>Acceptance</p>
+      <p>In order to receive this license, you must agree to its rules. The rules of this license are both obligations under that agreement and conditions to your license. You must not do anything with this software that triggers a rule that you cannot or will not follow.</p>
+      <p><optional>## </optional>Copyright</p>
       <p>Each contributor licenses you to do everything with this software that would otherwise infringe that contributor's copyright in it.</p>
-      <p>## Notices</p>
-      <p>You must ensure that everyone who gets a copy of any part of this software from you, with or without changes, also gets the text of this license or a link to &lt;https://blueoakcouncil.org/license/1.0.0&gt;.</p>
-      <p>## Excuse</p>
-      <p>If anyone notifies you in writing that you have not complied with [Notices](#notices), you can keep your license by taking all practical steps to comply within 30 days after the notice.  If you do not do so, your license ends immediately.</p>
-      <p>## Patent</p>
+      <p><optional>## </optional>Notices</p>
+      <p>You must ensure that everyone who gets a copy of any part of this software from you, with or without changes, also gets the text of this license or a link to <optional>&lt;</optional>https://blueoakcouncil.org/license/1.0.0<optional>&gt;</optional>.</p>
+      <p><optional>## </optional>Excuse</p>
+      <p>If anyone notifies you in writing that you have not complied with <alt match="Notices">[Notices](#notices)</alt>, you can keep your license by taking all practical steps to comply within 30 days after the notice. If you do not do so, your license ends immediately.</p>
+      <p><optional>## </optional>Patent</p>
       <p>Each contributor licenses you to do everything with this software that would otherwise infringe any patent claims they can license or become able to license.</p>
-      <p>## Reliability</p>
+      <p><optional>## </optional>Reliability</p>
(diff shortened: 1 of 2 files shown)


swinslow:
This looks good to me! I haven't pulled it to run the tests, and it's still listed as draft so I don't think the CI has run. But from a basic look it seems fine.

kemitchell:
Rebased.

swinslow:
Hi @kemitchell -- hope you don't mind me adding a commit to the branch here (based on your comments on JPNIC, I'm assuming you wouldn't!)

Based on the Travis CI output, I think this was failing because there was a missing `name=` attribute in the `alt` tag for the Notices reference. I've inserted it and expecting (hoping) it will pass the test now. If not, I'll take a closer look.

I'm expecting to discuss BlueOak-1.0.0 on the Legal Team call this Thursday, 2019-04-18, to look for the go-ahead from the team. So if this PR is ready at that point, and the Legal Team gives the thumbs-up, then we can go ahead and merge it.

swinslow:
OK, although that isn't throwing the schema error now, unfortunately it's now showing a mismatch between the XML file and the test license text file. Instead of me pushing a bunch more commits to see what happens, let me check it out offline and see if I can get the tests passing...

kemitchell:
Thanks, @swinslow.

kemitchell:
@swinslow, for the record here, I don't currently plan to attend to the legal team call this week. I understand that's not necessary.

If minds change, let me know. I'll be happy to dial in.

swinslow:
No worries, thanks @kemitchell and understood. If there are questions for you that come up from the call, we can circle back with you.

src/BlueOak-1.0.0.xml
```diff
@@ -17,15 +17,15 @@
       <p><optional>## </optional>Copyright</p>
       <p>Each contributor licenses you to do everything with this software that would otherwise infringe that contributor's copyright in it.</p>
       <p><optional>## </optional>Notices</p>
-      <p>You must ensure that everyone who gets a copy of any part of this software from you, with or without changes, also gets the text of this license or a link to <optional>&lt;</optional>https://blueoakcouncil.org/license/1.0.0<optional>&gt;</optional>.</p>
+      <p>You must ensure that everyone who gets a copy of any part of this software from you, with or without changes, also gets the text of this license or a link to &lt;https://blueoakcouncil.org/license/1.0.0&gt;.</p>
```
kemitchell: The angle brackets here are nonsemantic. Blue Oak is still Blue Oak without them. Moreover, many Markdown renders will autodetect the URI and render as a hyperlink either way.

swinslow: Agreed -- I think that's fine, and I think omitting the `optional` tags here doesn't change that. My understanding of the SPDX matching guidelines is that angle brackets would be interpreted as optional in any event, as I don't see `optional` tags used for angle brackets in similar places in other license texts (e.g. [AGPL-3.0-only](https://github.com/spdx/license-list-XML/blob/master/src/AGPL-3.0-only.xml), [WTFPL](https://github.com/spdx/license-list-XML/blob/master/src/WTFPL.xml) as a couple random examples).

That said -- primarily I wanted to get the tests passing so that we could merge it and get it added. If you want to iterate on the markup for things like optional tags once it's added, that's fine, it isn't locked in stone for minor markup edits like that. In any event, it won't be appearing on the main license list at https://spdx.org/licenses until the next release, so we can continue to iterate. Either way is fine by me.

kemitchell: Don't let me change the process. I am trying to learn *your* process. It's all fine by me!

swinslow:
@jlovejoy @bradleeedmondson @goneall -- I'd welcome one more set of eyes on this before I push the merge button. 

The license has been approved for inclusion (see #800), this is just looking for a +1 that the markup here looks good to go.

kemitchell:
Thanks to any and all who find time!

src/BlueOak-1.0.0.xml
zvr: @swinslow where in the matching guidelines are the angle brackets described as optional?
I believe that other licenses do not have the `<optional>` tag simply because they haven't thought of it: they have selected one way (with or without the angle brackets) and any instance has to match -- and it usually does, since the text is copied.

swinslow: @zvr Quite possibly it's my own wrong assumptions...

I'd understood this to be the case, but looking at the [matching guidelines](https://spdx.org/spdx-license-list/matching-guidelines) now, I can't point to a specific rule that says so. I don't think this falls into punctuation which is included per rule 5.

Would you be comfortable with us merging this as-is currently, and perhaps we open a separate issue to address this across licenses?

swinslow:
To the folks I tagged, if you do review -- note also the in-line conversation above with @zvr re: angle brackets and matching guidelines.

goneall:
FYI -  there is related known issue in the tools when using single character optional text: https://github.com/spdx/LicenseListPublisher/issues/42

src/BlueOak-1.0.0.xml
zvr: We could merge either of the two versions (with or without the angle brackets) and hope that the people who use it match our choice.
Since the origin URL has a markdown version *with* brackets, I'd rather we also use this one.

Once we manage to fix the issue with optional characters without spaces, we can re-visit for an optimization...

swinslow:
Thanks @goneall and @zvr for your comments. I'm going to go ahead and merge this, and we can continue to optimize / iterate as desired.

Issue #805, opened 2019-03-13
New license request: Parity-6.0.0 [SPDX-Online-Tools]


kemitchell:
**1.** License Name: The Parity Public License 6.0.0
**2.** Short identifier: Parity-6.0.0
**3.** License Author or steward: Artless Devices LLC
**4.** URL: https://paritylicense.com/versions/6.0.0.html
**4.** OSI Status: Not Submitted

kemitchell:
@jilayne, I included some comments in the SPDX-Tools form when submitting, but don’t see them here. Did I screw up? Should I try to reconstruct those notes from memory, in an issue comment?

goneall:
@kemitchell I'll take a look at the online app and see what's the issue is.

goneall:
The notes are going into the notes section of the XML and not in the issues list itself.  I've added [spdx-tools issue #98](https://github.com/spdx/spdx-online-tools/issues/98) to track.

As a workaround, you can go to the submitted licenses page, click on the license you submitted to retrieve the notes.

goneall:
Here are the notes submitted by @kemitchell 

The Parity Public License 6.0.0 is a very strong copyleft or reciprocal public software license. Its terms were professionally drafted and further developed with extensive legal and programmer input.

A spiritual predecessor to Parity, the License Zero Reciprocal Public License, was submitted to OSI for approval, but later effectively withdrawn.

Suffice it to say that Parity remains highly controversial, under various rubrics of "free software", "open source", and standing on its own. However, nomenclature and adjudicative process aside, Parity is a public license for collaborative software development and source code distribution, a descendant of the QPL-1.0, RPL-1.1, and RPL-1.5 line, and far less restrictive than CC-BY-NC-* and CC-BY-NC-ND-*.

This license has been adopted by a diversity of licensors.  Many of those licensors are particularly concerned about the lack of an SPDX identifier, to ensure that those using automated tools can correctly identify Parity-licensed projects as an exception to the rule of permissive-dominant licensing in a variety of language and packaging communities, including npm, Maven, Cargo, and RubyGems.

Three public projects kindly gave permission to be listed in support of this application:

https://freckles.io/#licensing

http://scion.scxml.io/

https://github.com/agreements-network/blackstone

Some projects use Parity as the public license part of a commercial dual-licensing business model. Others, like Monax, developer of Blackstone, use it simply for its strong reciprocal effect.  Artless Devices, the license steward, uses the license in the latter way, as well.

[...]

PR #830, opened 2019-03-22
Add Parity-6.0.0

kemitchell:
This draft PR for #805 gave me another change to make an XML file.

The only interesting bit in this one, I think, is the source line. I vaguely remember another BSD or MIT variant doing something similar, but I'm not sure it was templatized. In any case, I called the placeholder "source".

### DIFF
--- a/src/Parity-6.0.0.xml
+++ b/src/Parity-6.0.0.xml
@@ -25,7 +25,7 @@
       </list>
       <p>To contribute software, publish all its source code, in the preferred form for making changes, through a freely accessible distribution system widely used for similar source code, and license contributions not already licensed to the public on terms as permissive as this license accordingly.</p>
       <p>You are excused for unknowingly breaking 1, 2, or 3 if you contribute as required, or stop doing anything requiring this license, within 30 days of learning you broke the rule.</p>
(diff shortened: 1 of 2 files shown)


swinslow:
@kemitchell If the legal team gives the thumbs-up for including Parity on the list, is this XML file ready to go from your perspective?

If so, and if you want to move it out of "Draft" mode, I assume (without really knowing) that the CI checks will run and we can confirm in parallel whether there's any formatting issues.

ghuntley:
Looking forward to this addition!

kemitchell:
Amending to address some matching issues.

kemitchell:
I was able to get the matching test passing by dropping the `<optional></optional>` around the asterisks that Parity uses to make its warranty disclaimer and damages exclusion conspicuous.

swinslow:
Thanks @kemitchell. If we need to iterate later to clarify and address the asterisks you mentioned, that's the sort of template markup that we can address in subsequent PRs. 

Parity was approved by the Legal Team this morning to add to the list. This PR looks good to me -- @jlovejoy @goneall @bradleeedmondson, would one of you mind just being a second set of eyes to also give this a +1, and then I can merge? Thanks!

PR #831, opened 2019-03-22
add SHL-0.5

jlovejoy:
add Solderpad Hardware License v0.5


jlovejoy:
pushing out for 3.6 along with other version as need to confer with license author on short identifier

src/SHL-0.5.xml
```diff
@@ -0,0 +1,109 @@
+<SPDXLicenseCollection xmlns="http://www.spdx.org/license">
+    <license isOsiApproved="false" licenseId="Solderpad-0.5" name="Solderpad Hardware License v0.5">
```
swinslow: "Solderpad-0.5" is listed here as the licenseId, but the filename is "SHL-0.5" so it can't find the corresponding test file. I'm not sure which one is the preferred ID, but I believe that's why the test is failing so you'll want these to be aligned.

jlovejoy:
updated short identifier as per license author's input, tests should work now...

jlovejoy:
@swinslow - this is still showing changes requested from you, but I think it's all fixed??  Can you check (and merge, since I made PR)?

swinslow:
Yes, sorry, I had missed that you updated the ID in the XML file. I'm figuring out how to clear that and change my review to "approved"  =)  After I do, I'll go ahead and merge. Thanks!

PR #832, opened 2019-03-22
Add GitHub pull-request template for new licenses

kemitchell:
Ref: https://github.com/spdx/license-list-XML/pull/828#issuecomment-475792200

cc @jlovejoy

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE/new-license.md
+++ b/.github/PULL_REQUEST_TEMPLATE/new-license.md
@@ -1,0 +1,17 @@
+New License Request Issue: #
+
+- [ ] Add a plain-text sample of the license to `test/simpleTestForGenerator`, using the license ID as basename and `.txt.` as extension.
+
+- [ ] Add an XML file to `src` with the license ID as basename and `.xml` as extension.
+
+  - [ ] Set `listVersionAdded` on `<license>` if you know the relevant license-list version.
+
+  - [ ] Surround optional text with `<optional>`.
+
+  - [ ] Surround variable text placeholders with `<alt name="$name" match="$pattern"></alt>`.
+
+    For a sense of common names, try `grep -rhoE 'alt name="[^"]+"' src | sort -u`
+
+  - [ ] Place any title text in `<titleText>`.  See [Apache 2.0](/src/Apache-2.0.xml) for an example.
+
+  - [ ] Place any standard license header in `<standardLicenseHeader>`.  See [Apache 2.0](/src/Apache-2.0.xml) for an example.


jlovejoy:
@kemitchell - ah crikey, I should have pointed you to https://github.com/spdx/license-list-XML/blob/master/DOCS/new-license-workflow.md (which I was just editing).  I  like the checklist, I think I'll move it to DOCS and then I can link to that from the workflow doc - might be easier to break it up this way.
(@swinslow  - thoughts?)

kemitchell:
I'd strongly consider a PR template, or a step-by-step guide in `CONTRIBUTING`. Folks willing to make PRs will want to get something done. They won't want to read the whole process doc from soup to nuts.

jlovejoy:
yeah, the CONTRIBUTING needs an update as well... but the soup to nuts is important (and we sorely needed to document that) as there's some important stuff in there.  
I also realized that you probably did the xml file from scratch, but the intention is to be able to pull it from the online tool (if the license was submitted there) - so slightly different starting point. not sure if we want to document two different entry points, though...

kemitchell:
Copying an XML file, hacking it up, and waiting for CI feedback isn't too much to ask of the programmer-types. I was worried about the custom schema, but it wasn't too bad.

The purpose of good process, and good process doc, is really to defend your time, and the time of the team more generally. But in terms of how to go about that, I'd strongly recommend putting as little in front of new eyes as possible. Within that rubric, custom tooling is newer and less familiar than editing text on GitHub. At least for the hacker types. I'm not sure what your pull-request demographics look like. Mostly lawyer-types?

jlovejoy:
at the moment, it's been mainly @swinslow @bradleeedmondson @zvr and @goneall doing everything in terms of PRs here. We discussed on the call this week whether we could require people who submit a license, once mark as accepted, to do the XML and txt file PR themselves, then have a regular team member review and merge. If we went that route, then we'd have to assume the lowest familiarity with Github, I think. 

more documentation tweaking is needed for sure. but better to have more/complete/longer to start with then pair it down. the tooling has also been improving, which then impacts the instructions (good)! more on this all later...

kemitchell:
@jlovejoy I haven't been keeping an eye on your request volume. But I suspect you could require PRs containing a definitive plain-text file for the test suite, and actually be more efficient having knowledgeable folks do the XML conversion than proofing newbies' work.

kemitchell:
@jlovejoy would you like me to tweak this PR to prototype what that might look like?

jlovejoy:
if you haven't used the license submission tool to generate the XML. you might want to try that first (the other Solderpad license would be a good one to try, as I just added it)

and yes, I added an issue in the online tool repo about adding the .txt file automatically :)

kemitchell:
@jlovejoy I went through the tool to submit both #830 and #800, but I'd almost certainly just do a manual PR for any new submission after those.

goneall:
I think most hacker / technical types would prefer to use their own XML editor that they are familiar with and create their own PR (myself included).

I also think it is asking too much of the less technical folks to come up to speed on our XML format in order to submit.  Creating an XML by hand I think will have a high failure rate.

IMHO, supporting both approaches (using the tools to submit new PR's and allowing direct PR's for new licenses) would be important to get the broadest usage.

The tools definitely need to be improved, but there are some good suggestions out there already and we have a good group of motivated GSoC students to help this year.

kemitchell:
@goneall I think you’re right about each of the two kinds of folks who will tend so send license texts in.

On the tools-for-non-nerds side, I would focus on getting a plain text copy. If that’s the official text, the test is pure nerd stuff. Nerds can handle it. That includes me.

kemitchell:
Closing as stalled.

Issue #827, opened 2019-03-22
New license request: SHL-0.51 [SPDX-Online-Tools]


jlovejoy:
**1.** License Name: Solderpad Hardware License v0.51
**2.** Short identifier: Solderpad-0.51
**3.** License Author or steward: Andrew Katz
**4.** Comments: adding to make complete list of open hardware licenses
**5.** URL: https://solderpad.org/licenses/SHL-0.51/
**6.** OSI Status: Not Submitted

jlovejoy:
@andrewjskatz - any comments here?

jlovejoy:
have PR #833 almost ready to merge and have updated short identifier, so closing this

PR #833, opened 2019-03-22
Add SHL-0.51

kemitchell:
Ref: #827


kemitchell:
I did not see an official plain-text rendering on solderpad.org, so I copied the text content of the page and did a little tweaking by hand to get the subsection numbering within Redistribution. Apache uses `(lower-alpha)` for subsections, but SHL uses `arabic` again.

kemitchell:
Probably want to squash-merge this one.

jlovejoy:
pushing this out for 3.6 as it looks like SHL may be used already for short form and need to confer with license author

jlovejoy:
reached out to Andrew Katz, license author re: short identifier and updated file names as per his input (to use SHL)

PR #841, opened 2019-03-29
update for 3.5

jlovejoy:


RELEASE-NOTES.md
```diff
@@ -1,5 +1,25 @@
 # Release Notes for SPDX License List
 
+## version 3.5 - 2019-
+
+New licenses/exceptions added: 7
+
+1. JPNIC
+1. libpng-2.0
```
bradleeedmondson: Presumably this should either be "2." or a non-numeric bullet?

swinslow: Looks odd, but I believe it's valid (GitHub-flavored) Markdown for an ordered list -- see https://guides.github.com/features/mastering-markdown/

PR #843, opened 2019-04-04
Create license inclusion principles file

swinslow:
Copying over text from https://spdx.org/spdx-license-list/license-list-overview

Identical to that site's top content except listed URL for OSD, rather than including
the text of the OSD here.

### DIFF
--- a/DOCS/license-inclusion-principles.md
+++ b/DOCS/license-inclusion-principles.md
@@ -1,0 +1,22 @@
+## License Inclusion Principles
+
+### Background Principles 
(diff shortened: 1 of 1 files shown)


DOCS/license-inclusion-principles.md
```diff
@@ -0,0 +1,22 @@
+## License Inclusion Principles
```
kemitchell:
```suggestion
## License Inclusion Principles

_This document was copied from <https://spdx.org/spdx-license-list/license-list-overview> in early April of 2019. See that pages for the latest version._
```

swinslow: Hi @kemitchell, I think the goal is that we may begin iterating on the inclusion principles here in the repo, rather than on the website. So we'll likely merge this for now as-is, and may eventually move to this being the "latest version" / master copy if it evolves over time.

PR #858, opened 2019-05-16
remove field descriptions

jlovejoy:
remove field descriptions as now in tool directly, so don't need here


bradleeedmondson:
Is it also time to take out the pink to the old Google doc? It made sense to have it there for a while after the GitHub transition, but at this point it's stale and everything that's happening is actually in GitHub.

swinslow:
Looks good to me!

PR #859, opened 2019-05-16
Create license-fields.md

jlovejoy:
create separate page with definition of license fields


jlovejoy:
this is the same text as used on https://spdx.org/spdx-license-list/license-list-overview that we are moving to its own page here. merging as no actual updates

PR #861, opened 2019-05-16
further updates to license inclusion principles

jlovejoy:
move open source definitions into list of inclusion principles to make whole page shorter and move background info to bottom of page

### DIFF
--- a/DOCS/license-inclusion-principles.md
+++ b/DOCS/license-inclusion-principles.md
@@ -1,6 +1,22 @@
-## License Inclusion Principles
+# License Inclusion Principles
 
-### Background Principles 
+
+## Candidate License Analysis
+
+Determining whether a candidate license should be included on the SPDX License List requires the SPDX Legal Team to engage in a case-by-case evaluation of each candidate license based on a number of factors. No one factor is dispositive and factors may weight differently. The ultimate decision will be based on the totality of the factors and SPDX's goals and objectives. Factors include:
+* Does the license substantially comply with one of the open source definitions:
+  * Open Source Definition (OSD) from the Open Source Initiative (OSI) - https://opensource.org/osd
+  * Free Software Definition from the Free Software Foundation (FSF) - https://www.gnu.org/philosophy/free-sw.en.html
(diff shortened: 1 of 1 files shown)


kemitchell:
I have not been involved in the discussion that led to this, and want to respect those that were. However, I might be the one most responsible for incorporation of the license list into validation routines for package managers, which leads to concern about:

> Consider existing use of the license in wild (e.g., wide use of license or use of license in prominent projects)

This becomes something of a Catch-22 for new license forms, since ability to code a choice in metadata that require an SPDX ID has a direct effect on adoption. If you want to impose an adoption hurdle for new submissions, which is totally fine to do, I'd make the line sharper. Perhaps something like:

> Evidence of existing use by 16 or more independent licensors.

jlovejoy:
yes, this scenario was discussed which is why the section now starts with:
"Determining whether a candidate license should be included on the SPDX License List requires the SPDX Legal Team to engage in a case-by-case evaluation of each candidate license based on a number of factors. No one factor is dispositive and factors may weight differently. The ultimate decision will be based on the totality of the factors and SPDX's goals and objectives. "

and besides the 'use-in-wild' factor you cite, is also:
"If it is a newly drafted license, it is drafted such that it can be used by anyone (e.g., it is not company-specific or a "vanity" license)"

Which came out of the discussion that some people are now submitting new licenses to SPDX up-front. Happy to clarify that part further, but having some kind of set number of developers using a license is not going to be practical.

kemitchell:
If the factors are really just starting points for conversation, rather than the process itself, I'd take a few more pains to call that out. For example, you could mention that inclusion is and remains a judgment call of the legal team, but that the following points usually come up in those discussions.

jlovejoy:
as per some discussion on legal call today - I think it'd be easier to merge this PR and then iterate on top of that. There are some word-smithing that could help, but I'd rather have a more granular iteration. if anyone has an objection to this approach, holler soon

swinslow:
Good with merging this as-is, and continuing to iterate as needed.

PR #870, opened 2019-05-30
add timing of release info

jlovejoy:
a couple people have asked about the release cycle timing, so I added that here as we didn't seem to have it stated anywhere. (there are probably other places we could add that, but this seemed like the most logical place to start)


jlovejoy:
added additional details based on call today and put in its own section

PR #872, opened 2019-05-31
Whitespace cleanup

waldyrious:
Removed trailing whitespace in all files, and added an [.editorconfig](https://editorconfig.org/) file to help preserve this from now onward.

Note: The changes are split into separate commits, so they can easily be removed in case some of them are not desired.


goneall:
Since this touches about every file, we'll need to rebase any in-process pull requests.  It looks like there are 7 other PR's right now, so maybe it is the time to do it.

Adding @zvr to the reviewer list - I'd like to get one more opinion before merging.

zvr:
The space deletions at the ends of lines look correct and see no problem merging them.

I don't use `.editorconfig` so only two comments on the file:

- I see a definition for `end_of_line` ; I usually rely on `git config core.autocrlf` to convert on check-out and commit. Wouldn't this definition make it problematic for Windows users to use the file in anything that doesn't understand `.editorconfig` ?

- Since not everyone is using it, we shouldn't rely that the settings hold for all new commits.

If we want to always have "clean" files on releases, we should probably have a script that rstrips every line for each file.

goneall:
I don't use `.editorconfig` either, so I don't have much of an opinion myself. 

I do agree on using git config - that is also my approach.

Should we add a `.gitattributes` file and specify the line endings there? 

@waldyrious What do you think about the `.gitattributes` approach and removing the `.editorconfig`?

waldyrious:
Sounds good 👍 Thanks for the review and comments. I'll try to update the PR later today.

waldyrious:
I've replaced the EOL definition from `.editorconfig` to a new `.gitattributes` file with auto CRLF configuration. However, I couldn't find a way to set `.gitattributes` to apply the other configurations (newline at EOF, and no whitespace at EOL), so I left `.editorconfig` there, even if just as documentation of those settings. (That said, it is already supported by many common editors — see https://editorconfig.org/#download).

In addition, I split the introduction of the "newline at EOF" definition to a separate commit, so that I could actually apply that change in the files that were missing the final newline.

Let me know if you prefer something else.

goneall:
@zvr Just one more checkin with you to see if we're good to merge.

waldyrious:
Thanks for the patience and thoughtful comments!

Issue #866, opened 2019-05-28
add markup to MIT


jlovejoy:
can we add markup to MIT license to accommodate differences in how "the 'Software'" is defined?  I don't think this is legally substantive, but just another way of saying a similar thing:

original says: "to any person obtaining a copy of this software and associated documentation files (the "Software")"

variation: "to any person obtaining a copy of this source file (the "Software")"

This variant of MIT is used in the Linux kernel

see: https://github.com/torvalds/linux/blob/master/drivers/block/xen-blkfront.c

jlovejoy:
also, it would be helpful to add markup in the disclaimer for the odd scenario when someone puts their name in there instead of just "the authors or copyright holders"

silverhook:
Makes sense to me.

If we’re already butchering MIT, I would suggest we wrap the copyright statement into an `<optional>` tag as well. I have reasons: https://github.com/fsfe/reuse-docs/issues/16

jlovejoy:
@silverhook but the copyright notice is always options as per the matching guidelines and has its own tag. I don't think we have also used the optional tag.  @goneall @zvr - can you confirm (or correct)?

goneall:
Currently, the license publisher doesn't create alt or optional text for the `copyrightText` elements in the XML.  I added an issue https://github.com/spdx/LicenseListPublisher/issues/48 to automatically add this when publishing.

I am thinking it should be implemented as a variable text (rather than optional).  The variable text could be empty which would also support optional (e.g. regex `.*`).

It is an easy fix and I could update this before the next release - @jlovejoy @silverhook @zvr - let me know if you agree/disagree with this approach.

silverhook:
I think variable text (regex `.*`) is a good approach.

I did not know that the copyright notice in MIT/BSD-*/etc. was already optional. That is a huge relief :) Do the SPDX Tools also take that into account?

Currently both MIT and BSD texts in license-list-data include the template text of the copyright statement in the license text/template. See e.g.:

- https://github.com/spdx/license-list-data/blob/master/text/BSD-3-Clause.txt
- https://github.com/spdx/license-list-data/blob/master/text/MIT.txt

goneall:
> Do the SPDX Tools also take that into account?

[...]

PR #883, opened 2019-06-24
MIT: add markup for disclaimer

jlovejoy:
add markup for disclaimer for copyright/author name


jlovejoy:
as per #866 - also want to add markup to MIT license to accommodate differences in how "the 'Software'" is defined:

original says: "to any person obtaining a copy of this software and associated documentation files (the "Software")"

variation: "to any person obtaining a copy of this source file (the "Software")"

This variant of MIT is used in the Linux kernel
see: https://github.com/torvalds/linux/blob/master/drivers/block/xen-blkfront.c

can someone with better schema/markup skills help with this part? I'm not sure how to indicate one option or another (versus simply replaceable) - @zvr ??

silverhook:
I would say the `<alt match=".+" name="copyright">` should stretch also over line 12, to include the first line of the copyright statement. Unless I’m 

My RegExp game is pretty weak, but if unless I’m mistaken, just matching the two options you list could be done with simply:

```regexp
this software and associated documentation files \(the "Software"\)|this source file \(the "Software"\)
```

Another approach would be perhaps:

```rexexp
copy of .*? (\(the "Software"\)){1}
```

jlovejoy:
discussed this on 6/27 call - will merge for markup for disclaimer. But difference in definition of "the Software" could be legally different, so holding on that (currently only found in one kernel file)

swinslow:
I agree with merging this for the disclaimer markup change.

jlovejoy:
@swinslow - can you merge as per our - someone-different-merges policy?  :)

swinslow:
done  =)

Issue #836, opened 2019-03-27
New license request: CC-PD [SPDX-Online-Tools]


reversi-fun:
**1.** License Name: Copyright-Only Dedication (based on United States law) or Public Domai
**2.** Short identifier: CC-PD
**3.** License Author or steward: creativecommons.org
**4.** Comments: This license likes spdx/Unlicense,but Terms are different.

creativecommons.org says "CC0 for Public Domain Dedication",
But There is a project using this license, Exsample is calculate-linux.org.
**5.** URL: https://creativecommons.org/licenses/publicdomain/
**6.** OSI Status: Not Submitted

pombredanne:
FWIW this is tracked in ScanCode as https://github.com/nexB/scancode-toolkit/blob/develop/src/licensedcode/data/licenses/cc-pd.yml and https://github.com/nexB/scancode-toolkit/blob/develop/src/licensedcode/data/licenses/cc-pd.LICENSE

swinslow:
I've come across this periodically in the wild, in some cases for works that moved to CC0 for later versions, but used this public domain statement for earlier releases. As the submitter noted, this seems reasonably comparable to something like The Unlicense. 

I think this is appropriate to include on the License List, even though Creative Commons has retired it and doesn't recommend it for new works going forward.

My only comment is that we might want to use the name "Public Domain Certification" and the short identifier "CC-PDC". It appears that Creative Commons separately has a "Public Domain Mark" option, see https://creativecommons.org/publicdomain/mark/1.0/, and we should probably make sure the name + identifier align to the old tool.

mlinksva:
If not CC-PD I would recommend CC-PDDC. It started as a dedication only https://web.archive.org/web/20021222174155/creativecommons.org/licenses/publicdomain but was  characterized as a dedication and certification early on https://web.archive.org/web/20040402125107/http://creativecommons.org/licenses/publicdomain/ and referred to as CC-PDDC various places eg https://wiki.creativecommons.org/wiki/PDM_FAQ#What_about_CC.E2.80.99s_Public_Domain_Dedication_and_Certification.3F_Can_that_tool_still_be_used.3F

swinslow:
@mlinksva Thanks for the background -- this is very helpful. Agreed, CC-PDDC sounds like a good choice for the ID here.

swinslow:
Discussed on 2019-04-18 Legal Team call, agreed that this can be added to the License List.

bradleeedmondson:
[...]

PR #884, opened 2019-06-25
New license: CC-PDDC

jlovejoy:
create new license xml file for CC-PD

### DIFF
--- a/src/CC-PD.xml
+++ b/src/CC-PD.xml
@@ -1,0 +1,40 @@
+<SPDXLicenseCollection xmlns="http://www.spdx.org/license">
+    <license isOsiApproved="false" licenseId="CC-PD" name="Copyright-Only Dedication (based on United States law) or Public Domain Certification" listVersionAdded="3.6">
+        <crossRefs>
+            <crossRef>
+https://creativecommons.org/licenses/publicdomain/</crossRef>
+        </crossRefs>
+        <standardLicenseHeader>
+THIS WORK IS IN PUBLIC DOMAIN:</standardLicenseHeader>
+        <notes />
+        <text>
+            <p>
+THIS WORK IS IN PUBLIC DOMAIN:</p>
+            <p>
+The person or persons who have associated work with this document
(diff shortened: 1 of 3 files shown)


jlovejoy:
this is going to fail b/c I have not made the .txt file yet. But still have questions on short identifier and name - see #836

swinslow:
Based on the comment at https://github.com/spdx/license-list-XML/issues/836#issuecomment-505228524 it looks like ID can be "CC-PDDC" and name can be "Creative Commons Public Domain Dedication and Certification"

jlovejoy:
@bradleeedmondson  or @swinslow  - assuming this passes checks, can one of you review/check and merge?

swinslow:
I think the check is failing because in the XML file the text starts with "THIS WORK IS IN PUBLIC DOMAIN:" (line 12), but in the test file it starts with "Creative Commons Public Domain Dedication and Certification". I can look at addressing this tomorrow (in mtgs all day today), this should be easy for us to take care of. Thanks @jlovejoy!

jlovejoy:
> I think the check is failing because in the XML file the text starts with "THIS WORK IS IN PUBLIC DOMAIN:" (line 12), but in the test file it starts with "Creative Commons Public Domain Dedication and Certification". I can look at addressing this tomorrow (in mtgs all day today), this should be easy for us to take care of. Thanks @jlovejoy!

good catch - not even sure where line 12 came from, as it's not in the original license, so have removed (it also got tagged as the standard license header - gotta watch for weird things being inserted there!)

jlovejoy:
argh, still failing tests!

swinslow:
I think it's because the title in line 1 of the test file ("Creative Commons Public Domain Dedication and Certification") is absent from the XML file... I'd say either remove it from the test file or add it to the XML file and hopefully either would take care of it.

swinslow:
🎉 

@jlovejoy looks good to me, want me to merge?

jlovejoy:
yes please!

PR #886, opened 2019-07-01
Add Swift-exception

### LINKED ISSUE #874: New exception request: Swift-exception
Reported on spdx-legal list 2019-06-04:

Hi,

I would like to request below license exception to be added to the SPDX specification.

This exception is popping up in open source project related the Swift programming language maintained by Apple.

Its text is close in intention to the LLVM-exception but lack the whole GPL-2.0 text found in LLVM.

(...omitting comments re: license submission tool...)

**License Full Name**: Apache 2.0 Runtime Library Exception
**Short Identifier**:  Apache-Runtime-exception-2.0
**Source/URL**:
- https://swift.org/LICENSE.txt
- https://github.com/apple/swift-package-manager/blob/master/LICENSE.txt
- https://github.com/JetBrains/jediterm/blob/64e461c932fb0aac0e05c4546a0002fef47711e7/LICENSE-APACHE-2.0.txt

**OSI Status**: Not Submitted

**License (Exception) Text**:

As an exception, if you use this Software to compile your source code and
portions of this Software are embedded into the binary product as a result,
you may redistribute such product without providing attribution as would
otherwise be required by Sections 4(a), 4(b) and 4(d) of the License.

Thank you,

Thomas Steenbergen
Head of Open Source, HERE Technologies

swinslow:
Fixes #874 

### DIFF
--- a/src/exceptions/Swift-exception.xml
+++ b/src/exceptions/Swift-exception.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <SPDXLicenseCollection xmlns="http://www.spdx.org/license">
-   <exception licenseId="Swift-exception" name="Swift Exception">
+   <exception licenseId="Swift-exception" name="Swift Exception" listVersionAdded="3.6">
       <crossRefs>
          <crossRef>https://swift.org/LICENSE.txt</crossRef>
          <crossRef>https://github.com/apple/swift-package-manager/blob/7ab2275f447a5eb37497ed63a9340f8a6d1e488b/LICENSE.txt#L205</crossRef>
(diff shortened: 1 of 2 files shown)


swinslow:
@jlovejoy This looks ready to merge, do you agree?

jlovejoy:
license list version? 
is the OSI approved tag required in all entries? (not sure myself...)

swinslow:
I thought about those but didn't include them b/c this is an exception rather than a license list entry. For other recent additions to the exception list (e.g. GPL-CC-1.0) we haven't included those bits...

jlovejoy:
hmmm... seems like we should always have license list version, tho?  we count added exceptions!

swinslow:
I'll try it and see -- not sure whether the exceptions schema permits it to be added, but if it does then I agree, we should include it!

jlovejoy:
already done. we have used it in the past, just missed the last one...

swinslow:
OK, thanks  =)  Noted for future reference! Thanks for adding it here.

goneall:
FYI - to see if anything is required or optional, you can check the [schema](https://github.com/spdx/license-list-XML/blob/master/schema/ListedLicense.xsd).  Not that easy to read, but you'll find all the elements documented.   `minOccurs="0"`,is optional and `minOccurs="1"` is required.

PR #888, opened 2019-07-01
new license: SSPL-1.0

jlovejoy:


swinslow:
I think the test is failing because the test file is called "SSPL-1.0-", rather than "SSPL-1.0.txt"...

src/SSPL-1.0.xml
```diff
@@ -0,0 +1,643 @@
...
+          including scripts to control those activities. However, it does
+          not include the work's System Libraries, or general-purpose tools
+          or generally available free programs which are used unmodified in
+          performing those activities but which are not part of the work.
+          For example, Corresponding Source includes interface definition
+          files associated with source files for the work, and the source
+          code for shared libraries and dynamically linked subprograms
+          that the work is specifically designed to require, such as by
+          intimate data communication or control flow between those
+                        <br></br>
```
swinslow: This `<br></br>` looks like a stray line break mid-paragraph, can this come out?

src/SSPL-1.0.xml
```diff
@@ -0,0 +1,643 @@
...
+                    <p>
+&#8220;Service Source Code&#8221; means the Corresponding Source for the Program or the modified version, and the Corresponding Source for all programs that you use to make the Program or modified version available as a service, including, without limitation, management software, user interfaces, application program interfaces, automation software, monitoring software, backup software, storage software and hosting software, all such that a user could run an instance of the service using the Service Source Code you make available.</p>
```
swinslow: I think these `&#8220;` and `&#8221;` tags can just be regular quotation marks instead -- not sure if this will mess up the matching...

src/SSPL-1.0.xml
```diff
@@ -0,0 +1,643 @@
...
+                </item>
+                <item>
+                    <bullet>14.</bullet>Revised Versions of this License.
+                    <p>
+MongoDB, Inc. may publish revised and/or new versions of the Server Side Public License from time to time. Such new versions will be similar in spirit to the present version, but may differ in detail to address new problems or concerns.</p>
+                    <p>
+Each version is given a distinguishing version number. If the Program specifies that a certain numbered version of the Server Side Public License &#8220;or any later version&#8221; applies to it, you have the option of following the terms and conditions either of that numbered version or of any later version published by MongoDB, Inc. If the Program does not specify a version number of the Server Side Public License, you may choose any version ever published by MongoDB, Inc.</p>
```
swinslow: Same comment as line 587 above re: these numbered tags vs. regular quote marks that are used elsewhere in the license.

swinslow:
Hi @jlovejoy, I'm looking at this also, will push some edits if I can.

swinslow:
So I found the latest error, but when I'm running it on my local system I'm getting a subsequent error that for the life of me I can't figure out what is wrong. We'll see if it comes up from the CI build also...

goneall:
I looked at it too - can't figure out why it is failing.  I'll do some deeper research with the tools running in debug.

goneall:
Found it.   Bug in the license matching tool (sorry about that).

Here's the issue: https://github.com/spdx/tools/issues/198

You can work around it by removing the line breaks between "copyright" and "holder".  Attached is a working SSPL-1.0.txt test file.

Suggest going with the workaround for now, I'll work on a fix but it takes a lot of time and steps to get this into the license-list-xml repo (update the SPDX tools with the fix, publish a new version, update the publisher to use the fixed tools, publish a new version, update the makefile in license-list-xml to use the new publisher)
[SSPL-1.0.txt](https://github.com/spdx/license-list-XML/files/3352728/SSPL-1.0.txt)
.

jlovejoy:
oh geez, didn't see your last comment before doing a bunch of re-work (that was probably unnecessary...)

jlovejoy:
hallelujah, it passed!!

PR #891, opened 2019-07-02
Add variable text around all copyright elements when producing the matching templates

goneall:
This requires an update to the tools and removing any optional and alternate text elements currently inside copyright text.

### DIFF
--- a/src/TCL.xml
+++ b/src/TCL.xml
@@ -6,12 +6,15 @@
          <crossRef>https://fedoraproject.org/wiki/Licensing/TCL</crossRef>
       </crossRefs>
     <text>
-      <copyrightText>
-         <p>This software is copyrighted by the Regents of the University of California, Sun Microsystems, Inc.,
-        Scriptics Corporation, ActiveState Corporation and other parties. The following terms apply to all
+	  <p>
+         <copyrightText>
+            This software is copyrighted by the Regents of the University of California, Sun Microsystems, Inc.,
+            Scriptics Corporation, ActiveState Corporation and other parties. 
+	     </copyrightText>  
+		The following terms apply to all
         files associated with the software unless explicitly disclaimed in individual files. 
       </p>
-      </copyrightText>
+
 
       <p>The authors hereby grant permission to use, copy, modify, distribute, and license this software and its
          documentation for any purpose, provided that existing copyright notices are retained in all copies and
--- a/src/TCL.xml
+++ b/src/TCL.xml
@@ -7,10 +7,9 @@
       </crossRefs>
     <text>
       <copyrightText>
-         <p>This software is copyrighted by
-        <alt match=".+" name="authors">the Regents of the University of California, Sun Microsystems, Inc.,
-        Scriptics Corporation, ActiveState Corporation</alt> and other parties. The following terms apply to all
-        files associated with the software unless explicitly disclaimed in individual files.
+         <p>This software is copyrighted by the Regents of the University of California, Sun Microsystems, Inc.,
+        Scriptics Corporation, ActiveState Corporation and other parties. The following terms apply to all
+        files associated with the software unless explicitly disclaimed in individual files. 
       </p>
       </copyrightText>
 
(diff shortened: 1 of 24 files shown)


goneall:
@swinslow @jlovejoy Could one of you  take a look at the license changes just to makes sure I didn't make any copy/paste errors while removing the `<optional>` or `<alt>` tags within the copyright fields?  I reviewed it couple times, but I just want a separate set of eyes to make sure I didn't miss anything.

I want to merge this in prior to the next release.

jlovejoy:
@goneall having a look now. to make sure I understand - we are removing the <optional> and <alt> tags and will just have <copyrightText> tag - which will tie to the Matching Guideline that says copyright notice doesn't need to match, correct?
will this also mean this text will appear red in the website pages?

goneall:
> we are removing the and tags and will just have tag - which will tie to the Matching Guideline that says copyright notice doesn't need to match, correct?

Correct

> will this also mean this text will appear red in the website pages?

It should.  I won't know for sure until we publish the next release.  We should probably publish to preview to make sure everything looks OK BTw.

src/TU-Berlin-2.0.xml
```diff
@@ -6,8 +6,7 @@
     </crossRefs>
     <text>
       <copyrightText>
-        <p>Copyright
-		      <alt match=".+" name="copyrightYears">1992, 1993, 1994</alt> by Jutta Degener and Carsten Bormann,
+        <p>Copyright 1992, 1993, 1994 by Jutta Degener and Carsten Bormann,
 	      </p>
```
jlovejoy: does this need a new <p> for the last name?

src/TCL.xml
```diff
@@ -7,10 +7,9 @@
       </crossRefs>
     <text>
       <copyrightText>
-         <p>This software is copyrighted by
-        <alt match=".+" name="authors">the Regents of the University of California, Sun Microsystems, Inc.,
-        Scriptics Corporation, ActiveState Corporation</alt> and other parties. The following terms apply to all
-        files associated with the software unless explicitly disclaimed in individual files.
+         <p>This software is copyrighted by the Regents of the University of California, Sun Microsystems, Inc.,
+        Scriptics Corporation, ActiveState Corporation and other parties. The following terms apply to all
+        files associated with the software unless explicitly disclaimed in individual files.
```
jlovejoy: I dont' think the sentence beginning with "The following terms" should be part of the copyright notice and tag

jlovejoy:
looks good, made a couple comments where a change might be needed.

looks like the <P> tags get thrown in in places that really ought to be a line break (or nothing), for example when "All rights reserved" is on a second line, but we can see how things display and clean up from there

src/TU-Berlin-2.0.xml
goneall: Not sure I understand - there is a paragraph separating the last name and the university

src/TCL.xml
goneall: I agree - just pushed a fix

goneall:
@jlovejoy Thanks for the quick review!

The formatting will be driven by the `<br />` and `<p>` tags.  I just duplicated the same formatting as the original xml files.  Should render the same as before this PR.

goneall:
@jlovejoy Could you take one more look after my updates?  If it looks good, I'll go ahead and merge.

src/TCL.xml
jlovejoy: so, now the beginning <p> tag is before the copyrighttext tag, and end </p> tag after the bit of text. not sure it matters, but that seems a bit weird?

goneall: The paragraph tag is used only for formatting.  You can't have a paragraph start tag outside the copyright tag and the end tag inside the copyright tag.  So, to keep the formatting the same as the original, I moved the start paragraph tag outside the copyrighttext tag and left the end tag where it was.

jlovejoy: i see. just looks weird, but fine by me!

PR #893, opened 2019-07-03
BSD-3-Clause-LBNL: change to OSI-approved

jlovejoy:
OSI recently approved this license, so updating that info in our database


jlovejoy:
@swinslow - can you check and merge? This is so small, would be nice to just sweep it into 3.6 :)

swinslow:
I like when they're this easy  :)

Issue #892, opened 2019-07-03
Proposal: Add http URL consideration to matching guidelines


goneall:
Propose adding the following to the matching guidelines:

13. URL references

12.1 Purpose:  To avoid a license mismatch merely because a different in protocol for a URL reference (e.g. ```http://``` vs. ```https://```.

12.1.1 Guideline:  The two URL protocol prefixes ```http://``` and ```https://``` will be considered equivalent and interchangeable for matching purposes.

zvr:
+1

But let's use the official nomenclature:

> `http` and `https` URI schemes shall be considered equivalent for matching purposes.

(or something like that)

goneall:
Closing this as a duplicate of #633  This issue has been updated with the proposal text.

PR #895, opened 2019-07-03
Remove optional 's' from URLs

goneall:
The `http<optional>s</optional>://...` are causing formatting issues in JSON and plain text output.  Although we could add the spacing attribute to fix this, I believe it would be easier for humans to read and for computers to match if we adopted the matching guidelines suggested in issue #892 and removed these optional elements.


goneall:
Looks like I need to update the tools version before we can merge this... closing for now

PR #896, opened 2019-07-04
Add spacing to fix formatting issues for JSON and plain text files

goneall:
Resolves issue where spaces are being inserted in the non-HTML versions of the license text (e.g. license-text, JSON).

### DIFF
--- a/src/AFL-1.1.xml
+++ b/src/AFL-1.1.xml
@@ -16,7 +16,7 @@
          (the "Licensor") has placed the following notice immediately following the copyright notice for the
          Original Work:</p>
          <standardLicenseHeader>
-         <p><optional>"</optional>Licensed under the Academic Free License version 1.1.<optional>"</optional></p>
+         <p><optional spacing="none">"</optional>Licensed under the Academic Free License version 1.1.<optional spacing="none">"</optional></p>
          </standardLicenseHeader>
       </optional>
 
--- a/src/AGPL-3.0-only.xml
+++ b/src/AGPL-3.0-only.xml
@@ -21,7 +21,7 @@
       <p>
         You should have received a copy of the GNU Affero General Public License
         along with this program. If not, see
-	      &lt;http<optional>s</optional>://www.gnu.org/licenses/&gt;</p>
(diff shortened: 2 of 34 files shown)


goneall:
Resolves https://github.com/spdx/license-list-data/issues/42

goneall:
@jlovejoy @swinslow @zvr I don't think this needs a detailed review, but I wanted to alert you to a change which should just impact the formatting for text and JSON files.  It should not impact the HTML or website.  I did a spot check of the text formatting and made sure the matching still worked.

I would like to merge before the next release, let me know if you want time to review or if you have see any issues.

jlovejoy:
@goneall - had a look, not sure I'm the best person to do so, but fine by me. looks like the addition is either value= none, after, or before. might be good to understand the difference and document that, as this is something I'm assuming we want to remember to do for future markup like these?

goneall:
@jlovejoy Good point on the documentation.  The XML Schema includes documentation, but it is quite hard to read IMHO.  Perhaps we should add something to the DOCS folder that describes the details for the XML format?

zvr:
I don't think it will break anything, but the changes are a little inconsistent in the `spacing` attribute values.
The http/https should of course be `none` -- no question there.
For an optional comma, `after` is also correct.

For things enclosing a sentence (like optional quotes before and after), I'd go with `before` and `after` instead of `none` and `none`. You also use sometimes the pair `none` and `after`, which is... strange.

zvr:
... and once we do #892 we can go and delete many of these... ;)

goneall:
> For things enclosing a sentence (like optional quotes before and after), I'd go with before and after instead of none and none. You also use sometimes the pair none and after, which is... strange.

I was taking into account paragraph and line breaks before or after the quote.  Since you can't do a search/replace for the enclosing paragraphs, I hand edited each one and if there was a line break before a starting quote, I used none.   I was a bit worried the it may force a space after the linebreak if I did a before.  Similar approach on ending.

goneall:
I'll go ahead and merge - I can test out the formatting for the spacing before following a line break and update for consistency in a second pass.

PR #901, opened 2019-07-11
Add softwareName replacement of Vim license

othree:
Hi All

I am working on let GitHub to detect **Vim License**. Had some discussion with Bram(Vim author)[1] and GitHub staff[2]. We found the Vim License text have some variable text placeholders. If any other project(most are Vim scripts/plugins) uses Vim License. These parts should be replaced to corresponding value to make it right to use. And it will be great if SPDX license list have definitions of these placeholders.

This PR adds these placeholders. Only one variable: `softwareName`. The name of the variable is from existing one in [SGI-B][3]. Not sure is it fits for our case(Can script/plugin also called software?). I used to use `projectName`. So if `softwareName` is not fit our case. I can change to `projectName`.

[1]:https://groups.google.com/forum/#!searchin/vim_dev/license%7Csort:date/vim_dev/DlTVMew1ZSo/s3KkBdJ5BgAJ
[2]:https://github.com/licensee/licensee/issues/385#issuecomment-506386613
[3]:https://github.com/spdx/license-list-XML/blob/ccfec34109bec311c5da8f236274eedaab6fb4c1/src/SGI-B-1.0.xml#L288

Refs:

* [Post at mailing list][4]

[4]:https://lists.spdx.org/g/Spdx-legal/topic/32343388#2635

### DIFF
--- a/src/Vim.xml
+++ b/src/Vim.xml
@@ -12,15 +12,15 @@
       <list>
         <item>
             <bullet>I)</bullet>
-          There are no restrictions on distributing unmodified copies of Vim except that they must include
-             this license text. You can also distribute unmodified parts of Vim, likewise unrestricted
+          There are no restrictions on distributing unmodified copies of <alt match=".+" name="softwareName">Vim</alt> except that they must include
+             this license text. You can also distribute unmodified parts of <alt match=".+" name="softwareName">Vim</alt>, likewise unrestricted
              except that they must include this license text. You are also allowed to include executables
-             that you made from the unmodified Vim sources, plus your own usage examples and Vim
+             that you made from the unmodified <alt match=".+" name="softwareName">Vim</alt> sources, plus your own usage examples and Vim
              scripts.
         </item>
         <item>
             <bullet>II)</bullet>
-          It is allowed to distribute a modified (or extended) version of Vim, including executables and/or
+          It is allowed to distribute a modified (or extended) version of <alt match=".+" name="softwareName">Vim</alt>, including executables and/or
              source code, when the following four conditions are met:
           <list>
                <item>
@@ -29,14 +29,14 @@
             </item>
                <item>
                   <bullet>2)</bullet>
-              The modified Vim must be distributed in one of the following five ways:
+              The modified <alt match=".+" name="softwareName">Vim</alt> must be distributed in one of the following five ways:
               <list>
                      <item>
                         <bullet>a)</bullet>
-                  If you make changes to Vim yourself, you must clearly describe in the distribution how to contact
-                    you. When the maintainer asks you (in any way) for a copy of the modified Vim you distributed,
+                  If you make changes to <alt match=".+" name="softwareName">Vim</alt> yourself, you must clearly describe in the distribution how to contact
+                    you. When the maintainer asks you (in any way) for a copy of the modified <alt match=".+" name="softwareName">Vim</alt> you distributed,
                     you must make your changes, including source code, available to the maintainer without fee.
-                    The maintainer reserves the right to include your changes in the official version of Vim. What
+                    The maintainer reserves the right to include your changes in the official version of <alt match=".+" name="softwareName">Vim</alt>. What
                     the maintainer will do with your changes and under what license they will be distributed is
                     negotiable. If there has been no negotiation then this license, or a later version, also
                     applies to your changes. The current maintainer is Bram Moolenaar &lt;Bram@vim.org&gt;. If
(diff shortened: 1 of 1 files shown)


src/Vim.xml
```diff
@@ -29,14 +29,14 @@
...
               <list>
                      <item>
                         <bullet>a)</bullet>
-                  If you make changes to Vim yourself, you must clearly describe in the distribution how to contact
-                    you. When the maintainer asks you (in any way) for a copy of the modified Vim you distributed,
+                  If you make changes to <alt match=".+" name="softwareName">Vim</alt> yourself, you must clearly describe in the distribution how to contact
+                    you. When the maintainer asks you (in any way) for a copy of the modified <alt match=".+" name="softwareName">Vim</alt> you distributed,
                     you must make your changes, including source code, available to the maintainer without fee.
-                    The maintainer reserves the right to include your changes in the official version of Vim. What
+                    The maintainer reserves the right to include your changes in the official version of <alt match=".+" name="softwareName">Vim</alt>. What
```
mlinksva: I don't understand how making Vim replaceable here is compatible with hardcoding the maintainer of Vim's name in the following sentence.

othree: All "maintainer" in the license text refers to "Vim Maintainer" according to Bram [[1][]]

Which means if someone distributed a modified software with Vim License and not open source.
Bram as Vim maintainer should be the one to ask for the source.
But this case is very rare. Because Vim script/plugin are not easy to distribute without open source.

[1]: https://groups.google.com/d/msg/vim_dev/DlTVMew1ZSo/h1C0MTBYCgAJ

mlinksva: Ok, that's odd but I understand, thanks for explaining. Reading with that in mind, I have at least one other comment elsewhere.

mlinksva: Actually, in the same place. I still don't understand how replacing the Vim here makes sense given that maintainer always refers to the Vim maintainer. If person X releases software Y under the Vim License, isn't X and not the Vim maintainer in a position to make changes to the official version of Y?

I'd be curious to see an example of how this is done in the wild, I might be missing something obvious.

othree: The license text is old. I believe the first version is not consider other software to use it. So the word is not perfectly match it.

In this case. If there is a modified software Y called Y'. And the source code of Y' is not distributed. Vim maintainer should be the one to request it. And decide to give codes to X. Explanation by Bram[[2][]]:

> Note that "Vim maintainer" is different, that is not the product name
and should remain as-is.  In the bigger context this part is a bit
confusing, but we can't change it now.  The basic idea is that if
someone wants to make changes to [product name] but not publish the
changed code, the maintainer will have the right to decide about that.
This if you publish a plugin with the Vim license, I get to decide what
happens when someone makes a modified version and doesn't want to
publish it.  In practice this never happened, thus it's a precaution,
while keeping the door open for a commercial spin-off. 

Two points:

* Vim Script/Plugin is always distributed with source codes. This situation is rare to Vim scripts/Plugin.
* This never happened before. So there is no real case in the wild now.

[2]:https://groups.google.com/d/msg/vim_dev/DlTVMew1ZSo/QddqQZsMCgAJ

mlinksva: I understand the part about only the Vim maintainer having the right to demand the source. That's not what I'm asking about here, rather:

> The maintainer reserves the right to include your changes in the official version of <alt match=".+" name="softwareName">Vim</alt>.

It seems like this instance of Vim ought not be replaceable, i.e., it should not be wrapped with `alt`. How is the Vim maintainer supposed to include your changes in the official version of software they don't maintain, ie non-Vim software? Make a PR/send a patch, I guess? OK, maybe that makes sense, even odder though in my book. Just thinking though it though, thanks for patiently explaining.

...

My question about 'in the wild' was not about whether Bram had made source code requests, but whether the Vim license had been used by other projects with Vim replaced: would this PR result in a SPDX template powered matcher running against the universe match any files that would not have been matched without this PR, or would it just expend more energy? :smile:

othree: > The maintainer reserves the right to include your changes in the official version of Vim.

I didn't confirm this part. But we can ask it at vim_dev. I just think for this case Bram don't need to really decide for the non-Vim software. He only decide should these changes open or close. Not 100% match the license text but ok for me.

In the wild, I know [tagbar][] uses modified version of Vim License. But Tagbar License(the modified version) replaced everything. Which might be what you are expect to see(my first version also replaces all these information).

I have 1 or 2 project use software name replaced Vim License text. For other projects using Vim License. Most of they didn't change the License text. But I don't think it's correct way to use Vim License. So my next step is to help them update their license text.

PS: I believe only several people know about this "Vim Maintainer" fact before I start working on this.

[tagbar]:https://github.com/majutsushi/tagbar/blob/master/LICENSE

swinslow:
Hi @othree, thanks for submitting this!

I agree with many of @mlinksva's comments above. In particular, it would be very helpful if you can post here some examples of where else this license has been used in the wild, with replacements in place of "Vim".

I don't have experience with the Vim community, but my gut reaction is that this license seems tightly integrated with Vim specifically as a project. The specific individual named as the "maintainer" in the license text is one example, but also things like e.g. the list of URLs where a change to the maintainer would be announced; the reference to a specific Vim command in section II(3); etc.

When we add "optional" tags to license texts, it's typically where it can be fully replaceable and used with any sort of FOSS software. This one seems to me like it's so closely bound to Vim that I'm having some trouble seeing how it would be used with other tools.

All that said, though, SPDX's core purpose is to help identify what we see happening in practice -- not necessarily to opine on how the license text works. So if you can provide links to other instances of this being used in the wild, that would be extremely helpful. (It's fine to send us links to your own projects, but it would also be helpful to see instances where this is being used by other people as well.)  Thank you!

othree:
H @swinslow 

Thanks for the reply.

There are several replaced license can be found in the wild(not much).

* https://github.com/majutsushi/tagbar/blob/master/LICENSE
* https://github.com/kidaa/aur/blob/master/vim-languagetool/license.txt
* https://github.com/Teuse/hopper/blob/master/LICENSE
* https://github.com/Teuse/ogrep/blob/master/LICENSE

But all above peojects replaced every `Vim` word in the text. Including the license name. There is a project belongs to me uses the Vim License with the replaced text. This is also a tool for creating the replaced license text.

* https://github.com/othree/vim-license/blob/master/LICENSE

I have several vim plugin uses "same as vim" now. And I plan to update their license file after there is a result for this PR.

I can understand the "in practice" you mean. In practice there is not much use case for this. But what I concerns about is if I released a software with Vim License. But the software's name is not replaced in the license text. Is it means the license is not working to the software at all?

Issue #885, opened 2019-06-27
New license request: Polyform-1.0.0 licenses [EDITED]


MarkAtwood:
The https://github.com/polyformproject is a thing, working to develop a CC- like templated source available license set.

We should otherwise avoid assigning license IDs beginning with "PF-" and encourage the Polyform Project to use a prefix of "PF-".

jlovejoy:
FYI @heathermeeker

kemitchell:
@jlovejoy I think both @heathermeeker and I would prefer `Polyform-`.

We're planning to release 1.0.0 versions of a few licenses on that pattern this weekend. I will handle the PRs here.

swinslow:
@kemitchell has added a PR for the Polyform licenses at https://github.com/spdx/license-list-XML/pull/903/files

I'm going to tag this as "discuss on legal call" as it looks to me like each of these is intended to place limits on use, redistribution or both.

heathermeeker:
That is correct, Polyform licenses contain license scope limitations.  They are similar, but not identical, to those imposed by Creative Commons NC https://spdx.org/licenses/CC-BY-NC-4.0.html and Creative Commons ND https://spdx.org/licenses/CC-BY-ND-4.0.html. Polyform project is intended, in some measure, to be an analog to Creative Commons for software licenses.

pombredanne:
Most proprietary licenses have been not been added in the past on the grounds that they were proprietary or not OSI-approved or not open source-like or were not (yet) used by any open source projects... 

I am all for adding as many licenses to the list as relevant and evolve the ["License Inclusion Principles"](https://spdx.org/spdx-license-list/license-list-overview) accordingly. Yet I am a tad confused of late as to how these principles seem to be evolving tacitly. 

I would be curious to find out if these new licenses meet any or all of the current principles and if yes, if this sets some precedent.

swinslow:
Discussed on 2019-08-08 legal team call, with a couple of outcomes: 

1) For this issue specifically, the original question was whether to informally "set aside" a prefix for licenses from Polyform. From this thread it sounds like the preferred prefix would be "Polyform-" rather than "PF-". The legal team agreed that we wouldn't adopt other licenses from unrelated submissions with the prefix "Polyform-".

[...]

PR #903, opened 2019-07-16
Polyform 1.0.0 Licenses

kemitchell:
This pull request includes the recently recent 1.0.0 source-available licenses from [Polyform Project](https://polyformproject.org): Noncommercial, Strict, Small Business, Internal Use, and Free Trial.

Like BlueOak-1.0.0, Polyform licenses are published in Markdown markup. Those who prefer to peruse rendered HTML can visit https://polyformproject.org/licenses/.

cc @heathermeeker @tieguy

### DIFF
--- a/src/PolyForm-Free-Trial-1.0.0.xml
+++ b/src/PolyForm-Free-Trial-1.0.0.xml
@@ -1,38 +1,33 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <SPDXLicenseCollection xmlns="http://www.spdx.org/license">
-  <license isOsiApproved="false" licenseId="Polyform-Small-Business-1.0.0" name="Polyform Small Business License 1.0.0" listVersionAdded="3.9">
+  <license isOsiApproved="false" licenseId="PolyForm-Free-Trial-1.0.0" name="PolyForm Free Trial License 1.0.0" listVersionAdded="3.9">
     <crossRefs>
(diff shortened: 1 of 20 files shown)


kemitchell:
Rebased to use the column-wrapped versions as tests for all licenses.

kemitchell:
CI is green. #904 helped a lot!

kemitchell:
Should I plan to attend tomorrow's legal team call if possible?

jlovejoy:
I think it'd be best to hold on these until we complete the license inclusion principles update work. (Admittedly, under the current guidelines, these licenses would probably not be accepted, but given the conversations on changes to the inclusion principles, this may change)

rkarthik007:
Wanted to leave a comment here saying that [Yugabyte](https://github.com/yugabyte/yugabyte-db#license) currently uses Polyform licenses for the Platform product. Our rationale behind adopting this was as follows: while our database product (YugabyteDB) is 100% open-source, our commercial Yugabyte Platform product is source-available. We wanted to pick a standard license rather than craft our own license which becomes confusing for our users.

There is a need for such a standardization across the board, otherwise figuring out the nuances of each vendor's license becomes onerous.

kemitchell:
@heathermeeker, should the SPDX IDs for the licenses be `PolyForm-*` instead of `Polyform-*`?

kemitchell:
Per #885, this PR is being split so that -Noncommercial and -Small-Business can land, with the rest still pending.

PR #904, opened 2019-07-16
Add helper script to test one license (close #880)

### LINKED ISSUE #880: How to test a single license
@goneall, is there an easy way to check matching on a single license at a time? Perhaps we could write that in as a Make target.

I haven't done `time make` on my machine, but the full check takes a few minutes at least. I see Travis CI is similar.

kemitchell:
This PR adds a short, hacky BASH script that automates testing a single license at a time by suffixing all other license XML files.

### DIFF
--- a/test-one-license
+++ b/test-one-license
@@ -1,0 +1,22 @@
+#!/bin/bash
+if [ -z "$1" ]; then
+  echo "Usage: <License Identifier, e.g. 'MIT'>"
+  exit 1
+fi
+
+SKIP_SUFFIX=".skip"
+
+# Add the skip suffix to the XML files for all licenses,
+# except the one we want to test.
+while IFS= read -d '' -r file; do
+  mv "$file" "$file$SKIP_SUFFIX"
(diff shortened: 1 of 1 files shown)


kemitchell:
cc @goneall

goneall:
@kemitchell With the last license release, I update the License List Publish tool to take a single file as a parameter rather than the entire directory (see https://github.com/spdx/LicenseListPublisher/pull/52/files)

I was thinking we could update the make file and only run this against the changed files using a single file as a parameter.

You may also be able to use this feature of the tool for testing a single license.

kemitchell:
@goneall, that all sounds good, though I don’t quite follow in every detail.

The proposed script works today, and pretty quickly on an SSD. Can we merge?

goneall:
> The proposed script works today, and pretty quickly on an SSD. Can we merge?

Will do

swinslow:
Thank you @kemitchell! This will be extremely useful for speeding up the test process for new / edited licenses...

PR #908, opened 2019-07-29
fix and update from http://www.mirbsd.org/MirOS-Licence.asc

mirabilos:
-  the name is “The MirOS Licence”, note the spelling; fixed
-  short ID is “MirBSD” not “MirOS”; kept as-is because I assume otherwise all hell would break loose
-  fix totally fucked-up text apparently copied from OSI by replacing with the official text’s ASCII transcription
- add standard licence header, from official licence site http://www.mirbsd.org/MirOS-Licence

### DIFF
--- a/src/MirOS.xml
+++ b/src/MirOS.xml
@@ -1,12 +1,31 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <SPDXLicenseCollection xmlns="http://www.spdx.org/license">
-   <license isOsiApproved="true" licenseId="MirOS" name="MirOS License">
+   <license isOsiApproved="true" licenseId="MirOS" name="The MirOS Licence">
       <crossRefs>
          <crossRef>https://opensource.org/licenses/MirOS</crossRef>
       </crossRefs>
+    <standardLicenseHeader>
+      Copyright © <alt name="copyright" match=".+">year, year, year, …<br/>
+           First M. Last &lt;user@host.domain&gt;</alt>
+      <p>
+	Provided that these terms and disclaimer and all copyright notices<br/>
+	are retained or reproduced in an accompanying document, permission<br/>
+	is granted to deal in this work without restriction, including un‐<br/>
+	limited rights to use, publicly perform, distribute, sell, modify,<br/>
+	merge, give away, or sublicence.</p>
+      <p>
+	This work is provided “AS IS” and WITHOUT WARRANTY of any kind, to<br/>
+	the utmost extent permitted by applicable law, neither express nor<br/>
+	implied; without malicious intent or gross negligence. In no event<br/>
+	may a licensor, author or contributor be held liable for indirect,<br/>
+	direct, other damage, loss, or other issues arising in any way out<br/>
+	of dealing in the work, even if advised of the possibility of such<br/>
+	damage or existence of a defect, except proven that it results out<br/>
+	of said person’s immediate fault when using the work as intended.</p>
+    </standardLicenseHeader>
     <text>
       <titleText>
-         <p>MirOS License</p>
+         <p>The MirOS Licence</p>
       </titleText>
       <copyrightText>
          <p>Copyright [YEAR]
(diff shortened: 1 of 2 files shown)


swinslow:
Hi @mirabilos, apologies for slow response on this one.

I think the technical changes are fine (like removing the underscores in the title lines). This pre-dates my involvement in SPDX but I'd guess yes, the current version was probably copied as-is from OSI's website.

On the substantive changes to the "Instructions" text, the links, etc., I'm a bit more hesitant to include those as _modifications_ to the existing license entry on the list. We generally would not make substantive changes to a license's existing text, but would instead add a new entry to the license list for it. 

Here, where the changes are substantive but are to the optional instructions, not the primary license text itself, I'm less certain. But some of the rationale changes may arguably change the interpretation of the license (e.g. the removal of the "We believe you are not liable..." paragraph) so I don't think we can treat the instructions as irrelevant to the license, if they're going to be present in the license list entry.

@jlovejoy @goneall Do you have any thoughts on this one, particularly since the original looks like it was pulled from the OSI website in the early days of the list?

mirabilos:
Hi Steve,

>On the substantive changes to the "Instructions" text, the links,
>etc., I'm a bit more hesitant to include those as _modifications_ to
>the existing license entry on the list. We generally would not make
>substantive changes to a license's existing text, but would instead
>add a new entry to the license list for it.

the licence text is the first block *ONLY*. The rest is just copied
from the text file we carry the licence in and could just as well
be removed.

>But some of the rationale changes may arguably change the
>interpretation of the license (e.g. the removal of the "We believe you

That one was rather historic and refers to licences in general, not
this specific one. I’ve got on my long-term TODO to add a web page
with interpretations, but would prefer to keep them out of the main
licence text etc. (and I removed it as it was not worded very well
anyway and, as I said, as something general has no place there).
Historic raisins, all over the place…

bye,
//mirabilos
-- 
I believe no one can invent an algorithm. One just happens to hit upon it
when God enlightens him. Or only God invents algorithms, we merely copy them.
If you don't believe in God, just consider God as Nature if you won't deny
existence.		-- Coywolf Qi Hunt

swinslow:
Hi @mirabilos, thanks for the responses and for confirming that the "license text" proper is just the first block.

Oddly, I see that the existing XML file already has all of the rest (Instructions, Rationale, etc.) contained within an `<optional>` tag. That should mean that on the [rendered license list entry](https://spdx.org/licenses/MirOS.html) it ought to be showing up as blue text, to indicate that it's optional and can be omitted.

@goneall see above -- this looks like it might be a bug in the license list publisher?

mirabilos:
It’s indeed optional. At $orkplace for example it’s used [like this](https://evolvis.org/plugins/scmgit/cgi-bin/gitweb.cgi?p=tartools/mvnparent.git;a=blob;f=src/main/ancillary/LICENCE.hdr;h=b44e1329a44f6d24f97a8d7697fc365eb5313867;hb=HEAD)¹, and the Mycila Maven Licence plugin then copies that, with the per-language appropriate comment markers (`/*`, `#`, `%`, `--`, etc.) prefixed, into the source files.

I’d prefer for the optional part to be either fixed or completely removed.

① which is an interesting special case, in that copyright (droit d’auteur) owner and licensor (owner of the exploitation rights) are not identical; I found that listing the authors first and then the licensor works well, we also have larger projects with dozens of authors

goneall:
@swinslow There does appear to be a bug in the HTML/Stylesheets - but I'll need some help from someone more HTML savvy than myself (with is a rather large pool of people ;)

I filed an [issue in the licenseListPublisher](https://github.com/spdx/LicenseListPublisher/issues/57) - if you could take a look and provide me any pointers for a solution, I'll try to get a fix in before the next license list release.

jlovejoy:
Please note that the SPDX matching guidelines, 2.1.4 https://spdx.org/spdx-license-list/matching-guidelines - note that optional (aka omitable) text has to either match the present text or be missing altogether.
Thus, the technically correct solution here, would be to add variable (2.1.3) text markup to the optional text to indicate it can match on both versions. 
Given this whole section is optional, that seems like a lot of work for little gain.

Removing the optional section altogether is therefore tempting, my only hesitation here is that this license was added in this format in the first version of the SPDX License List b/c it's on the OSI-approved list. It feels a bit weird to have our text display differently than their page (I say "feels weird" as I don't have a better way to express that!)
thoughts?

mirabilos:
Jilayne Lovejoy dixit:

>Please note that the SPDX matching guidelines, 2.1.4
>https://spdx.org/spdx-license-list/matching-guidelines - note that
>optional (aka omitable) text has to either match the present text or be
>missing altogether.

OK, then we’re talking about different things.

The (both present and current) “optional” text is a rationale,
much like the FSF’s GPL FAQ, and subject to change.

It is *never* part of the licence as copied into projects.

If I’m reading you right, your “optional” parts are parts of the
licence that may be omitted or may be present, such as the infamous
<optional>with or</optional> without fee of the ISC licence, correct?
If so, the text in question is NOT, and never was, an optional text
in your sense.

>hesitation here is that this license was added in this format in the
>first version of the SPDX License List b/c it's on the OSI-approved
>list.

The OSI person who added it fucked up the formatting, and despite
me requesting them to fix it _multiple times_ nothing ever happened.

bye,
//mirabilos
-- 
I believe no one can invent an algorithm. One just happens to hit upon it
when God enlightens him. Or only God invents algorithms, we merely copy them.
If you don't believe in God, just consider God as Nature if you won't deny
existence.		-- Coywolf Qi Hunt

jlovejoy:
@mirabilos - thanks for the response and additional info, I was kind of wondering if you had reached out to the OSI as well.

given the totality of the above discussion, seems like removing the rationale text makes the most sense to me

other SPDX'ers, can we get a +1 or two on this?
@bradleeedmondson @swinslow @goneall ?

swinslow:
Agree, +1 on removing rationale section.

mirabilos:
Steve Winslow dixit:

>Agree, +1 on removing rationale section. 

OK.

Back to the main text, if you’re really going into automated
matching, asides from different comment markers and varying
formatting of the list of authors/licensors (see the example
I’ve mentioned; also, multiple © or even (c) lines come into
play), there’s two versions out that are identical except one
uses ASCII quotation marks and hyphen-minus instead of the
original UTF-8.


# Provided that these terms and disclaimer and all copyright notices
# are retained or reproduced in an accompanying document, permission
# is granted to deal in this work without restriction, including un‐
# limited rights to use, publicly perform, distribute, sell, modify,
# merge, give away, or sublicence.
#
# This work is provided “AS IS” and WITHOUT WARRANTY of any kind, to
# the utmost extent permitted by applicable law, neither express nor
# implied; without malicious intent or gross negligence. In no event
# may a licensor, author or contributor be held liable for indirect,
# direct, other damage, loss, or other issues arising in any way out
# of dealing in the work, even if advised of the possibility of such
# damage or existence of a defect, except proven that it results out
# of said person’s immediate fault when using the work as intended.

vs.

# Provided that these terms and disclaimer and all copyright notices
# are retained or reproduced in an accompanying document, permission
# is granted to deal in this work without restriction, including un-
# limited rights to use, publicly perform, distribute, sell, modify,
# merge, give away, or sublicence.
#
# This work is provided "AS IS" and WITHOUT WARRANTY of any kind, to
# the utmost extent permitted by applicable law, neither express nor
# implied; without malicious intent or gross negligence. In no event
# may a licensor, author or contributor be held liable for indirect,
# direct, other damage, loss, or other issues arising in any way out
# of dealing in the work, even if advised of the possibility of such
# damage or existence of a defect, except proven that it results out
# of said person's immediate fault when using the work as intended.

(Some might use -, –, or even U+00AD instead of ‐ after “un” in the
third line, I don’t recall all early versions around before I learnt
about all those dashes in Unicode, and others might have miscopied,
too, similar how the correct short name is “MirBSD”, not “MirOS”…)


Make of that what you want.

Thanks,
//mirabilos
-- 
<diogenese> Beware of ritual lest you forget the meaning behind it.
<igli> yeah but it means if you really care about something, don't
    ritualise it, or you will lose it. don't fetishise it, don't
    obsess. or you'll forget why you love it in the first place.

mirabilos:
Jilayne Lovejoy dixit:

>@mirabilos - thanks for the response and additional info, I was kind of
>@wondering if you had reached out to the OSI as well.

I’ve written earlier in this discussion that, yes, I’ve requested
them to fix their page multiple times over the years, including
recently when they started their github repository. So far, nothing
has changed.

Thanks,
//mirabilos
-- 
16:47⎜«mika:#grml» .oO(mira ist einfach gut....)      23:22⎜«mikap:#grml»
mirabilos: und dein bootloader ist geil :)    23:29⎜«mikap:#grml» und ich
finds saugeil dass ich ein bsd zum booten mit grml hab, das muss ich dann
gleich mal auf usb-stick installieren	-- Michael Prokop über MirOS bsd4grml

jlovejoy:
it was easier (for me) to make a new PR to resolve as we decided above (see #942 ) so I'm going to close this one.

PR #912, opened 2019-08-17
Update URLs for bzip2 licenses

BKPepe:
While I was looking for SPDX License Identifier for `bzip2`, I didn't find it for 2 versions (1.0.7 and 1.0.8), which was released recently after [9 years](https://sourceware.org/git/?p=bzip2.git;a=blob;f=CHANGES;h=30afead2586b6d64f50988a41d394a0131b38949;hb=HEAD).

Currently, there is an SPDX License Identifier for version [1.0.5](https://spdx.org/licenses/bzip2-1.0.5.html) and [1.0.6](https://spdx.org/licenses/bzip2-1.0.6.html). The first one is using incorrect URL as the previous owner of bzip.org let it go. More details about it can be found [here](https://lwn.net/Articles/762264/).
As official website they are using their former project home https://sourceware.org/bzip2/ For the other one, the Github repository doesn't seem to be correct, because since version 0.1 till 1.0.6 there was just one developer - [Julian Seward](https://sourceware.org/git/?p=bzip2.git;a=shortlog). Also, in the Github repository there are just two commits and it seems to be abandoned nor official repository for bzip2.

### DIFF
--- a/src/bzip2-1.0.5.xml
+++ b/src/bzip2-1.0.5.xml
@@ -4,6 +4,7 @@
             name="bzip2 and libbzip2 License v1.0.5">
       <crossRefs>
          <crossRef>https://sourceware.org/bzip2/1.0.5/bzip2-manual-1.0.5.html</crossRef>
+          <crossRef>http://bzip.org/1.0.5/bzip2-manual-1.0.5.html</crossRef>
       </crossRefs>
     <text>
       <titleText>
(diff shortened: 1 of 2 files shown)


jlovejoy:
thanks for the research!  We have a policy of not removing old links (even if no longer active) but simply adding the new ones. See: https://github.com/spdx/license-list-XML/blob/master/DOCS/license-fields.md

Can you update this PR along those lines?

We also might consider adding the newer versions, but that would be a separate issue and PR

jlovejoy:
Note for completeness here: the only difference I can see between 1.06 and 1.08 is the copyright year and the email address and dates on the last line, which is not enough to constitute a new license.

There does not seem to be a 1.07 for the license that I can find.

jlovejoy:
added back old urls

jlovejoy:
this is ready to merge once all checks have passed

swinslow:
LGTM, thanks all!

PR #201, opened 2016-03-30
Automated PR for SCEA

myndzi:

PR #915, opened 2019-08-19
LAL-1.3: Minor adjustments, mainly spacing

lkeane:
The formatting of LAL-1.3 in its current form seems to suffer from some odd spacing and line breaks. Maybe this spacing is cleaner?

### DIFF
--- a/src/LAL-1.3.xml
+++ b/src/LAL-1.3.xml
@@ -133,11 +133,9 @@
               originaux (initiaux et/ou conséquents).
             </item>
           </list>
-          <p>
-            Les auteurs des originaux pourront, s’ils le souhaitent, vous 
-            autoriser à diffuser l’original dans les mêmes conditions que les 
-            copies.
-          </p>
+          Les auteurs des originaux pourront, s’ils le souhaitent, vous 
+          autoriser à diffuser l’original dans les mêmes conditions que les 
+          copies.
         </item>
         <item>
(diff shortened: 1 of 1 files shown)


mlinksva:
It appears the tests (CI) don't like NNBSP. I'm not sure whether that's a bug in the tests, or if it makes sense to revert 672eef6. Probably others like @goneall will have an opinion or mandate. 😄

zvr:
I would *strongly* suggest we leave spaces as simple ASCII ones.

lkeane:
I just replaced NNBSP with NBSP, and it is now passing checks. NBSP is within Extended ASCII, being character number 255. This xml, LAL's canonical version being French, already uses a number of characters from there such as é (no. 130), à (no. 133), &c. so I am assuming NBSP is acceptable.

goneall:
Agree with @zvr .  It would be a nice enhancement for the compare tools as well.  I submitted [issue #201](https://github.com/spdx/tools/issues/201) on the SPDX tools to track.  Feel free to comment on that issue if you disagree with the change.

jlovejoy:
@goneall @zvr - where does that leave us with this PR then?

goneall:
@jlovejoy The PR is fine as far as I'm concerned.  

I fixed issue 201 on the SPDX tools, so the changes to the spacing will not be necessary in the future.  Keeping the changes in the PR isn't a problem, however.

Issue #894, opened 2019-07-03
New license request: UCL-1.0 [SPDX-Online-Tools]


jlovejoy:
**1.** License Name: Upstream Compatibility License v1.0
**2.** Short identifier: UCL-1.0
**3.** License Author or steward: Larry Rosen & Nigel Tzeng
**4.** Comments: This license was approved by OSI in 2017 and somehow we missed it.
**5.** Standard License Header: Licensed under the Upstream Compatibility License 1.0
**6.** URL: https://opensource.org/licenses/UCL-1.0
**7.** OSI Status: Approved

jlovejoy:
+1 to add - it's already OSI approved, we should have added it when it was approved by OSI, but missed it.

swinslow:
This looks quite similar to the OSL-3.0 license -- reviewing to see whether the differences are substantive for SPDX matching guidelines purposes...

jlovejoy:
difference is only in 1(c) - appears this was discussed on OSI mailing list (from someone on call looking back at those logs)

swinslow:
Thanks @jlovejoy -- I'm +1 to add also, given that it's OSI approved and I agree the differences are sufficiently substantive to call it a different license from OSL-3.0 for SPDX purposes.

pombredanne:
FWIW, this has been tracked in ScanCode for a while, but in hindsight there is no use in the wild I could find except for that one small project https://github.com/ftc-team-11553/ftc_2017 

All other references are to license tools (and several are to clones of scancode).
So this does not seem to be used much anywhere. @jlovejoy do you have some pointers to actual usages by chance?

Note that in any case since this is an officially OSI approved license, adding it even if not much used probably makes sense.

swinslow:
@pombredanne I'm not seeing any usage in the wild other than that. But agreed, I think we've generally viewed that all OSI-approved licenses are appropriate to include on the list regardless of usage in the wild.

swinslow:
Discussed on 2019-08-08 legal team call, agreed that this should be added.

johnmhoran:
I'd be happy to work on this.

swinslow:
Fantastic, thanks @johnmhoran!

johnmhoran:
Hi @swinslow .  I'm belatedly tackling this issue, have just completed my initial draft of `UCL-1.0.xml`, and have one item to draw your attention to:

[...]

PR #922, opened 2019-09-04
Add UCL-1.0.xml #894

johnmhoran:
* Used OSL-3.0.xml as a template.

Signed-off-by: John M. Horan <johnmhoran@gmail.com>


swinslow:
This is for issue #894

swinslow:
@johnmhoran I made a couple of tweaks, to add the license list version where this will first be added (3.7) and to make the angle-bracketed text optional.

On the latter point, that's because the instances where we've seen this license used (https://github.com/ftc-team-11553/ftc_2017/blob/master/LICENSE.md and the OSI website), that language has been absent -- but in the [original submission to OSI](http://lists.opensource.org/pipermail/license-review_lists.opensource.org/2017-February/002980.html), it was present. Probably easiest for SPDX purposes to mark that bit as `<optional>` so that either occurrence would match. Does that work for you?

johnmhoran:
Thanks @swinslow .  Very cool that you found Nigel Tzeng's original submission.  From a drafting/interpretation point of view the missing language (or equivalent) seems mandatory, not optional -- but if our primary goal is matching incoming text, `<optional>` seems the way to go.

swinslow:
Discussed on 2019-09-05 call, ready to merge with optional text included; @jlovejoy to flag for OSI

PR #930, opened 2019-09-25
CERN-OHL-1.2: Align on American English "License" in the name

sschuberth:


goneall:
From a license matching perspective, the change isn't required since the [license matching guidelines](https://spdx.org/spdx-license-list/matching-guidelines) section 8.1.1 row 22 specify License and Licence as equivalent. 

I am leaning towards leaving the original Licence since the license steward uses that spelling (see https://ohwr.org/cernohl).  However, I'll defer to @jlovejoy on this.

sschuberth:
For comparison, version 1.1 of the same license uses "License" already, see https://github.com/spdx/license-list-XML/blob/d85658123a501c33b9310779a3f112aedc395d23/src/CERN-OHL-1.1.xml#L2.

I stumbled upon this because we're using the `name` field to auto-generate mappings from human readable names to SPDX ids in ORT, also see https://github.com/spdx/package-licenses-mapping.

swinslow:
I agree with @goneall that (1) for SPDX matching guidelines purposes the two are considered equivalent; and (2) in the titles, we should defer where possible to the license steward's choice of spelling.

As @sschuberth noted though, there's a difference in the spellings on the SPDX license list between [1.1](https://spdx.org/licenses/CERN-OHL-1.1.html) and [1.2](https://spdx.org/licenses/CERN-OHL-1.2.html). It looks to me like this was inadvertent, and that 1.1 on the license list is spelling it incorrectly.

So I'd suggest that instead we fix 1.1 by reverting it to "Licence" since that's what the steward does in the text itself. I'll open a separate issue for that, so that it's tracked in the issues.

sschuberth:
I'm fine with fixing it either way, as long as it's consistent. I would have appreciated, however, if @swinslow would have filed the PR with the alternate fix right away, instead of filing an issue, which in this case is about the same amount of work.

swinslow:
Yes, understood. Main reason I did is because there are some newer participants in submitting PRs to the license list and this was a "good first issue" type of fix for them to do. If it isn't picked up during the legal team call today then I'll just make the change myself prior to the upcoming 3.7 release.

Issue #927, opened 2019-09-06
Proposal: Add a machine readable form of equivalent words


goneall:
The license matching guidelines list a number of words which are equivalent in the documentation.  Adding this in a machine readable form in the license-XML repository would allow for tools to pick up the latest and greatest with minimal effort.

goneall:
There are several alternatives to the format and naming for the file.

To keep it simple, yet well defined, I would propose:
- A file named "equivalentwords.txt" at the root of the repo
- The file be a UTF-8 encoded file
- The format of the file is the 2 equivalent words separated by a comma on separate lines
- Line endings are single LF characters (Linux style)

For example:
```
acknowledgment,acknowledgement
analogue,analog
analyse,analyze
```

swinslow:
@goneall Makes sense to me! I'll resist the urge to over-engineer this, I think the simple text file approach you described sounds good  :)

swinslow:
This seemed straightforward so I went ahead and did it  :)  @goneall, let me know what you think.

This was based on the list of words from the current matching guidelines at https://spdx.org/spdx-license-list/matching-guidelines.

A couple of notes to add to the format details listed above:
* these have all been converted to lowercase
* they are _not_ fully alphabetized, instead I've maintained the ordering from the matching guidelines page
* note that there is some overlap in terms: -- separate lines for `sublicense,sub-license` and `sub-license,sub license`

For the latter point, I wonder if these should be grouped into a single line?

`sublicense,sub-license,sub license`

Seems clearer to me, but it would mean that there could be >1 comma per line.

swinslow:
@goneall Do you want to give #929 a quick review? If you approve then I can go ahead and merge it, if in the next couple days then can include it for the 3.7 release.

goneall:
> For the latter point, I wonder if these should be grouped into a single line?

Restricting it to 2 per line makes it a little easier to program to (e.g. can use a simple map), but it is a bit harder to read/maintain by humans.

I'm leaning to to 2 per line but I don't feel strongly about it.

zvr:
I also have no preference, as long as we are all clear in the semantics:

Since we're talking "equivalence",

- the pairs X,Y and Y,Z also cover X,Z
- the same way X,Y also covers Y,X.

Actually, @goneall, I'd think that having all equivalent forms in one line would be easier to program, since there would be no need to check for pre-existing equivalents...

Do we plan to add quotes and hyphens?

goneall:
[...]

PR #941, opened 2019-10-16
Update matching guidelines to allow HTTP:// and HTTPS:// to match

goneall:
The wording on this can likely be improved - please feel free to recommend changes.


swinslow:
@goneall I'm in favor of this, and fine with the wording. My only comment is that if it gets a thumbs-up, we'll want to update the website and then merge it here (since the DOCS/matching-guidelines file is currently still just a mirror of the website).

If this gets added then we can bump the guidelines version number to v2.1, together with the equivalent-words change in #927 / #969.

@zvr @bradleeedmondson or others -- any concerns with @goneall's proposal here?

swinslow:
Discussed on 2020-01-30 legal team call, fine with adding this. @goneall let's sync on when I should update the SPDX website version of the matching guidelines.

swinslow:
The website is now updated (see https://spdx.org/spdx-license-list/matching-guidelines) and the local copy was merged in #974. I will close this PR, but thank you @goneall for providing the content!

PR #949, opened 2019-11-06
Parity 7.0.0

kemitchell:
Due to extensive feedback, and with the involvement of several users, Parity was recently rewritten more in the style of Blue Oak and the Polyform licenses. There are also substantial substantive improvements.

The validation script only started failing when I put the numbered items in `<list>`, rather than `<p>`. Not sure what's going on there. Maybe a typo?


goneall:
It looks like it is failing when comparing to the text file:

```
Test for license Parity-7.0.0 failed: Test for license ID Parity-7.0.0 failed due to difference found Unable to find the text ' , ' following a variable rule 'blueOakRef' starting at line #71 column #225 "".  Last optional text was not found due to the optional difference: 
	Normal text of license does not match starting at line #9 column #0 "this" when comparing to template text " ##"
```

It may be due to the changes in the optional text.  The matching algorithm works on tokens and it may be tripping up on the single character optional.

kemitchell:
I’ll try replacing the word and its markup with an alt that allows either.

kemitchell:
Fiddled until it passed. Glad I wrote that script.

zkat:
Very excited to see this happen, as I have several projects licensed under Parity 7 already

kemitchell:
@swinslow any chance this new version was discussed today, as well?

swinslow:
@kemitchell It wasn't, but I think that is primarily because (for better or worse) on most legal team calls we end up working primarily off of the issues list, rather than the open PRs. 

Is there an issue open for Parity 7.0.0? If not, then if you wouldn't mind opening one and tagging me in it, I'll make sure we at least touch on it during the next call. It's very helpful that you've got an XML file that's passing the tests all ready to go  :)

swinslow:
@kemitchell this looks good to me! My only edit was to change the license list version to 3.9, as that's what we're on now.

Since Parity-7.0.0 has been accepted (per notes in #987) I'll go ahead and merge this as soon as the tests finish re-running. Thanks!

PR #952, opened 2019-12-04
PSF-2.0: add new license

### LINKED ISSUE #919: Python-2.0 license should not include all the history of Python licensing
The Python-2.0 license is used in Python proper and in other projects. Yet the Python-2.0 text @ SPDX is really the whole Python.org license notice and history, making this is license text useless outside of the Python.org licensing.

rohieb:
Closes #919

Example for the test was taken directly from
https://github.com/python/cpython/blob/2.7/LICENSE#L62

### DIFF
--- a/src/PSF-2.0.xml
+++ b/src/PSF-2.0.xml
@@ -1,0 +1,77 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<SPDXLicenseCollection xmlns="http://www.spdx.org/license">
+  <license licenseId="PSF-2.0" name="Python Software Foundation License 2.0">
+    <crossRefs>
+      <crossRef>https://opensource.org/licenses/Python-2.0</crossRef>
+    </crossRefs>
+    <notes>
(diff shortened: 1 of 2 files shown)


bradleeedmondson:
This approach makes sense to me.

src/PSF-2.0.xml
```diff
@@ -0,0 +1,77 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<SPDXLicenseCollection xmlns="http://www.spdx.org/license">
+  <license licenseId="PSF-2.0" name="Python Software Foundation License 2.0">
```
swinslow: In the `license` tag, can you also add another attribute: `listVersionAdded="3.8"`

swinslow:
Hi @rohieb -- we discussed on the legal team call, I'll go ahead and merge and will then add the `listVersionAdded` property in a subsequent commit. Thanks again for preparing this!

PR #962, opened 2019-12-31
MIT: Expand markup for Software definition

### LINKED ISSUE #866: add markup to MIT

swinslow:
This adds markup to the specific language used to define "Software",
enabling MIT to match when used in cases where e.g. "this source file"
is used instead of "this software and associated documentation files".

Fixes #866 

### DIFF
--- a/src/MIT.xml
+++ b/src/MIT.xml
@@ -13,8 +13,8 @@
          </p>
       </copyrightText>
 
-      <p>Permission is hereby granted, free of charge, to any person obtaining a copy of this software and
-         associated documentation files (the "Software"), to deal in the Software without restriction,
+      <p>Permission is hereby granted, free of charge, to any person obtaining a copy of <alt match=".+" name="software">this software and
+         associated documentation files</alt> (the "Software"), to deal in the Software without restriction,
          including without limitation the rights to use, copy, modify, merge, publish, distribute, sublicense,
          and/or sell copies of the Software, and to permit persons to whom the Software is furnished to do so,
          subject to the following conditions:</p>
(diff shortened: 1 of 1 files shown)


swinslow:
@jlovejoy This is to address the issue you flagged in #866 -- assuming the tests pass, can you take a look and approve if you're good with it?

swinslow:
Per legal team call 2020-01-30, I'll be updating this PR to instead allow a choice just between original language and "this source file" in the regex, and will then merge.

jlovejoy:
@swinslow - is this ready for review? seems like a minor update, no?

swinslow:
Hi @jlovejoy, this was discussed on some of the calls in January and there was discomfort about adding a blanket "match everything" option. See https://github.com/spdx/license-list-XML/issues/866#issuecomment-570285502 and below for the discussion.

The outcome was that for now, we would switch it so that the regex would match either the original language or "this source file". Let me see if I can update the PR right now to fix that, and then we can get it in for 3.9.

swinslow:
@jlovejoy This should be good to go now  :)  Take a look and feel free to merge, or if you don't have any concerns then I'll merge it before doing the release. Thanks!

jlovejoy:
Fixes #866

Issue #840, opened 2019-03-28
Apache-1.1 template markup


reversi-fun:
**1.** License Name: CA Open Source Licence Version 1.0
**2.** Short identifier: CAOSL-1.0
**3.** License Author or steward: Computer Associates.
**4.** Comments: This License likes (("Apache-1.1" + Entessa + "calculate-Linux/JNIC")/3).

The license holder seems to be stuck with the difference, AS there:
https://gitweb.gentoo.org/repo/gentoo.git/commit/?id=15abaf9459ca5f92b77c048f578c609f1f7d34e6
**5.** URL: https://www.calculate-linux.org/packages/licenses/CAOSL
**6.** OSI Status: Not Submitted

reversi-fun:
I add evidence that I have analyzed similarity.

https://github.com/reversi-fun/license_doc_similality1/blob/master/data/lic_graph.sfdp.svg
![CAOSL-license](https://user-images.githubusercontent.com/2231549/55133888-86b67b00-516a-11e9-9404-ebc791c08a7c.png)

pombredanne:
This is a variant of Apache-1.1 and IMHO not worthy of a new license id.

reversi-fun:
IMHO?
1. CAOSL-1.0 has less disclaimer than Apache-1.1.
  CAOSL-1.0 does not indicate the existence of any other copyright holder and the possibility of betrayal.
  Apache-1.1 sayes "This software consists of voluntary contributions made ....". 
  It suggests a legal risk.

2. The prohibition conditions in paragraph 5 are different.
  CAOSL-1.0 prohibits only the name of the copyright holder.
  Apache-1.1 prohibits  some names(copyright holder and product name trademark?).
 for exsample:
  "my-apache-commmon" is drived from CAOSL-1.1 licensed product.
   It is not a violation in CAOSL-1.1. 
   but It's a violation in Apache-1.1 .

jlovejoy:
it is a match to Apache-1.1 with a couple minor variations:
* clause 4 removes the last sentence, "For written permission..."
* clause 5 states "Products may not include <NAME> ..." as opposed to "Products derived from this software may not be called <NAME>..." 

I don't think either of these differences as being substantively different from a legal point of viewand could be accommodated with additional markup to Apache-1.1 - @swinslow @bradleeedmondson - do you agree?

swinslow:
[...]

PR #963, opened 2020-01-07
Apache 1.1 markup

goneall:
Resolves issue #840


swinslow:
Hi @goneall, the changes to Apache-1.1.xml look great to me. 

But it looks like this PR also includes a commit to modify the matching guidelines, which doesn't look to me like it was meant to be part of addressing this issue (unless I'm misunderstanding?)  Is it possible to remove that commit, and just merge the Apache-1.1.xml changes?

goneall:
@swinslow I'm not sure how that got into the PR - I'll update it later today after the tech call.

goneall:
@swinslow Removed the changes to the matching guidelines.  Going to go ahead and merge once the tests pass.

src/Apache-1.1.xml
```diff
@@ -43,18 +43,14 @@
           <alt match="name\(s\)|name|name" name="nameClause4">name</alt>
             <alt match=".+" name="organizationClause4">"Apache" and "Apache Software Foundation"</alt> must
           not be used to endorse or promote products derived from this software without prior written
-          permission. For written permission, please contact
-          <alt match=".+" name="contactClause4">apache@apache.org</alt>.
```
agr, line 47: Why was the markup around the email removed?

goneall: @agr The PR was done a while go, so I don't recall the reason.  Likely changing the entire sentence to optional seemed to obviate the need for the alternative clause around the contact.

There is no harm, however, in including the alternative tag as long as the optional tag is retained.

If removing this is causing an issue, feel free to create a PR adding it back.  If it causes an issue in the tests, let me know.

PR #965, opened 2020-01-16
Add separate -RFN and -no-RFN identifiers for OFL licenses

### LINKED ISSUE #724: New License Request: Reserved Font Name options for OFL-1.0 and OFL-1.1 [EDITED]
This request formalizes the inquiry discussed on the mailing list here: https://lists.spdx.org/g/Spdx-legal/message/2283

Proposed Full Name: Reserved Font Name

Proposed Short Identifier: RFN

URL Reference: https://scripts.sil.org/cms/scripts/page.php?item_id=OFL_web
(Additional exploration of the RFN clause by the license authors is available at https://scripts.sil.org/cms/scripts/page.php?site_id=nrsi&id=OFL_web_fonts_and_RFNs )

OSI approval: Yes. The RFN clause is an option for both the 1.0 and 1.1 (current) versions of the Open Font License (OFL), which is OSI-approved.

Explanation: 
The RFN clause is an optional condition of the Open Font License (OFL) that the copyright holder can invoke when publishing a font under the OFL. By default, the RFN clause is inactive. If invoked on a font package, it substantially alters the work required for recipients to comply with the OFL. 

The second URL reference above details the license author/custodian's intent regarding which changes to a font binary constitute a "modification" that triggers the clause.

Since there is currently no distinction between "vanilla OFL" and "OFL with an RFN declaration" in SPDX or other license-compliance tooling, remaining in compliance is made more difficult for downstream redistributors of OFL-with-RFN fonts.

Text: 
' "Reserved Font Name" refers to any names specified as such after the
copyright statement(s).'
[...]
'3) No Modified Version of the Font Software may use the Reserved Font
Name(s) unless explicit written permission is granted by the corresponding
Copyright Holder. This restriction only applies to the primary font name as
presented to the users.'

swinslow:
As discussed in #724, OFL-1.0 and OFL-1.1 have options permit a licensor
to designate OFL-licensed fonts as having a "Reserved Font Name." If so
designated, then certain provisions of the license are active.

The existing SPDX entries for OFL-1.0 and OFL-1.1 do not distinguish
between these options, so current use of "OFL-1.0" or "OFL-1.1" cannot
be assumed to assert anything about whether a Reserved Font Name is,
or is not, present.

Additionally, using "WITH Reserved-Font-Name" as an exceptions list
entry does not seem appropriate, because the Reserved Font Name is not
really an "exception" -- rather, it imposes additional obligations if
it is present.

This commit adds new license list entries for:
*  OFL-1.0-RFN
*  OFL-1.0-no-RFN
*  OFL-1.1-RFN
*  OFL-1.1-no-RFN

to enable optionally designating whether a Reserved Font Name applies.
The existing OFL-1.0 and OFL-1.1 identifiers, without a suffix, continue
to have their presumed existing meaning that they should not be read
as asserting whether or not a Reserved Font Name is present, and this is
clarified in the Notes sections for those licenses.

Finally, because the license text is identical for these variants, they
have been added to the expected-warnings file to suppress the build
error to that effect.

Fixes #724


bradleeedmondson:
This approach makes sense to me, and I think it makes sense to proceed. Before merging though, something in the build test seems to have failed. Maybe related to validating canonical license text(s)?

https://travis-ci.org/spdx/license-list-XML/jobs/638008380#L248

goneall:
It looks lie there is a duplicate license match that isn't in the ignore list: 

`Duplicates licenses: OFL-1.1, OFL-1.1-RFN`

swinslow:
Thanks @bradleeedmondson @goneall -- my mistake, I thought I'd added all the combos, since it passed on my system  :)  Will push another commit to fix this!

swinslow:
...I think it's possible that the duplicates checker might be generating the duplicates in different orders, on different systems -- and then failing when they aren't listed in the exact order in the expected-warnings file.

The prior commit included `"Duplicates licenses: OFL-1.1-RFN, OFL-1.1"` -- see https://github.com/spdx/license-list-XML/pull/965/commits/27856c54863b2b9d074e5388d36e6976e61e2c4d#diff-d00b85a566a4344b53d0a41c7ef266cfR1

which I am virtually certain was the ordering when I ran this before submitting the commit... but now it is failing on Travis with: `Duplicates licenses: OFL-1.1, OFL-1.1-RFN` -- the same, but flipped.

I'll add the reverse version, to get this passing -- but @goneall I don't know if this is expected behavior, or some fluke?

swinslow:
Ah -- now that I'm actually reading it, I see that the other licenses in the expected-warnings file are also included in both orders. My mistake!  :)  I'll do the same for each of the OFL entries.

swinslow:
Looks like it's passing now -- let me know any concerns, otherwise i'll merge. Thanks!

goneall:
> I see that the other licenses in the expected-warnings file are also included in both orders. 

Yeah - the tool is very simple and just filters out a string of warnings, so you need to do both directions when adding expected duplicates.  It would be nice to enhance to tool so you only need to do one direction, but it will probably be a while before I get a chance to update it.

swinslow:
Thanks all -- given the OK from @bradleeedmondson above, I'll go ahead and merge this.

PR #968, opened 2020-01-30
Add GPL-3.0-linking-exception and GPL-3.0-linking-source-exception

### LINKED ISSUE #939: Generic GPL linking exceptions (OpenSSL related)
As I gather from #720 and #824, there are already a few OpenSSL exceptions out in the wild. To add to that, I was looking at wget, which is under GPL-3.0-or-later, and has an exception for linking against OpenSSL in their source files (e.g. [src/http.c]):

```
Additional permission under GNU GPL version 3 section 7

If you modify this program, or any covered work, by linking or
combining it with the OpenSSL project's OpenSSL library (or a
modified version of that library), containing parts covered by the
terms of the OpenSSL or SSLeay licenses, the Free Software Foundation
grants you additional permission to convey the resulting work.
```

This seems to be directly copied from the template which the FSF provides in its [FAQ]:

```
Additional permission under GNU GPL version 3 section 7

If you modify this Program, or any covered work, by linking or combining it 
with [name of library] (or a modified version of that library), containing 
parts covered by the terms of [name of library's license], the licensors of 
this Program grant you additional permission to convey the resulting work. 
{Corresponding Source for a non-source form of such a combination shall include 
the source code for the parts of [name of library] used as well as that of the 
covered work.}
```

I have so far only found this template in wget, substituted for OpenSSL. The FSF template is very generic so I'm not sure if it makes any sense to include it in SPDX with the current template variables, especially since (I think) there is no way to express template variables in SPDX License Identifiers. So after all we would only be able to express that a software is licensed under GPL-3.0-or-later with an exception to link it against some unspecified library.

The best alternative I can think of is to have one version for each library (e.g. OpenSSL), and add them to SPDX as they are found in the real world. However, this would probably yield a lot of `FSF-GPL-3.0-$LIBRARY-Exception` templates in the future…

[src/http.c]: https://github.com/mirror/wget/blob/master/src/http.c#L20
[FAQ]: https://www.gnu.org/licenses/gpl-faq.en.html#GPLIncompatibleLibs

swinslow:
This adds two variants of the recommended template exception from
the FSF's FAQ:
https://www.gnu.org/licenses/gpl-faq.en.html#GPLIncompatibleLibs

The only difference is that the "-source" variant incorporates the
second optional sentence from the FAQ template.

Note that in the identifier, "3.0" is intentionally included after
"GPL" rather than at the end of the identifier. That's because the
exception itself is not versioned, and this is meant to communicate
that the exceptions are intended for use with GPL-3.0.

Fixes #939

### DIFF
--- a/src/exceptions/GPL-3.0-linking-exception.xml
+++ b/src/exceptions/GPL-3.0-linking-exception.xml
@@ -1,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<SPDXLicenseCollection xmlns="http://www.spdx.org/license">
+  <exception licenseId="GPL-3.0-linking-exception" name="GPL-3.0 Linking Exception" listVersionAdded="3.8">
+    <crossRefs>
+      <crossRef>https://www.gnu.org/licenses/gpl-faq.en.html#GPLIncompatibleLibs</crossRef>
+    </crossRefs>
+    <notes>
+      This exception is based on the suggested template from the Free Software Foundation's FAQ about the GPL. This variant does not include the second optional sentence regarding Corresponding Source. For a variant with that sentence, please see GPL-3.0-linking-source-exception.
+    </notes>
+    <text>
+      <titleText>
+	<optional>
+	<p>
+	  Additional permission under GNU GPL version 3 section 7
+	</p>
+	</optional>
+      </titleText>
+      <p>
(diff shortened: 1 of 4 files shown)


swinslow:
Discussed on 2020-01-30 legal team call, OK to merge.

PR #981, opened 2020-02-27
updates to README

jlovejoy:
add more info on contributing to readme: add "how we work" section, update links

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,15 @@
 [![Build Status](https://api.travis-ci.org/spdx/license-list-XML.svg?branch=master)](https://travis-ci.org/spdx/license-list-XML)
 # SPDX License List
+##hat
 The [SPDX License List](https://spdx.org/licenses/) is a list of commonly found licenses and exceptions used in free and open source and other collaborative software or documentation. The purpose of the SPDX License List is to enable easy and efficient identification of such licenses and exceptions in an SPDX document, in source files or elsewhere. The SPDX License List includes a standardized short identifier, full name, vetted license text including matching guidelines markup as appropriate, and a canonical permanent URL for each license and exception.
+
+## Why
+The purpose of the SPDX License List is to enable efficient and reliable identification of such licenses and exceptions in an SPDX document, in source files or elsewhere. The SPDX short identifiers combined with the matching guidelines ensures that anyone can reliably know exactly what license text is being referred to for a given SPDX identifier.
+
+## How
+* For more about how SPDX license identifiers are used in an SPDX document, a software bill of materials, or other places that store license data, see [SPDX Specification](https://spdx.org/specifications), Sections 3, 4, and 6 and Appendices II, IV, and V.
+* For examples of use of SPDX license identifiers in source code, see https://spdx.org/ids.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,12 +1,28 @@
...
 
+## Why
+The purpose of the SPDX License List is to enable efficient and reliable identification of such licenses and exceptions in an SPDX document, in source files or elsewhere. The SPDX short identifiers combined with the matching guidelines ensures that anyone can reliably know exactly what license text is being referred to for a given SPDX identifier.
+
+## How
+* For more about how SPDX license identifiers are used in an SPDX document, a software bill of materials, or other places that store license data, see [SPDX Specification](https://spdx.org/specifications), Sections 3, 4, and 6 and Appendices II, IV, and V.
+* For examples of use of SPDX license identifiers in source code, see https://spdx.org/ids.
+Please note, a license not on the SPDX License List can be included in an SPDX document by using a License Ref and including the full license text as per the specification. If you think there is a license that should be on the SPDX License List, please follow the instructions [CONTRIBUTING](https://github.com/spdx/license-list-XML/blob/master/CONTRIBUTING.md)
+
```
jlovejoy: I think an "it's not just a list!" section here with links to other pertinent aspects would also be good. 

thoughts?

jlovejoy: i.e. move stuff at top of CONTRIBUTING (roughly, the "Before you contribute..."  section) here

README.md
```diff
@@ -1,12 +1,28 @@
 [![Build Status](https://api.travis-ci.org/spdx/license-list-XML.svg?branch=master)](https://travis-ci.org/spdx/license-list-XML)
 # SPDX License List
+##hat
```
swinslow: I think a space and a W got eaten here. Unless this is the hat section  :)

README.md
swinslow: This is great! and yes, I think moving that top part of the CONTRIBUTING file to the README would make sense.

README.md
```diff
@@ -1,12 +1,28 @@
...
 
+## Why
+The purpose of the SPDX License List is to enable efficient and reliable identification of such licenses and exceptions in an SPDX document, in source files or elsewhere. The SPDX short identifiers combined with the matching guidelines ensures that anyone can reliably know exactly what license text is being referred to for a given SPDX identifier.
+
+## How
+* For more about how SPDX license identifiers are used in an SPDX document, a software bill of materials, or other places that store license data, see [SPDX Specification](https://spdx.org/specifications), Sections 3, 4, and 6 and Appendices II, IV, and V.
+* For examples of use of SPDX license identifiers in source code, see https://spdx.org/ids.
+Please note, a license not on the SPDX License List can be included in an SPDX document by using a License Ref and including the full license text as per the specification. If you think there is a license that should be on the SPDX License List, please follow the instructions [CONTRIBUTING](https://github.com/spdx/license-list-XML/blob/master/CONTRIBUTING.md)
```
swinslow: Minor nit, instead of "License Ref" I would change it to `LicenseRef-` -- that's LicenseRef- in backticks and with a trailing hyphen. That way it'll mirror the actual formatting used in a custom license ID.

README.md
jlovejoy: I think I've fixed this...

jlovejoy: I think it might be good to merge this and then have another look and add this, as I've lost track of where I am just a bit...

PR #983, opened 2020-02-27
Delete what-why-how.md

jlovejoy:
I forgot about this!  makes more sense to put front and center in readme, since first paragraph is repeated anyway


swinslow:
Makes sense, I'll hold on merging this until we also merge #981 so that the content remains visible in the meantime.

PR #985, opened 2020-02-27
update license-inclusion principles

jlovejoy:
opening paragraphs edits and move to top

### DIFF
--- a/DOCS/license-inclusion-principles.md
+++ b/DOCS/license-inclusion-principles.md
@@ -6,8 +6,8 @@ Determining whether a candidate license should be included on the SPDX License L
 **Definitive factors:**
 * The submitted license must not match another license already on the SPDX License List as per the SPDX matching guidelines.
 * All OSI-approved licenses will be included on the SPDX License List.
-  * This is consistent with early collaboration between SPDX and OSI (2011) to ensure that all OSI-approved licenses were represented on the SPDX License List. Also att that time, OSI began using the SPDX license identifiers in their URLs and on the license pages.
-* Binary software licenses or licenses that do not provide for the avilabilty of the source code will not be included on the SPDX License List.
+  * This is consistent with early collaboration between SPDX and OSI (2011) to ensure that all OSI-approved licenses were represented on the SPDX License List. Also at that time, OSI began using the SPDX license identifiers in their URLs and on the license pages.
+* Binary-only software licenses or licenses that do not provide for the availability of the source code will not be included on the SPDX License List.
   
 **Other factors (roughly in order of importance):**
 * The license substantially complies with one of the following open source definitions (even if not submitted for approval, as in the case of OSI) or these organization consider the license to meet their definition:
@@ -15,8 +15,8 @@ Determining whether a candidate license should be included on the SPDX License L
   * [Free Software Definition](https://www.gnu.org/philosophy/free-sw.en.html) from the Free Software Foundation (FSF)
   * [Open Source Hardware Definition](https://www.oshwa.org/definition/) from the Open Source Hardware Association (OSHWA)
   * [Open Definition](http://opendefinition.org/od/2.1/en/) from the Open Knowledge Foundation (for open data)
-* The license is structured to be generally usabe by anyone. It is not a "vanity" license or specific to one project, consortium or corporation. 
-* The license has actual, substantial use such that it is likely to be encountered. Substantial use may be demonstreated via use in many projects, or in one or a few highly significant projects. 
+* The license is structured to be generally usable by anyone. It is not a "vanity" license or specific to one project, consortium or corporation.
+* The license has actual, substantial use such that it is likely to be encountered. Substantial use may be demonstrated via use in many projects, or in one or a few highly significant projects.
 * The license has identifiable and stable text; it is not in the midst of drafting. The license steward is committed not modifying after addition to the list and to versioning new versions in the future.
 * If the license does not substantively comply with one of the above open source definitions, the license requires the availability of the source code with limited restrictions and meets other factors listed here.
 * The license steward supports or is at least aware of and does not oppose its submission to the SPDX License List.
--- a/DOCS/license-inclusion-principles.md
+++ b/DOCS/license-inclusion-principles.md
@@ -18,7 +18,7 @@ Determining whether a candidate license should be included on the SPDX License L
 * The license is structured to be generally usable by anyone. It is not a "vanity" license or specific to one project, consortium or corporation.
 * The license has actual, substantial use such that it is likely to be encountered. Substantial use may be demonstrated via use in many projects, or in one or a few highly significant projects.
 * The license has identifiable and stable text; it is not in the midst of drafting. The license steward is committed not modifying after addition to the list and to versioning new versions in the future.
-* If the license does not substantively comply with one of the above open source definitions, the license requires the availability of the source code with limited restrictions and meets other factors listed here.
+* If the license does not substantively comply with one of the above open source definitions, then the license is primarily intended for free distribution of content (including, in the case of software, its source code) with limited restrictions, and also meets other factors listed here.
 * The license steward supports or is at least aware of and does not oppose its submission to the SPDX License List.
 
 The SPDX legal team will also consider any information considered relevant by the license steward or open source community.
--- a/DOCS/license-inclusion-principles.md
+++ b/DOCS/license-inclusion-principles.md
@@ -3,13 +3,13 @@ Determining whether a candidate license should be included on the SPDX License L
 
 ## Candidate License Analysis
 
-Definitive factors:
+**Definitive factors:**
 * The submitted license must not match another license already on the SPDX License List as per the SPDX matching guidelines.
 * All OSI-approved licenses will be included on the SPDX License List.
   * This is consistent with early collaboration between SPDX and OSI (2011) to ensure that all OSI-approved licenses were represented on the SPDX License List. Also att that time, OSI began using the SPDX license identifiers in their URLs and on the license pages.
 * Binary licenses or licenses that do not provide for the avilabilty of the source code will not be included on the SPDX License List.
   
-Other factors (roughly in order of importance):
+**Other factors (roughly in order of importance):**
 * The license substantially complies with one of the following open source definitions (even if not submitted for approval, as in the case of OSI) or these organization consider the license to meet their definition:
   * [Open Source Definition (OSD)](https://opensource.org/osd) from the Open Source Initiative (OSI)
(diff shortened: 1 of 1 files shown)


DOCS/license-inclusion-principles.md
```diff
@@ -1,24 +1,32 @@
...
-Comments on license inclusion will be noted in the Github issue for the license (or reference to meeting minutes, as necessary). Some license requests may be decided solely via comments in the Github issue and some may involve discussion on the bi-weekly legal call. Thus it is recommended that requestors join the mailing list and calls to fully participate.
-
-
-## Backstory
+**Definitive factors:**
+* The submitted license must not match another license already on the SPDX License List as per the SPDX matching guidelines.
+* All OSI-approved licenses will be included on the SPDX License List.
+  * This is consistent with early collaboration between SPDX and OSI (2011) to ensure that all OSI-approved licenses were represented on the SPDX License List. Also att that time, OSI began using the SPDX license identifiers in their URLs and on the license pages.
+* Binary licenses or licenses that do not provide for the avilabilty of the source code will not be included on the SPDX License List.
+
```
jlovejoy: do we want to draw a line in the sand on binary licenses? if we are opening up to "source available" then making this restriction seems reasonable. it has effectively been the reality we've operated under all along

kestewart: Please no.  What about data licenses?    I can see cases where folks will want to capture information in SPDX documents - especially training data sets for machine learning, etc.   We want to make it easy for folk to share this sort of info,  not raise barriers.

DOCS/license-inclusion-principles.md
```diff
@@ -1,24 +1,33 @@
...
+  * [Free Software Definition](https://www.gnu.org/philosophy/free-sw.en.html) from the Free Software Foundation (FSF)
+  * [Open Source Hardware Definition](https://www.oshwa.org/definition/) from the Open Source Hardware Association (OSHWA)
+  * [Open Definition](http://opendefinition.org/od/2.1/en/) from the Open Knowledge Foundation (for open data)
+* The license is structured to be generally usabe by anyone. It is not a "vanity" license or specific to one project, consortium or corporation. 
+* The license has actual, substantial use such that it is likely to be encountered. Substantial use may be demonstreated via use in many projects, or in one or a few highly significant projects. 
+* The license has identifiable and stable text; it is not in the midst of drafting. The license steward is committed not modifying after addition to the list and to versioning new versions in the future.
+* If the license does not substantively comply with one of the above open source definitions, the license requires the availability of the source code with limited restrictions and meets other factors listed here.
```
kestewart: See earlier comments.    Data is not Source code, and we want to encourage sharing/automation here, not restrict.

DOCS/license-inclusion-principles.md
swinslow: I think the intention is to say "binary-only software licenses won't be added to the list."  E.g. binary-use-only EULAs / software licenses. We already have various open data licenses and there's no intent to close off more of those, but rather to say that the focus for the SPDX License List is licenses that enable sharing of source code.

DOCS/license-inclusion-principles.md
swinslow: The Open Definition from the Open Knowledge Foundation addresses open content + open data, and that's one of the reference definitions listed above. So I think that should be fine?

swinslow:
@jlovejoy Thanks so much for pushing this forward! I added a couple responses above but I'll do a more thorough read-through / review as well.

DOCS/license-inclusion-principles.md
david-a-wheeler: I agree with @kestewart - the SPDX list should be able to apply to data as well as software. The text here seems to be more restrictive and I don't think that was intentional.

DOCS/license-inclusion-principles.md
jlovejoy: yes, this was about binary software and is consistent with what we've always done.  we could add "software" in there which I think would help clarify any confusion?

DOCS/license-inclusion-principles.md
jlovejoy: agree with @swinslow  - we have included various open * definitions as a first point. and if data is not source code, then this wouldn't apply to data...

jlovejoy: but we could also add "software" here, I suppose. the point is we are not opening the flood gate to any and all licenses. we still need to emphasize open / source available. I think that's been a clear intention from our discussions on calls and comments, no?

DOCS/license-inclusion-principles.md
jlovejoy: done - added "software" to modify "binary"

DOCS/license-inclusion-principles.md
jlovejoy: actually, I don't think we need "software" here - isn't the open definitions enough to clarify that open data licenses are relevant? (not to mention, we already have some on the list)

richardfontana: Some of the OSI-approved licenses are "vanity" licenses (that was the context where the term originally arose).

richardfontana: "[T]he license requires the availability of the source code" - this would only be true of copyleft licenses; is that what is intended here?

DOCS/license-inclusion-principles.md
```diff
@@ -1,24 +1,33 @@
...
-* Relevant open source community factors
-
-Comments on license inclusion will be noted in the Github issue for the license (or reference to meeting minutes, as necessary). Some license requests may be decided solely via comments in the Github issue and some may involve discussion on the bi-weekly legal call. Thus it is recommended that requestors join the mailing list and calls to fully participate.
-
-
-## Backstory
+**Definitive factors:**
+* The submitted license must not match another license already on the SPDX License List as per the SPDX matching guidelines.
+* All OSI-approved licenses will be included on the SPDX License List.
+  * This is consistent with early collaboration between SPDX and OSI (2011) to ensure that all OSI-approved licenses were represented on the SPDX License List. Also att that time, OSI began using the SPDX license identifiers in their URLs and on the license pages.
```
swinslow: minor nit: "att" => "at"

DOCS/license-inclusion-principles.md
```diff
@@ -1,24 +1,33 @@
...
+* Binary software licenses or licenses that do not provide for the avilabilty of the source code will not be included on the SPDX License List.
```
swinslow: "avilability" => "availability"

DOCS/license-inclusion-principles.md
```diff
@@ -1,24 +1,33 @@
...
+  
+**Other factors (roughly in order of importance):**
+* The license substantially complies with one of the following open source definitions (even if not submitted for approval, as in the case of OSI) or these organization consider the license to meet their definition:
+  * [Open Source Definition (OSD)](https://opensource.org/osd) from the Open Source Initiative (OSI)
+  * [Free Software Definition](https://www.gnu.org/philosophy/free-sw.en.html) from the Free Software Foundation (FSF)
+  * [Open Source Hardware Definition](https://www.oshwa.org/definition/) from the Open Source Hardware Association (OSHWA)
+  * [Open Definition](http://opendefinition.org/od/2.1/en/) from the Open Knowledge Foundation (for open data)
+* The license is structured to be generally usabe by anyone. It is not a "vanity" license or specific to one project, consortium or corporation.
```
swinslow: "usabe" => "usable"

DOCS/license-inclusion-principles.md
```diff
@@ -1,24 +1,33 @@
...
+  
+**Other factors (roughly in order of importance):**
+* The license substantially complies with one of the following open source definitions (even if not submitted for approval, as in the case of OSI) or these organization consider the license to meet their definition:
+  * [Open Source Definition (OSD)](https://opensource.org/osd) from the Open Source Initiative (OSI)
+  * [Free Software Definition](https://www.gnu.org/philosophy/free-sw.en.html) from the Free Software Foundation (FSF)
+  * [Open Source Hardware Definition](https://www.oshwa.org/definition/) from the Open Source Hardware Association (OSHWA)
+  * [Open Definition](http://opendefinition.org/od/2.1/en/) from the Open Knowledge Foundation (for open data)
+* The license is structured to be generally usabe by anyone. It is not a "vanity" license or specific to one project, consortium or corporation. 
+* The license has actual, substantial use such that it is likely to be encountered. Substantial use may be demonstreated via use in many projects, or in one or a few highly significant projects.
```
swinslow: "demonstreated" => "demonstrated"

DOCS/license-inclusion-principles.md
swinslow: @richardfontana I think the intention is more to be clear that the license list is not meant for binary-only software licenses; and/or, that this is intended to pick up "source available but not FOSS" licenses. Perhaps something like the following instead?

* If the license does not substantively comply with one of the above open source definitions, then the license is primarily intended for free distribution of content (including, in the case of software, its source code) with limited restrictions, and also meets other factors listed here.

swinslow:
Just as an example, while we're looking at this:

One of the handful of issues that came along with this rework of the license inclusion principles was the Polyform family of licenses -- see #885 and https://polyformproject.org/licenses/

Assuming for the moment that we are looking to have a set of inclusion principles that would be capable of perhaps covering the Polyform licenses (some or all of them?) -- would the working draft of principles in this PR cover those licenses?

I'm not suggesting we fully workshop that issue's analysis in this PR thread, but just mentioning it as the type of "source available" licenses that might get proposed. I think that the draft revised principles would at least bring the Polyform licenses into the realm of probable inclusion.

richardfontana:
> One of the handful of issues that came along with this rework of the license inclusion principles was the Polyform family of licenses -- see #885 and https://polyformproject.org/licenses/
> 
> Assuming for the moment that we are looking to have a set of inclusion principles that would be capable of perhaps covering the Polyform licenses (some or all of them?) -- would the working draft of principles in this PR cover those licenses?
 
I think appropriately not because Polyform licenses do not (AFAIK) have "actual, substantial use such that [they are] likely to be encountered". This relates to my previously-voiced concern that SPDX inclusion might be sought specifically to confer notoriety on otherwise-obscure licenses.

DOCS/license-inclusion-principles.md
jlovejoy: @swinslow wording added

jlovejoy:
@richardfontana @swinslow @kestewart @david-a-wheeler and anyone else - do you think we can merge this as the next iteration and then slate for discussion on Thursday's call?

swinslow:
@jlovejoy +1 from me, thank you!

Issue #925, opened 2019-09-05
License inclusion guidelines


jlovejoy:
A few months ago we began a project to update the license inclusion principles (as well as some other documentation updates in terms of both substance and location).

Specifically, currently the license inclusion guidelines are posted here: https://spdx.org/spdx-license-list/license-list-overview  These were written around 2013 and we recently recognized a need to update them.  (We also had decided we ought to move them to the Github repo, but that is somewhat tangentially to the substantive question here.)  

We have made a first draft of changes based on discussions on the calls https://wiki.spdx.org/view/Legal_Team/Minutes/2019-05-02  You can see that initial iteration here: https://github.com/spdx/license-list-XML/blob/master/DOCS/license-inclusion-principles.md

We recognized there would need to be further iterations. In particular, we have had recent submissions that are arguably not substantive open source licenses (e.g., Commons Clause, Polyform licenses) - should these be included on the SPDX License List and if so, then how does the inclusion principles need to change and where do we draw the line?  

Can you please comment here, so we can iterate on this and then come to a final, new set of inclusion principles.

jlovejoy:
discussion on 4 Sept call: some support for "source available" licenses, but maybe need for other criteria/requirements to kick in, in that case such as: must be re-usable, must be used in wild.

interim draft (https://github.com/spdx/license-list-XML/blob/master/DOCS/license-inclusion-principles.md) notes a bunch of factors that are considered with no delineation of which factors prevail. Perhaps some factors weigh more where others are missing and we can explain that.

MarkAtwood:
My thoughts:
licenses added to the top namespace under this new regime:
* have to be "generally usable", not usable by only one project, consortium, or corporation
* have to be "source available".
* the license has to be "stable", not constantly changing or redlining.
* have to have a license steward who can technically maintain the XML SPDX definitions via github pull requests without excessive hand-holding.

Licenses that dont meet those criteria instead use the DNS namespace.

baileywainwright:
My thought -- and maybe this ought to apply to any proposed license, not just those that don't seem to be "open source-ish": 

- The license must already be seeing some actual substantial use in the wild such that ordinary consumers of open source/open source-ish code are likely to encounter it. If it is reasonably suspected that the license proposer primarily wishes to use addition to the SPDX top namespace as a mechanism for publicizing a new, currently-obscure license, the license should not be included.

richardfontana:
[...]

PR #990, opened 2020-03-12
license inclusion guidelines updates

jlovejoy:
revised wording on binary license bullet (as per legal call March 11th) and date to March

### DIFF
--- a/DOCS/license-inclusion-principles.md
+++ b/DOCS/license-inclusion-principles.md
@@ -16,7 +16,7 @@ Determining whether a candidate license should be included on the SPDX License L
   * [Open Source Hardware Definition](https://www.oshwa.org/definition/) from the Open Source Hardware Association (OSHWA)
   * [Open Definition](http://opendefinition.org/od/2.1/en/) from the Open Knowledge Foundation
 * The license is structured to be generally usable by anyone. It is not a "vanity" license or specific to one project, consortium or corporation.
-* The license has actual, substantial use such that it is likely to be encountered. Substantial use may be demonstrated via use in many projects, or in one or a few highly significant projects.
+* The license has actual, substantial use such that it is likely to be encountered. Substantial use may be demonstrated via use in many projects, or in one or a few highly significant projects. For new licenses, there is definitive plans to be used in one or a few significant projects.
 * The license has identifiable and stable text; it is not in the midst of drafting. The license steward is committed not modifying after addition to the list and to versioning new versions in the future.
 * If the license does not substantively comply with one of the above open source definitions, then the license is primarily intended for free distribution of content (including, in the case of software, its source code) with limited restrictions, and also meets other factors listed here.
 * The license steward supports or is at least aware of and does not oppose its submission to the SPDX License List.
--- a/DOCS/license-inclusion-principles.md
+++ b/DOCS/license-inclusion-principles.md
@@ -8,7 +8,8 @@ Determining whether a candidate license should be included on the SPDX License L
 * All OSI-approved licenses will be included on the SPDX License List.
   * This is consistent with early collaboration between SPDX and OSI (2011) to ensure that all OSI-approved licenses were represented on the SPDX License List. Also at that time, OSI began using the SPDX license identifiers in their URLs and on the license pages.
 * Software licenses that apply only to executables and do not provide for the availability of the source code will not be included on the SPDX License List.
-* The license has identifiable and stable text; it is not in the midst of drafting. The license steward is committed to not modifying after addition to the list and to versioning new versions in the future.
+* The license has identifiable and stable text; it is not in the midst of drafting.
+* The license steward, if any, is committed to not modifying after addition to the list and to versioning new versions in the future.
   
 **Other factors (roughly in order of importance):**
 1. The license substantially complies with one of the following open source definitions (even if not submitted for approval or these organization have not considered the license):
--- a/DOCS/license-inclusion-principles.md
+++ b/DOCS/license-inclusion-principles.md
@@ -17,7 +17,7 @@ Determining whether a candidate license should be included on the SPDX License L
   * [Open Source Hardware Definition](https://www.oshwa.org/definition/) from the Open Source Hardware Association (OSHWA)
   * [Open Definition](http://opendefinition.org/od/2.1/en/) from the Open Knowledge Foundation
 * The license is structured to be generally usable by anyone. It is not a "vanity" license or specific to one project, consortium or corporation.
-* The license has actual, substantial use such that it is likely to be encountered. Substantial use may be demonstrated via use in many projects, or in one or a few highly significant projects. For new licenses, there is definitive plans to be used in one or a few significant projects.
+* The license has actual, substantial use such that it is likely to be encountered. Substantial use may be demonstrated via use in many projects, or in one or a few highly significant projects. For new licenses, there is definitive plans for the license to be used in one or a few significant projects.
 * If the license does not substantively comply with one of the above open source definitions, then the license is primarily intended for free distribution of content (including, in the case of software, its source code) with limited restrictions, and also meets other factors listed here.
 * The license steward supports or is at least aware of and does not oppose its submission to the SPDX License List.
 
--- a/DOCS/license-inclusion-principles.md
+++ b/DOCS/license-inclusion-principles.md
@@ -16,6 +16,7 @@ Determining whether a candidate license should be included on the SPDX License L
   * [Free Software Definition](https://www.gnu.org/philosophy/free-sw.en.html) from the Free Software Foundation (FSF)
   * [Open Source Hardware Definition](https://www.oshwa.org/definition/) from the Open Source Hardware Association (OSHWA)
   * [Open Definition](http://opendefinition.org/od/2.1/en/) from the Open Knowledge Foundation
+  * [Free Cultural Works definition](https://freedomdefined.org/Definition) from Freedom Defined.org
 * The license is structured to be generally usable by anyone. It is not a "vanity" license or specific to one project, consortium or corporation.
 * The license has actual, substantial use such that it is likely to be encountered. Substantial use may be demonstrated via use in many projects, or in one or a few highly significant projects. For new licenses, there is definitive plans for the license to be used in one or a few significant projects.
 * If the license does not substantively comply with one of the above open source definitions, then the license is primarily intended for free distribution of content (including, in the case of software, its source code) with limited restrictions, and also meets other factors listed here.
--- a/DOCS/license-inclusion-principles.md
+++ b/DOCS/license-inclusion-principles.md
@@ -23,7 +23,7 @@ Determining whether a candidate license should be included on the SPDX License L
 4. If the license does not substantively comply with one of the above open source definitions, then the license is primarily intended for free distribution of content (including, in the case of software, its source code) with limited restrictions, and meets other factors listed here.
 5. The license steward supports or is at least aware of and does not oppose its submission to the SPDX License List.
 
-The SPDX legal team will also consider any other information considered relevant by the license steward or open source community.
+The SPDX legal team will also consider any additional information considered relevant by the license steward or open source community.
 
 Comments on license inclusion will be noted in the Github issue for the license (or reference to meeting minutes, as necessary). Some license requests may be decided solely via comments in the Github issue and some may involve discussion on the bi-weekly legal call. Thus it is imperative that requestors join the mailing list and calls to fully participate.
 
--- a/DOCS/license-inclusion-principles.md
+++ b/DOCS/license-inclusion-principles.md
@@ -7,7 +7,7 @@ Determining whether a candidate license should be included on the SPDX License L
 * The submitted license must not match another license already on the SPDX License List as per the SPDX matching guidelines.
 * All OSI-approved licenses will be included on the SPDX License List.
   * This is consistent with early collaboration between SPDX and OSI (2011) to ensure that all OSI-approved licenses were represented on the SPDX License List. Also at that time, OSI began using the SPDX license identifiers in their URLs and on the license pages.
-* Binary-only software licenses or licenses that do not provide for the availability of the source code will not be included on the SPDX License List.
+* Software licenses that apply only to executables and do not provide for the availability of the source code will not be included on the SPDX License List.
   
 **Other factors (roughly in order of importance):**
 * The license substantially complies with one of the following open source definitions (even if not submitted for approval, as in the case of OSI) or these organization consider the license to meet their definition:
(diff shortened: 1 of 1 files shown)


jlovejoy:
various other updates as per discussion on March 12th call and noted in #925 made - ready for final review!

jlovejoy:
randomly asked for reviewers - comments and +1s welcome!!!

DOCS/license-inclusion-principles.md
```diff
@@ -7,42 +7,41 @@ Determining whether a candidate license should be included on the SPDX License L
...
-* If the license does not substantively comply with one of the above open source definitions, then the license is primarily intended for free distribution of content (including, in the case of software, its source code) with limited restrictions, and also meets other factors listed here.
-* The license steward supports or is at least aware of and does not oppose its submission to the SPDX License List.
+  * [Open Definition](http://opendefinition.org/od/2.1/en/) from the Open Knowledge Foundation
+  * [Free Cultural Works definition](https://freedomdefined.org/Definition) from Freedom Defined.org
+2. The license is structured to be generally usable by anyone. It is not a "vanity" license or specific to one project, consortium or corporation.
+3. The license has actual, substantial use such that it is likely to be encountered. Substantial use may be demonstrated via use in many projects, or in one or a few highly significant projects. For new licenses, there is definitive plans for the license to be used in one or a few significant projects.
```
swinslow: My only comment here is, in the last sentence, should it be "one or a few **highly** significant projects" so that it's aligned with the standard in the preceding sentence?

kemitchell:
@jlovejoy is the historical background section "operative", or could it be separated into a different document with a clarification that the rules and factors are the rules and factors, and the historical background is just that?

DOCS/license-inclusion-principles.md
```diff
@@ -7,42 +7,41 @@ Determining whether a candidate license should be included on the SPDX License L
 * The submitted license must not match another license already on the SPDX License List as per the SPDX matching guidelines.
 * All OSI-approved licenses will be included on the SPDX License List.
   * This is consistent with early collaboration between SPDX and OSI (2011) to ensure that all OSI-approved licenses were represented on the SPDX License List. Also at that time, OSI began using the SPDX license identifiers in their URLs and on the license pages.
-* Binary-only software licenses or licenses that do not provide for the availability of the source code will not be included on the SPDX License List.
+* Software licenses that apply only to executables and do not provide for the availability of the source code will not be included on the SPDX License List.
```
kemitchell:
```suggestion
* Software licenses that only license executables, and don't apply to source code, will not be included on the SPDX License List.
```

DOCS/license-inclusion-principles.md
```diff
@@ -7,42 +7,41 @@ Determining whether a candidate license should be included on the SPDX License L
...
+* The license has identifiable and stable text; it is not in the midst of drafting. The license steward is committed to not modifying after addition to the list and to versioning new versions in the future.
```
kemitchell:
```suggestion
* The license has identifiable and stable text; it is not in the midst of drafting.
* The license steward, if any, is committed to not modifying after addition to the list and to versioning new versions in the future.
```

Historical licenses may not have active stewards, either because they're no longer maintaining the license, or because they're defunct, &c.

DOCS/license-inclusion-principles.md
```diff
@@ -7,42 +7,41 @@ Determining whether a candidate license should be included on the SPDX License L
...
   
 **Other factors (roughly in order of importance):**
-* The license substantially complies with one of the following open source definitions (even if not submitted for approval, as in the case of OSI) or these organization consider the license to meet their definition:
+1. The license substantially complies with one of the following open source definitions (even if not submitted for approval or these organization have not considered the license):
```
kemitchell: It's good to have a list of these definitions, but I'm still against license submission by proxy. The "substantial" compliance perhaps conveys that SPDX discussion is no substitute for submission to other orgs. But it's not particularly clear how we'll go about discussing "substantially", either.

Kind of reminds me of federal courts deciding whether there's been a "material" effect under state contract law. Awkward.

To make this more concrete: Does this mean no more noncommercial licenses, from Creative Commons or anyone else?

DOCS/license-inclusion-principles.md
```diff
@@ -7,42 +7,41 @@ Determining whether a candidate license should be included on the SPDX License L
...
-* The license has actual, substantial use such that it is likely to be encountered. Substantial use may be demonstrated via use in many projects, or in one or a few highly significant projects.
-* The license has identifiable and stable text; it is not in the midst of drafting. The license steward is committed not modifying after addition to the list and to versioning new versions in the future.
-* If the license does not substantively comply with one of the above open source definitions, then the license is primarily intended for free distribution of content (including, in the case of software, its source code) with limited restrictions, and also meets other factors listed here.
-* The license steward supports or is at least aware of and does not oppose its submission to the SPDX License List.
+  * [Open Definition](http://opendefinition.org/od/2.1/en/) from the Open Knowledge Foundation
+  * [Free Cultural Works definition](https://freedomdefined.org/Definition) from Freedom Defined.org
+2. The license is structured to be generally usable by anyone. It is not a "vanity" license or specific to one project, consortium or corporation.
```
kemitchell:
```suggestion
2. The license is not specific to one project, consortium or corporation.
```

"Vanity" is pejorative, but not particularly clear. I associate it mostly with antiproliferation, in the sense of not welcoming new licenses with nearly the same purported legal effect as existing terms. But SPDX is full of licenses with substantially similar purported effects. And I've often found it useful precisely for that broader comprehensiveness. That allowed us to code [Blue Oak's permissive license list](https://blueoakcouncil.org/list) in SPDX IDs, for example.

DOCS/license-inclusion-principles.md
kemitchell:
```suggestion
3. The license has actual, substantial use such that it is likely to be encountered. Substantial use may be demonstrated via use in many projects, or in one or a few highly significant projects. For new licenses, there are definitive plans for the license to be used in one or a few significant projects.
```

Typo fix above, but in general, I'm a bit dismayed by this rule. The words "actual", "substantial", "likely", "many", "significant", and "definitive" imply a bar being raised, but I'm not sure how high.

DOCS/license-inclusion-principles.md
```diff
@@ -7,42 +7,41 @@ Determining whether a candidate license should be included on the SPDX License L
...
+3. The license has actual, substantial use such that it is likely to be encountered. Substantial use may be demonstrated via use in many projects, or in one or a few highly significant projects. For new licenses, there is definitive plans for the license to be used in one or a few significant projects.
+4. If the license does not substantively comply with one of the above open source definitions, then the license is primarily intended for free distribution of content (including, in the case of software, its source code) with limited restrictions, and meets other factors listed here.
```
kemitchell: I think folks will definitely stop at number one when they want to oppose a license, and not bother reading this far.

I wonder what motivated the idea of "limited" restrictions.

DOCS/license-inclusion-principles.md
```diff
@@ -7,42 +7,41 @@ Determining whether a candidate license should be included on the SPDX License L
...
+5. The license steward supports or is at least aware of and does not oppose its submission to the SPDX License List.
```
kemitchell: Does SPDX still see submission of historical, perhaps historically significant, license terms and variations?

DOCS/license-inclusion-principles.md
```diff
@@ -7,42 +7,41 @@ Determining whether a candidate license should be included on the SPDX License L
...
 
-*last updated Feb 2020*
+*last updated Mar 2020*
 
-## Historical Background
-The first beta version of the SPDX License List was published in August 2010 and had approximately one hundred licenses on it. The initial set of licenses was included based on informal discussion and consensus on the SPDX working group calls and email list. Although various "guidelines" were identified in regards to which licenses to include or not during this time, formal guidelines were not recorded beyond the meeting minutes. We had to start somewhere, and the obvious was the easy beginning point. Decisions or guidelines that evolved out of discussion or by implication included the following:
-* Include commonly found open source licenses. Although discussion was not explicit in regards to how to define an "open source license," this was always an implicit guiding principle
-* Include all OSI approved licenses, both current and those approved but now deprecated. The rationale being that once OSI-approved, always OSI-approved and deprecated licenses still appear "in the wild"
+# Historical Background
```
kemitchell: Per my earlier comment on the PR, I wonder whether this is part of the guidelines, to be read with them, or could be moved to a separate document.

DOCS/license-inclusion-principles.md
```diff
@@ -7,42 +7,41 @@ Determining whether a candidate license should be included on the SPDX License L
...
+2. The license is structured to be generally usable by anyone. It is not a "vanity" license or specific to one project, consortium or corporation.
+3. The license has actual, substantial use such that it is likely to be encountered. Substantial use may be demonstrated via use in many projects, or in one or a few highly significant projects. For new licenses, there is definitive plans for the license to be used in one or a few significant projects.
+4. If the license does not substantively comply with one of the above open source definitions, then the license is primarily intended for free distribution of content (including, in the case of software, its source code) with limited restrictions, and meets other factors listed here.
+5. The license steward supports or is at least aware of and does not oppose its submission to the SPDX License List.
 
-The SPDX legal team will also consider any information considered relevant by the license steward or open source community.
+The SPDX legal team will also consider any other information considered relevant by the license steward or open source community.
```
zvr: "additional" in place of "other" ?

DOCS/license-inclusion-principles.md
jlovejoy: hmmm... or remove "highly" from the previous sentence?  I'm not sure I know the difference b/w "highly significant" or "significant"...

jlovejoy:
> @jlovejoy is the historical background section "operative", or could it be separated into a different document with a clarification that the rules and factors are the rules and factors, and the historical background is just that?

@kemitchell - could move it, my only concern is that if we move it, where to and will people find it? We are kind of in the midst of re-organizing, so we could consider moving it later as other things shape up. In any case, open to ideas whether for now or for later.

DOCS/license-inclusion-principles.md
jlovejoy: so, executables instead of binaries is the key change here... I don't have a strong opinion on that one. 
@zvr @goneall @kestewart - thoughts?

DOCS/license-inclusion-principles.md
jlovejoy: that seems like a good addition, thanks!

DOCS/license-inclusion-principles.md
jlovejoy: @kemitchell - you are a lawyer, you are used to awkward, gray area delineators!! :)
we do *not* want to be in the position of saying what does or does not meet someone else's definition (nor have we ever wanted to be in that position). That would be inappropriate.  So, we need that flexibility.  

Also, given all we've observed over the past year and a half, flexibility will allow the guidelines to flex as needed going forward (we don't want to have to update these too often, I'd hope!)

As for NC licenses - see historical background ;)  (which is why I wanted to include the historical background, so we don't have to keep answering that question and can say - here, read this!) I don't know - if it meets all the other criteria, I suppose it could be possible?  

case-by-case scenario... totality of the circumstances...

DOCS/license-inclusion-principles.md
jlovejoy: huh, never really thought of "vanity" as pejorative - just that it's a term used that seems to have general understanding among open source license folks (who've been around awhile) - but it's not a necessary term.

I do think that we should keep the "The license is structured to be generally usable by anyone" part (or the like) - I was trying to phrase things in the positive (then add the negative for color as needed). 

perhaps:
The license is structured to be generally usable by anyone. It is not specific to one project, consortium or corporation.

??

DOCS/license-inclusion-principles.md
jlovejoy: well, if we think about how we've weighed this factor - it started as simply "use in the wild" but over time, that isn't enough as people write new licenses and want to have an SPDX id from the get-go. So we were trying to provide a flexible guideline, that also still allows rejecting silly licenses that one project uses and are really similar to a well-known existing license and they really ought to just use that instead of making something new. (we've had some examples of this)

DOCS/license-inclusion-principles.md
jlovejoy: I think that goes along with "substantially complies"...

DOCS/license-inclusion-principles.md
jlovejoy: yes

DOCS/license-inclusion-principles.md
jlovejoy: see comment above

DOCS/license-inclusion-principles.md
jlovejoy: sure, that works

DOCS/license-inclusion-principles.md
swinslow: That works too and is a fair point. I think as long as they're aligned then I could be fine with either.

DOCS/license-inclusion-principles.md
swinslow: I don't have a strong opinion, but lean in favor of changing to "executables" as @kemitchell suggested. I think that's closer to capturing the intent of what we're trying to say, since not all programming languages' "executables" would be "binary-only" in a literal sense.

DOCS/license-inclusion-principles.md
swinslow: +1 to @jlovejoy's comment here. 

Flexibility is relevant, and I think it should be understood in light of both (a) the remaining factors on this list; and (b) the intent of making the current guidelines changes, which is largely to make it permissible for more source-available licenses to be added to the list.

DOCS/license-inclusion-principles.md
swinslow: Agree with the change in @jlovejoy's comment here -- important to keep "structured to be generally usable by anyone" but fine with the change to remove "vanity license."

DOCS/license-inclusion-principles.md
swinslow: Yes, e.g. decades-old license statements from the Linux kernel or elsewhere which predate the standardization of license texts.

Many of these are clearly compliant with the OSD or other such standards, and are typically seen as appropriate to add to the list given their use in one or a few widely-propagated projects, even if not in a large number of different projects.

DOCS/license-inclusion-principles.md
jlovejoy: change to executables seems to have favor. but I still think we need to keep "do not provide for availability of source code" as that's more specific than "don't apply to source code"

kemitchell: Folks would say permissive licenses "don't provide for availability of source code".

jlovejoy:
discussed various comments logged here on 26th March legal call.
 ready for flight!

DOCS/license-inclusion-principles.md
jlovejoy: that was brought up before and I think this was the language that came out of that conversation - as in the source code is available (not requires availability downstream)

@swinslow just merged! you should have joined call, which just ended!! :)

jlovejoy: "only applies to executables" is also clarifying - no one would say that about MIT

kemitchell: I can't do long standing calls during business hours. I need that time for client work and client calls.

PR #1000, opened 2020-04-05
Deprecate BSD-2-Clause-NetBSD

### LINKED ISSUE #980: Deprecate BSD-2-Clause-NetBSD
[BSD-2-Clause-NetBSD](https://spdx.org/licenses/BSD-2-Clause-NetBSD.html) is identical to [BSD-2-Clause](https://spdx.org/licenses/BSD-2-Clause.html), with one exception -- the -NetBSD variant includes the following at the beginning:

```
This code is derived from software contributed to The NetBSD Foundation by 
```

Setting aside the fact that the sentence is incomplete and probably needs markup, if it remains...

I don't think that this is sufficiently distinctive from standard BSD-2-Clause that it warrants having a separate identifier. I view this line as being more akin to a copyright notice, describing perhaps the source of code but not saying anything relevant from a license perspective.

I'm not suggesting we remove it from the license list, since it already exists on the list and folks might be using the ID. But I would suggest that we move it to the deprecated category and add a note that it is viewed as being substantively identical to BSD-2-Clause.

swinslow:
This patch deprecates BSD-2-Clause-NetBSD in favor of the standard
BSD-2-Clause license, as discussed in #980.

For those who do choose to continue using it, this also fixes the
markup for the "This code is derived from..." line so that it does
not render on the license list as an incomplete sentence.

Fixes #980

### DIFF
--- a/src/BSD-2-Clause-NetBSD.xml
+++ b/src/BSD-2-Clause-NetBSD.xml
@@ -21,7 +21,7 @@
       </copyrightText>
 
       <p>This code is derived from software contributed to The NetBSD Foundation by
-        <alt match=".*" name="contributors">[contributor name]</alt>.
+        <alt match=".*" name="contributors"/>
       </p>
       <p>Redistribution and use in source and binary forms, with or without modification, are permitted provided
          that the following conditions are met:</p>
--- a/src/BSD-2-Clause-NetBSD.xml
+++ b/src/BSD-2-Clause-NetBSD.xml
@@ -11,9 +11,9 @@
       </crossRefs>
       <notes>
         NetBSD adopted this 2-clause license in 2008 for code contributed to NetBSD Foundation.
-        It is being deprecated as a separate license because the only difference from BSD-2-Clause
-        is the line describing that the code is derived from software contributed to NetBSD,
-        which is not viewed as being a substantive part of the license text.
+        It is being deprecated as a separate license because it is a match to BSD-2-Clause.
+        The line describing that the code is derived from software contributed to NetBSD
+        is not viewed as being a substantive part of the license text.
       </notes>
     <text>
       <copyrightText>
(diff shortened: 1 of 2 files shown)


swinslow:
@jlovejoy @goneall @silverhook -- would love to get a +1 from any of you if you're in favor of this, per the discussion in #980  :)

src/BSD-2-Clause-NetBSD.xml
```diff
@@ -1,18 +1,27 @@
...
+        which is not viewed as being a substantive part of the license text.
+      </notes>
     <text>
       <copyrightText>
          <p>Copyright (c) 2008 The NetBSD Foundation, Inc. All rights reserved.</p>
       </copyrightText>
 
       <p>This code is derived from software contributed to The NetBSD Foundation by
-        <alt match=".*" name="contributors"/>
+        <alt match=".*" name="contributors">[contributor name]</alt>.
```
goneall, line 24: I think we may want to leave this with the incomplete sentence to be compatible with the "official" text at https://www.netbsd.org/about/redistribution.html.  There has been a lot of issues raised on the license list publisher, the license list data, and the license list XML that we don't retain the canonical text.  Making this change will deviate from the canonical text.

swinslow: fair enough! thanks @goneall, I'll update the PR to revert this part.

swinslow:
@goneall the change to the matching part of the template is now reverted. Are you otherwise good with the rest of the PR? Thanks!

swinslow:
Hmmm... it looks like Travis failed, not on the validation but with a rather unhelpful error message in the Node.js stage:

image

@goneall Any thoughts what might cause this?

goneall:
This is indeed a rather odd error.

I restarted the build and it looks like it passed it without error this time.

swinslow:
Discussed on legal team call 2020-04-09, updated notes to clarify. Will wait until tests pass and then merge.

PR #1002, opened 2020-04-05
Add test.log to .gitignore

swinslow:
This is just to add test.log to the excluded files list, which is generated locally when using the test-one-license command.


swinslow:
I can't imagine anyone will object to this, so I'm merging it  :)

PR #1008, opened 2020-04-14
Add CAL-1.0.xml

VanL:
Here is a first take on CAL-1.0.xml. I have not yet created the second CAL-1.0-Combined-Work-Exception; I figure that we can get this in place and the second will be based off of this.


VanL:
Updated as follows:
- Change word wrapping
- Fix errors (Section 6 was missed; typos)
- Make markdown the "official" text format
- Add a .txt (actually markdown) to the test folder for comparison by TravisCI

VanL:
@swinslow: Can you take a look at the CI build to see why it is failing? I tried to match what was done in the Parity license to mark the minimal markdown elements as optional, but the CI build has warnings.

VanL:
Updated optional marks to try to pass CI test.

swinslow:
Hi @VanL, apologies for delayed response. From a quick glance I don't see what's causing this to fail, so I will dig into it in more detail over the weekend if I'm not able to later today. Thanks for submitting this and I'll make sure we get it included in this release after we sort out the issue here.

VanL:
@swinslow: Any thoughts on why this might be failing?

src/CAL-1.0.xml
```diff
@@ -0,0 +1,470 @@
...
+      <crossRef>https://opensource.org/licenses/CAL-1.0</crossRef>
+    </crossRefs>
+    <notes>The first draft of this license was released February 2019, and the fourth revision was approved by the OSI February 2020.</notes>
+    <text>
+      <titleText>
+        <p><optional># </optional>The Cryptographic Autonomy License, v. 1.0</p>
+      </titleText>
+        <p>
+<optional>*</optional>This Cryptographic Autonomy License (the "License") 
+applies to any Work whose owner has marked it with any of the following
```
goneall, line 15: The tests are probably failing due to the `<optional>*</optional>`.  There is a known bug with the tool that publishes the license where it doesn't deal with single characters being optional.  If this is the case, you can work around this by making it a variable text '<alt match="(\*This|This)" name="this">*This</alt>`

swinslow:
Thanks @goneall for finding the issue here -- and @VanL, apologies again for my being slow to respond in the midst of current chaos. Glad to see this is passing now!

My only edit is going to be to add `listVersionAdded="3.9"` as an attribute in the `<license>` tag on line 3, since that tracks the license list release for which this one was first added. I'm going to add that as an additional commit now, and after the tests complete successfully again, I'll merge the PR. Thanks!

PR #1014, opened 2020-04-22
Add CAL-1.0-Combined-Work-Exception

VanL:


expected-warnings
```diff
@@ -1,2 +1,2 @@
...
...licates licenses: LGPL-2.1, LGPL-2.1-only","Duplicates licenses: LGPL-2.1-only, LGPL-2.1","Duplicates licenses: MPL-2.0-no-copyleft-exception, MPL-2.0","Duplicates licenses: GFDL-1.3-or-later, GFDL-1.3-only","Duplicates licenses: GFDL-1.2-or-later, GFDL-1.2-only","Duplicates licenses: GFDL-1.3-or-later, GFDL-1.3-only","Duplicates licenses: AGPL-3.0-or-later, AGPL-3.0-only","Duplicates licenses: GPL-1.0-only, GPL-1.0-or-later","Duplicates licenses: AGPL-1.0-only, AGPL-1.0-or-later","Duplicates licenses: AGPL-3.0-only, AGPL-3.0-or-later","Duplicates licenses: GFDL-1.2-only, GFDL-1.2-or-later","Duplicates licenses: GFDL-1.3-only, GFDL-1.3-or-later","Duplicates licenses: GPL-3.0-only, GPL-3.0-or-later","Duplicates licenses: LGPL-3.0-only, LGPL-3.0-or-later","Duplicates licenses: OFL-1.1-RFN, OFL-1.1","Duplicates licenses: OFL-1.1, OFL-1.1-RFN","Duplicates licenses: OFL-1.0, OFL-1.0-no-RFN","Duplicates licenses: OFL-1.0-no-RFN, OFL-1.0","Duplicates licenses: OFL-1.0-RFN, OFL-1.0-no-RFN","Duplicates licenses: OFL-1.0-no-RFN, OFL-1.0-RFN","Duplicates licenses: OFL-1.0-RFN, OFL-1.0","Duplicates licenses: OFL-1.0, OFL-1.0-RFN","Duplicates licenses: CAL-1.0, CAL-1.0-Combined-Works-Exception"
```
swinslow, line 2: @VanL So here, frustratingly, I think you'll need to put the "Duplicates licenses" in both possible orders, e.g.:

"Duplicates licenses: CAL-1.0, CAL-1.0-Combined-Works-Exception","Duplicates licenses: CAL-1.0-Combined-Works-Exception, CAL-1.0"

I see that the first one appears twice in the list above, but I think the second might need to be added.

This is another one of those things where I'm sure we would improve the tooling for this if we had infinite time  :)  But try that and see if it fixes the duplicates error.

swinslow:
Oh, I see what it is -- you're using two different identifiers: it's "Work" in the .xml file but "Works" (with the s) in the expected warnings file...

I see from the license text that it's meant to be "Work" without the "s" -- here, I'll edit the expected warnings file to see if it fixes it...

VanL:
So this passes - but on balance I'd prefer "Combined Works" rather than "Combined Work." I think it sounds better.
 I'll update and resubmit.

VanL:
Actually, I changed my mind again. Singular "Work" makes the best sense in the various contexts. Please go ahead and merge.

swinslow:
Sounds good, thanks @VanL -- I was also just about to note the various places where it would need to be updated to switch from singular to plural  :)  I'll merge it as-is!

PR #1017, opened 2020-04-23
Add PolyForm-Small-Business-1.0.0

kemitchell:
Split off from #903
Issue: #885


swinslow:
Agreed to add per discussion in #885; tests and metadata looks good; merging.

kemitchell:
Thanks again!

PR #1019, opened 2020-04-23
Add PolyForm-Internal-Use-1.0.0

kemitchell:
Split off from #903
Issue: #885


jlovejoy:
not adding at this time

PR #1021, opened 2020-04-23
Add rename-license script

kemitchell:
This PR adds a short script for renaming licenses. The script handles XML file names, `licenseId` properties within XML files, and test files, when present.

### DIFF
--- a/rename-license
+++ b/rename-license
@@ -1,0 +1,25 @@
+#!/bin/bash
+set -e
+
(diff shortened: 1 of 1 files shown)


swinslow:
Excellent, thanks for submitting this @kemitchell! I know we don't change existing identifiers once they're on the list, but will be useful when testing new licenses before an identifier is finalized.

PR #1023, opened 2020-05-07
added that is now OSI approved

jlovejoy:
add tag that this license is now OSI-approved


swinslow:
Thanks @jlovejoy! Looks good, merging.

Issue #1027, opened 2020-05-12
Add Solderpad 2.1 Licence as exception to Apache 2.0


andrewjskatz:
Proposed SPDX Identifier: 

APACHE-2.0-WITH-SHL-2.1

Solderpad 2.0 has generated a small number of issues, and this had led to a small point release of Solderpad 2.1 (which has not been officially released yet: official release will coincide with acceptance of Solderpad 2.1 onto the SPDX license list (if approved). 

The changes from version 2.0 are summarised as follows:

1. The preamble has been clarified, to remove the implication that that the Apache 2.0 imposes obligations (as a licence, it contains conditions, but not obligations).
2. The text of version 2.0 contains bold and italic formatting to assist with interpretation (and to indicate changes from Apache 2.0). This was found to be confusing, and version 2.1 is formatted in text only. Accordingly, the last sentence of the preamble in 2.0 (which specifies that the bold and italic formatting is to be ignored) was removed for version 2.1 as it is no longer necessary
3. The definition of “Rights” has been modified slightly to ensure that the usage of “Mask Works” is consistent with the definition of “Object Form”

I attach a copy of the text version of Solderpad 2.1, which will (shortly!) be hosted at a persistent URL on the solderpad.org website. 
[Solderpad 2.1.txt](https://github.com/spdx/license-list-XML/files/4615863/Solderpad.2.1.txt)

swinslow:
Thanks @andrewjskatz! Appreciate all your help with this. It sounds like the 2.1 version will be cleaner both for general use and also for adding to the license list.

A couple of things I wanted to clarify:

* Is the intention that this would align with what we'd discussed for Solderpad 2.0, where it would be added to the Exceptions list? If so, then the identifier for that list would just be `SHL-2.1`, and in regular usage it would be `Apache-2.0 WITH SHL-2.1`. 

(Nitpicking a bit just to note that there would be no hyphens around the WITH keyword, and also noting that Apache-2.0 isn't all-caps. Though in [the latest version of the SPDX spec](https://spdx.github.io/spdx-spec/v2-draft/appendix-IV-SPDX-license-expressions/#case-sensitivity), we did say that license IDs should be matched in a case-_insensitive_ manner, so.)  :)

* Would you still be looking to add Solderpad 2.0 to the list (see #945), or should we close that issue if it's not being used going forward? I don't know whether you're expecting the Solderpad 2.0 license will remain in significant use or if most / all folks will be switching to 2.1.

[...]

PR #1028, opened 2020-05-13
New exception: SHL-2.1

jlovejoy:
xml file for SHL-2.1


jlovejoy:
hmmm, thought I add .txt file to PR, but not seeing it...

jlovejoy:
both files there now, but still failing - sigh

swinslow:
I'm on calls for next couple of hours, but I'll take a look  :)  Thanks for getting these going!

jlovejoy:
@swinslow - it seems to think there's no .txt file and its looking in the exceptions folder - but we've added other exceptions and passed tests?
@goneall might have an insight

swinslow:
looking at the test log, in lines 222 and 223 it looks like it's having issues with parsing the XML file as invalid:

>Processing License List19:32:45.022 [main] ERROR org.spdx.licensexml.LicenseXmlDocument - Invalid license XML file SHL-2.1.xml
>org.xml.sax.SAXParseException: cvc-complex-type.2.4.a: Invalid content was found starting with element '{"http://www.spdx.org/license":standardLicenseHeader}'. One of '{"http://www.spdx.org/license":p, "http://www.spdx.org/license":bullet, "http://www.spdx.org/license":list, "http://www.spdx.org/license":optional, "http://www.spdx.org/license":alt, "http://www.spdx.org/license":br, "http://www.spdx.org/license":titleText, "http://www.spdx.org/license":copyrightText}' is expected.

I'm taking a closer look, will see if I can tell why it might be failing...

swinslow:
Ah, I think it might be because the `<standardLicenseHeader>` section is appearing within the `<text>` section. Looking at another example I think it might need to be outside. I'm jumping on another call but will experiment afterwards to see if that fixes it...

jlovejoy:
> Ah, I think it might be because the `<standardLicenseHeader>` section is appearing within the `<text>` section. Looking at another example I think it might need to be outside.

I don't think it's that, as I have it the same way as in Apache-2.0

zvr:
@jlovejoy @swinslow the two `</br>` in lines 38 and 39 should probably be `<br/>`  :wink:

swinslow:
I just pushed another commit which should fix it (at least it's working on the test on my local machine) -- two changes:
* fixed the br tags -- actually just removed and replaced them with p tags
* removed the standardLicenseHeader tags

My best guess is that there's something about using standardLicenseHeader in an exception that is causing it to break. From a quick search it doesn't look like any of the other exceptions use this tag. Will wait to confirm that it passes...

Note also that I gather there have been some updates to the upstream text (per discussion in #1027) so we should make sure to capture those too before merging.

swinslow:
Good to go now. @andrewjskatz, could you take a look at https://raw.githubusercontent.com/spdx/license-list-XML/d2397ad0e0dac0c225bceccd65833705853cd3ec/test/simpleTestForGenerator/SHL-2.1.txt and confirm that you are good with locking it in as the 2.1 text?

andrewjskatz:
Looks great - thanks Steve!

Andrew Katz
+44 7970 835001
I'm emailing from my smartphone, so please excuse terseness!

swinslow:
Looks like we're good to go -- I'm merging this. Thank you @jlovejoy @andrewjskatz @zvr!

PR #1029, opened 2020-05-13
New exception: SHL-2.0

jlovejoy:
new exception for SHL-2.0


jlovejoy:
@andrewjskatz - have a look?

jlovejoy:
@swinslow @goneall @zvr  - do we have a way to make italics from the XML file and then display as such on the web pages?  
I don't think we've had that need ever before...

zvr:
@jlovejoy we never had italics or bold or larger font size, etc., since we always maintained that the license text was enough to convey the license terms. And most licenses also come in a pure-text format for people to add in their projects...

I went through the 5 first pages of Google search results for "solderpad hardware license" and everyone is pointing to the same page. It seems that the test file in this PR is the only pure text rendering of the license :D

PR #1035, opened 2020-05-15
CAL-1.0-Combined-Work-Exception: Fix title and notes

### LINKED ISSUE #1034: CAL-1.0 and CAL-1.0-Combined-Work-Exception have same display name
While preparing the release for later today, I noticed a problem:

For CAL-1.0 and CAL-1.0-Combined-Work-Exception, we gave them separate IDs to indicate whether or not the exception is being used. However, the XML files have exactly the same title in their metadata, and nothing in their notes to explain the difference.

This means that on the license list, it would correctly list the two separate identifiers, but the license names and notes for the two would be identical, and it would not be obvious what the difference is between the two.

I'm going to put together a quick PR to fix this. @jlovejoy @VanL because we're hoping to push the release later today, you might not have a chance to see this. But if you do and you have any problems with me making the change, please yell at me  :)

swinslow:
Fixes #1034


swinslow:
@jlovejoy @VanL let me know if you have any issues with this, otherwise assuming the tests pass I will plan to merge before the release later today.

Issue #992, opened 2020-03-13
Add notes to CPAL-1.0, NOSL, SPL-1.0 and Interbase-1.0 that they are modified MPL-1.1


silverhook:
The following licenses are basically rebranded MPL-1.1 with some additional terms:

- CPAL-1.0 – apart from “Initial Developer” expands to “Original Developer”, major changes are that  in §14 it includes an advertising claus  and in §15 it extends its (weak) copyleft also to network/SaaS use
- NOSL – new §7.1 that expands on the warranty disclaimer, in §11 change of jurisdiction and forum
- SPL-1.0 – expands §1.3 for “Covered Code” to include documentation, removes ‘as is’ clause
- Interbase-1.0 – new AMENDMENTS section includes clauses for no-trademark, no-endorsement, project naming rules and an advertising clause; and a new §6.4 explaining the origin of the licenmse and the changes in AMENDMENTS 

I’d suggest something adding notes to all of them, similar as with CUA-OPL-1.0:
https://github.com/spdx/license-list-XML/pull/711/files

BTW, this might be a good time to actually deprecate CUA-OPL-1.0, as it is marked as retired on OSI already – see https://github.com/spdx/license-list-XML/issues/690

swinslow:
I haven't yet compared them myself, so with that caveat :) I agree with @silverhook. It's great to be able to highlight to folks in the notes when one license is basically the same as another.

jlovejoy:
a note on each of these sounds like a good idea.
@silverhook - do you want to draft something in a PR for each?

silverhook:
Expect the pull requests in the next few hours. Sorry for the delay…

swinslow:
OK, I'll hold on the release until this afternoon then  :)

swinslow:
Fixed in #1036, #1037, #1038 and #1040, thanks @silverhook!

PR #1036, opened 2020-05-15
add notes to CPAL-1.0

silverhook:
Explain how CPAL-1.0 is a modified MPL-1.1.

partially solves https://github.com/spdx/license-list-XML/issues/992


swinslow:
Thanks @silverhook! Looks good to me.

swinslow:
I'm good with adding now so we can include the note for 3.9. If people take issue later with the characterization of these, we can always update the notes later.

PR #1041, opened 2020-05-22
Update PHP-3.01.xml

emmanueltournier:
Changing isOsiApproved from "False" to "True"


swinslow:
Looks good, thanks @emmanueltournier!

Issue #970, opened 2020-01-30
Consider license expression syntax for selecting options


swinslow:
(last updated 2020-02-06)

Background
------------

A handful of license list requests have involved questions of how to select or choose among options contained within a single license -- and whether that should be articulable in the license ID / expression syntax. Here are a few examples:

* #686: The GFDL licenses permit licensors to optionally designate certain content as Invariant Sections or Front-Cover / Back-Cover Texts. If any are present, then the impact and obligations for licensees is different than if they are not. (To the extent that Debian considers GFDL non-free if these are present, but free if they are not; see #686 for discussion.)

* #724: Similarly, for OFL, the licensor can choose to designate a "Reserved Font Name", or not; if they do, the licensee obligations change.

* #928: Some of the GPL family of licenses permit licensors to designate a proxy who can determine whether the code can optionally be used under future versions of the *GPL, rather than only this version.

* Similarly, the whole question of "-only" vs. "-or-later" is also in some sense a matter of choosing between two options articulated in the license, at least for *GPL.

* #939: For a templated exception to GPL-3.0 recommended by the FSF, the licensor can specify particular libraries under particular licenses that may be linked and conveyed. The template can match all of these, but the licensor may also want to represent _which_ libraries / licenses in the identifier expression.

Question
---------
Should the license list, and the license identifier syntax, take these "selection" options into account in some fashion? And if so, how?

Options discussed so far
-------------------------
These are some of the options that were discussed on the legal team call on 2020-01-16. The advantages and disadvantages listed below are mostly my own thoughts, as informed by that discussion.

1) **Don't change anything.** Keep the license list focused on just identifying license texts. It is up to licensors and licensees to interpret their obligations based on knowing that the license is present. If people want to be more precise, they can use `LicenseRef-` identifiers to define custom meanings.
    * _Advantages_: Least effort for license list contributors. Keeps the license list focused and short.
    * _Disadvantages_: Doesn't really solve the problem articulated above. Users of `LicenseRef-` will define them in non-standard ways that hinder interoperability.

2) **Add license list entries for all selections.** If there are two different possible selections within a license, then create two different entries on the list, one for each selection. (And also possibly a third entry, to mean "I know the license is here but I'm not asserting anything about the selection.")
    * _Advantages_: Most precise and standardized. Can specifically articulate what is selected.
[...]

PR #1048, opened 2020-06-16
Add GFDL variants for invariants and no-invariants

### LINKED ISSUE #686: GNU Free Document Licence with no Invariant Sections
Several old documentation files were written using GFDL licence, with was, on that time, believed to be a free license. However, a more modern read of the license considers it non-free, if an optional "Invariant Sections" part of the license is used. As a reference, see: https://www.debian.org/vote/2006/vote_001

Currently, at https://spdx.org/licenses/, it is only possible to state that a document follows FDL (for example, using SPDX identifier GFDL-1.1-or-later), but currently ther's no way to distinguish between a GFDL with Invariant Sections from a GFDL without it. As explained at: https://spdx.org/licenses/GFDL-1.1-or-later.html ADDENDUM, the files licensed under GFDL that doesn't use Invariant Sections should follow this rule:

	'If you have no Invariant Sections, write "with no Invariant 
	 Sections" instead of saying which ones are invariant. If you
	 have no Front-Cover Texts, write "no Front-Cover Texts"
	 instead of "Front-Cover Texts being LIST"; likewise for
	 Back-Cover Texts.'

So, IMHO, in order to be able to use SPDX without those optional parts of the spec, specially the Invariant Sections, an exception is needed. e. g. something like:

	'Permission is granted to copy, distribute and/or modify this document under
	 the terms of the GNU Free Documentation License; with no Invariant Sections,
	 no Front-Cover Texts and no Back-Cover Texts.'

swinslow:
This adds identifiers for the "-invariants" and "-no-invariants" versions of all of the GFDL licenses. It includes notes for each one explaining when they should be used.

It also cleans up and makes consistent the standard license headers for each version.

Fixes #686


swinslow:
Discussed on 2020-06-18 legal team call, @MarkAtwood will take a second look at this PR (thank you Mark!)

kestewart:
Why is this not being handled via an exception?    I would think that "WITH no-invariants" would be more appropriate.

swinslow:
@kestewart The format for the identifier here was based on the format from the requester at https://github.com/spdx/license-list-XML/issues/970#issuecomment-644578369 and https://github.com/spdx/license-list-XML/issues/686#issuecomment-644581190 (though I did pluralize it in this PR). I had thought also having the "-only" or "-or-later" immediately following the version number was clearer, so that it was more apparent what that phrase was referring to.

(Regarding not using WITH, this was considered -- discussed on calls last year (https://github.com/spdx/license-list-XML/issues/686#issuecomment-457289896) and one of the options mentioned in #970 (see item 3 on the list there), discussed on the legal team call and decided that because not all of the options are "exceptions" that it wouldn't work to add them as exceptions for use with the WITH operator.)

But I don't have a strong preference on the ordering of the sub-parts here. @mchehab given @kestewart's comment above, are you okay with me switching the identifier order so that it would be e.g. `GFDL-1.2-no-invariants-or-later` instead?

mchehab:
> @kestewart The format for the identifier here was based on the format from the requester at [#970 (comment)](https://github.com/spdx/license-list-XML/issues/970#issuecomment-644578369) and [#686 (comment)](https://github.com/spdx/license-list-XML/issues/686#issuecomment-644581190) (though I did pluralize it in this PR). I had thought also having the "-only" or "-or-later" immediately following the version number was clearer, so that it was more apparent what that phrase was referring to.
> 
> (Regarding not using WITH, this was considered -- discussed on calls last year ([#686 (comment)](https://github.com/spdx/license-list-XML/issues/686#issuecomment-457289896)) and one of the options mentioned in #970 (see item 3 on the list there), discussed on the legal team call and decided that because not all of the options are "exceptions" that it wouldn't work to add them as exceptions for use with the WITH operator.)
> 
> But I don't have a strong preference on the ordering of the sub-parts here. @mchehab given @kestewart's comment above, are you okay with me switching the identifier order so that it would be e.g. `GFDL-1.2-no-invariants-or-later` instead?

Yeah, `GFDL-1.2-no-invariants-or-later` works for me.

swinslow:
Great, thanks @mchehab and @kestewart. I'll rework the PR to switch the identifiers.

swinslow:
@kestewart PR update is done, see 258e461 -- grateful if you can re-review when you have a chance, then I'll merge this.

swinslow:
Discussed on 2020-07-02 legal team call, agreed to merge.

Issue #1049, opened 2020-06-17
Link to WTFPL is 404


vsoch:
Heyo! The link under "Other pages for this license" here https://spdx.org/licenses/WTFPL.html i 404. I think you could use any of the following:

 - https://spdx.org/licenses/WTFPL.html
 - www.wtfpl.net/

swinslow:
Hi @vsoch, thanks for filing this. I transferred it to the spdx/license-list-XML repo since this is where the team works on updates to the [SPDX License List](https://spdx.org/licenses).

For the existing URL cross-reference on the [WTFPL page](https://spdx.org/licenses/WTFPL.html), we leave old cross-references in place even if they are no longer valid links. That's because old code might have used those URLs to refer to the license, so it can still be useful for identifying what license applies to an old file.

Adding new cross-references is fine though, and it looks like http://www.wtfpl.net/about/ would be a useful additional cross-reference. If you wanted to submit a PR, that would be fantastic! The relevant file is at https://github.com/spdx/license-list-XML/blob/master/src/WTFPL.xml -- it should be self-explanatory where the additional cross-reference would go. Thanks for flagging this!

vsoch:
All set! https://github.com/spdx/license-list-XML/pull/1050

swinslow:
Fixed in #1050, thank you @vsoch!

PR #1050, opened 2020-06-17
Adding cross reference for WTFPL

vsoch:
As discussed in https://github.com/spdx/license-list-XML/issues/1049#issuecomment-645369852, this will add a cross reference to the currently functioning site for the WTFPL license. The previous link no longer works (404) but is maintained for older files that referenced it.


swinslow:
LGTM, thanks again @vsoch! Merging.

PR #1062, opened 2020-07-07
Typo in URL

camillem:
There is a redirect at OSI's level, but it's a bit tidier like this.
Cheers, 
Camille


swinslow:
Thank you @camillem! Although we usually don't update out-of-date URLs in cross-references, you're correct that this one appears to have been a typo (referencing 1.1 rather than 1.2). Thanks for the fix!

Issue #995, opened 2020-03-29
New license request: BUSL-1.1


adamretter:
**1.** License Name: Business Source License 1.1
**2.** Short identifier: BSL-1.1
**3.** License Author or steward: Adam Retter
**4.** Comments: This is a Source Available license, not an Open Source license. However it is in use and its adoption is growing. IMHO Source Available code needs even clearer licensing than Open Source as users can access the source code publicly (e.g. GitHub), but it has use restrictions.

Software using this license:

1. CockroachDB - https://github.com/cockroachdb/cockroach/blob/master/licenses/BSL.txt
2. Sentry.io - https://github.com/getsentry/sentry/blob/master/LICENSE
3. Materialize - https://github.com/MaterializeInc/materialize/blob/master/LICENSE
4. MariaDB Max Scale - https://github.com/mariadb-corporation/MaxScale/blob/2.4/LICENSE24.TXT
5. ZeroTier - https://github.com/zerotier/ZeroTierOne/blob/master/LICENSE.txt
6. FusionDB - https://fusiondb.com

Related blog posts about BSL:
1. http://monty-says.blogspot.com/2016/08/applying-business-source-licensing-bsl.html
2. https://blog.adamretter.org.uk/business-source-license-adoption/
**5.** Standard License Header: Copyright (c) ${date} ${organization}

Use of this software is governed by the Business Source License included
in the LICENSE file and at www.mariadb.com/bsl11.

Change Date: ${change-date}

On the date above, in accordance with the Business Source License, use
of this software will be governed by the ${change-license}.
**6.** URL: https://mariadb.com/bsl11/
**7.** OSI Status: Not Submitted

zvr:
The name "BSL" is already being used for Boost Software License: https://spdx.org/licenses/BSL-1.0.html

adamretter:
@zvr Ah okay, somehow I missed that! Perhaps we could use `BUSL-1.1`

jlovejoy:
I think MongoDB is the license steward. @adamretter - do you work for or represent them

swinslow:
Discussed on 2020-05-07 legal team call, agreed that the license steward should be looped into the discussion on whether this should be added (if that's different from the submitter of this issue).

adamretter:
@jlovejoy @swinslow I believe the "steward" if there is one, would be MariaDB (not MongoDB) - as they created the license.

jlovejoy:
we will work on contacting someone at MariaDB

[...]

PR #1074, opened 2020-07-22
Adding the Business Source License 1.1 (BUSL-1.1)

esplinr:
Definition files for the Business Source License 1.1 (BUSL-1.1)

See https://github.com/spdx/license-list-XML/issues/995

I hope that I did this right.

### DIFF
--- a/src/BUSL-1.1.xml
+++ b/src/BUSL-1.1.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <SPDXLicenseCollection xmlns="http://www.spdx.org/license">
   <license licenseId="BUSL-1.1" isOsiApproved="false"
-  name="Business Source License 1.1">
+  name="Business Source License 1.1" listVersionAdded="3.11">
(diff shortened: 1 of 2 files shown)


esplinr:
I misunderstood my local tests, but I now have all tests passing locally. Travis says that the tests succeeded, even though it says the tests are still running. It would be good to squash commits on merge (I don't remember if that's something I specify in GitHub.)

src/BUSL-1.1.xml
```diff
@@ -0,0 +1,126 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<SPDXLicenseCollection xmlns="http://www.spdx.org/license">
+  <license licenseId="BUSL-1.1" isOsiApproved="false"
+  name="Business Source License 1.1">
```
jlovejoy, line 4: need to add listVersionAdded="3.10"

swinslow: now would be "3.11" rather than "3.10"

PR #1079, opened 2020-07-26
Add ANTLR-PD-fallback

### LINKED ISSUE #1056: Review ANTLR-PD for potential omission
(copied from mailing list, see https://lists.spdx.org/g/Spdx-legal/message/2838)

= = = = =

Hello list,

I just found out that there is a deviation from
https://spdx.org/licenses/ANTLR-PD.html#licenseText to the linked text from
http://www.antlr2.org/license.html which contains the following language:

"In countries where the Public Domain status of the work may not be valid,
the author grants a copyright licence to the general public to deal in the
work without restriction and permission to sublicence derivates under the
terms of any (OSI approved) Open Source licence."

From the perspective from EU law this is an extremely important part since
it makes clear that a unrestricted license is intended if PD does not work.
This avoids (always disputable) interpretation of the PD text.

Is there any reason for the omission? Could the text be added?

Best regards,

Till

swinslow:
Fixes #1056 


swinslow:
Discussed on 2020-11-19 legal team call, agreed to add.

PR #1070, opened 2020-07-15
Improving the license workflow acceptance documentation

vmbrasseur:
Figured I'd send a WIP PR rather than bury the work over on my own account.

Lots more to do here. Please don't merge this, m'kay?

#1052

PR #1068, opened 2020-07-15
Add CC-BY-NC-ND-3.0-IGO license and test

vmbrasseur:
Gonna be real happy no longer to need to type that identifier, lemme tell ya.

Closes issue #767

PR #1080, opened 2020-07-26
Make minor fixes to CC-BY-NC-ND-3.0-IGO

swinslow:

### DIFF
--- a/src/CC-BY-NC-ND-3.0-IGO.xml
+++ b/src/CC-BY-NC-ND-3.0-IGO.xml
@@ -7,13 +7,13 @@
       </crossRefs>
     <text>
       <titleText>
-          <p>Creative Commons Attribution Non Commercial No Derivatives 3.0 IGO</p>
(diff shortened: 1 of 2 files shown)


swinslow:
Hi @vmbrasseur -- I think I led you down the wrong path for the CC-...-IGO license (my apologies!)

In my [comment](https://github.com/spdx/license-list-XML/pull/1068#discussion_r455882594) about the "title" of the license, I'd meant to just change the `name` property as it appears in the `license` tag. But for the actual `titleText` in the license text itself, I think we'll want to stick with the title as it appears on the license steward's site.

This PR changes that back, and it also replaces smart quotes in two spots with plain quotes (per discussion [here](https://github.com/spdx/license-list-XML/pull/1070#discussion_r460542197)). I know that the upstream original uses smart quotes, but it appears inadvertent since all the rest of the instances are plain quotes. So I'm actually not sure what the "right" answer here should be. Let's plan to take a couple of minutes to discuss with @jlovejoy on the call this week. Thanks!

swinslow:
discussed on call today, good to merge

PR #1081, opened 2020-07-29
Fix remaining items in new license workflow

### LINKED ISSUE #1052: Improving the license workflow acceptance documentation
Notes from working on #993.

The doc in question: https://github.com/spdx/license-list-XML/blob/master/DOCS/new-license-workflow.md#if-license-is-accepted

Reviewing 'If license is accepted'

- [x] Link `SPDX Online tools` in 7.i.
    * https://spdxtools.sourceauditor.com/
- [ ] 7.i assumes that the license exists in the License Requests list
    * In the case of #767, the CC-BY-NC-ND-3.0-IGO license isn't in the list.
- [x] 7.i.a could use a link to a "wrapped text version" example.
- [x] 7.i "edit XML" requires a login
    * Mention this in the doc
    * Only signin option is 'Login with GitHub'
- [x] 7.i "edit XML" is a button at the bottom of the license text (License Requests > click license > scroll to the bottom
    * This is NOT the 'License XML Editor' link from the top of the page
- [x] 7.i.b should include directions for where `listVersionAdded=` should be added, and perhaps an example same.
- [ ] 7.i.c says "standard license header" as though it's common knowledge. It's not. An example is required.
    * I don't know where to find this or to what it refers.
- [ ] 7.i.d has a broken link to an Overview page.
- [x] 7.i.f should link to a license to use as an example. It also should explain "the bullet or list tags" better. Where would these tags be found?
- [x] 7.ii says to run the tests locally, but the changes are happening using the online tool. There has to be a "copy your changes locally, then…" step in there.
    * Also, at no point was the user directed to clone the repo locally. Probably should ask them to do that? But if they're doing all the work in the online tool, maybe just drop the local test step?

swinslow:
This finishes off the remaining not-yet-fixed checkbox items listed in #1052.

Fixes #1052

### DIFF
--- a/DOCS/new-license-workflow.md
+++ b/DOCS/new-license-workflow.md
@@ -4,7 +4,7 @@
 
 This document provides guidance and checklists for SPDX legal team members who are assigned to shepherd a new license request.
 
-The instructions here assume the requestor has already submitted the new license/exception request via the [SPDX online tools](http://13.57.134.254/app/).
+The instructions here assume the requestor has already submitted the new license/exception request via the [SPDX online tools](http://13.57.134.254/app/). Note that it is also possible for license requests to be submitted directly as Issues in the GitHub repo; when this occurs, they will not appear on the SPDX online tools page.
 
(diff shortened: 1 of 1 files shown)


swinslow:
discussed today, good to merge

PR #1090, opened 2020-08-25
Correcting some items' order in SugarCRM-1.1.3.xml

camillem:
Hi, 
Order of 1.9-A and 1.9-B had been reversed. 2.2-a was also at the wrong end of the list.

### DIFF
--- a/src/SugarCRM-1.1.3.xml
+++ b/src/SugarCRM-1.1.3.xml
@@ -72,14 +72,14 @@
                          Original Code or any previous Modifications. When Covered Code is released as a series of
                          files, a Modification is:
                         <list>
-                            <item>
-                                <bullet>B.</bullet>
-                                Any new file that contains any part of the Original Code or previous Modifications.
-                            </item>
-                            <item>
+                           <item>
                                 <bullet>A.</bullet>
                                 Any addition to or deletion from the contents of a file containing Original Code or previous
                                  Modifications.
+                           </item>
+                           <item>
+                                <bullet>B.</bullet>
+                                Any new file that contains any part of the Original Code or previous Modifications.
                             </item>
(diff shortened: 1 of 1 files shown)


jlovejoy:
this needs a closer look, including as to why it's failing on the build

goneall:
Here's the error from the validation:

```
Test for license SugarCRM-1.1.3 failed: Test for license ID SugarCRM-1.1.3 failed due to difference found Normal text of license does not match starting at line #31 column #40 "any" when comparing to template text " Any addition to or deletion from the contents of a file containing Original Code or previous Modifi..."
```

From looking at the changes, it looks like the license XML was changed but there were no matching changes to the test file https://github.com/spdx/license-list-XML/blob/master/test/simpleTestForGenerator/SugarCRM-1.1.3.txt

goneall:
Just another note - if there are license texts similar to those in the current test file, we'll need to add variable text markers to allow both versions to match.

camillem:
Sorry, I was not aware of the tests. Depending on what's more convenient for you, I can create a new clean PR with the tests also corrected or add commits to this one. 
And actually, it seems that section 1.0.1 is also missing in current SDPX version as compared to https://web.archive.org/web/20110722045536/http://www.sugarcrm.com/crm/SPL

goneall:
> Depending on what's more convenient for you, I can create a new clean PR with the tests also corrected or add commits to this one.

Either approach is fine

jlovejoy:
oddly, this is how it appears in the standard text: https://www.sugarcrm.com/legal/sugarcrm-public-license/

(although we do need to update the URL.

Given that is how it appears in the wild, I think we should just leave it.  Sorry this took so long... !!!

Issue #1118, opened 2020-10-07
Update license request: DOC


jwillemsen:
The DOC license (see https://spdx.org/licenses/DOC.html) has a broken link, http://www.cs.wustl.edu/~schmidt/ACE-copying.html should be updated to http://www.dre.vanderbilt.edu/~schmidt/ACE-copying.html

swinslow:
Fixed by #1119

PR #1119, opened 2020-10-07
Add working DOC license url

jwillemsen:


jwillemsen:
Fixes #1118

vmbrasseur:
Thanks for the PR, @jwillemsen!

We have a policy never to remove/replace crossref URLs, instead preferring to add new ones. From [license-fields.md](https://github.com/spdx/license-list-XML/blob/master/DOCS/license-fields.md):

> Note that the source URL may refer to an original URL for the license which is no longer active. We don't remove inactive URLs. New or best available URLs may be added.

Could you please add the new URL rather than replace the old one?

swinslow:
Thanks @jwillemsen -- and thanks also @vmbrasseur, was just about to say the same thing and you beat me to it.  :)  Merging now.

Issue #1057, opened 2020-06-30
Two versions of LGPL-2.1


MarkAtwood:

emmanueltournier:
I'll be looking at this one.

emmanueltournier:
@swinslow just to clarify, we should have something like this? :

<optional>&lt;</optional>one line to give the library's name and an idea of what it does.<optional>&gt;</optional>
<optional>&lt;</optional>one line to give the library's name and a brief idea of what it does.<optional>&gt;</optional>

swinslow:
Hi @emmanueltournier, this one will involve creating an `<alt>` tag with a match option for the two different formats. You can see examples of this, for instance, at https://github.com/spdx/license-list-XML/blob/af50167c006737d07f97c4aec779cc594b36a0b1/src/Apache-1.1.xml#L43

Let's plan to discuss during today's spdx-legal team call (if you're able to join) and I can show specifically how this would work. Thanks!

emmanueltournier:
Hi @swinslow. I was planning to make the modification below. Let me know if you want to discuss before our next meeting. Thanks.

`<p>
	<optional>&lt;</optional>one line to give the library's name
	and <alt match="an idea">a brief idea</alt> of what it does.<optional>&gt;</optional>
        <br></br>
        Copyright (C)
        <optional>&lt;</optional>year<optional>&gt;</optional>
        <optional>&lt;</optional>name of author<optional>&gt;</optional>
 </p>`

swinslow:
```
<alt match="an idea|a brief idea">
```

PR #1132, opened 2020-11-05
Update LGPL-2.1-only.xml

emmanueltournier:
@swinslow  Fixing Issue "Two versions of LGPL-2.1 #1057"  - 
Added alternate language "a brief idea" | "an idea"

### DIFF
--- a/src/LGPL-2.1-only.xml
+++ b/src/LGPL-2.1-only.xml
@@ -618,7 +618,7 @@
       </p>
       <p>
 	<optional>&lt;</optional>one line to give the library's name
-	and <alt match="name" name="a brief idea">an idea</alt> of what it does.<optional>&gt;</optional>
+	and <alt match="an idea|a brief idea" name="ideaClause">an idea</alt> of what it does.<optional>&gt;</optional>
         <br></br>
         Copyright (C)
         <optional>&lt;</optional>year<optional>&gt;</optional>
--- a/src/LGPL-2.1-only.xml
+++ b/src/LGPL-2.1-only.xml
@@ -618,7 +618,7 @@
       </p>
       <p>
 	<optional>&lt;</optional>one line to give the library's name
-	and an idea of what it does.<optional>&gt;</optional>
+	and <alt match="name" name="a brief idea">an idea</alt> of what it does.<optional>&gt;</optional>
         <br></br>
         Copyright (C)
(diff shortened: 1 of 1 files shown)


swinslow:
Hi @emmanueltournier, thanks for this! I made a couple of tweaks, and wanted to note here why.

For the `match` property, I changed it to read `match="an idea|a brief idea"`. The `match` property is the one that specifies the regular expression for the different options that can apply. The pipe `|` character between the two options signals that either one is a correct match.

For the `name` property, I admit that I am not entirely sure how this is used in practice, but I believe it can enable tools that are applying the templates to be able to refer to and extract the exact text that was matched against for this field. So I usually see these given unique names if there are multiple `alt` matches within a document.

One other thing -- after the tests pass, we'll want to make the same edit to [`LGPL-2.1-or-later.xml`](https://github.com/spdx/license-list-XML/blob/master/src/LGPL-2.1-or-later.xml), [`LGPL-2.1.xml`](https://github.com/spdx/license-list-XML/blob/master/src/LGPL-2.1.xml) and [`LGPL-2.1+.xml`](https://github.com/spdx/license-list-XML/blob/master/src/LGPL-2.1%2B.xml). For historical reasons, the GPL family of licenses include many different entries on the list. If you get a chance to make this same edit in these files, that's great; if not, I'll make the change before the 3.11 release goes out. Thanks!

emmanueltournier:
Thanks @swinslow . I was also quite confused by the "name" property becasue in this case it is not used. It looks like your modification passed the checks so I'll go ahead with the other LGPL-2.1 licenses.

swinslow:
Thanks @emmanueltournier! If you get a chance to add the others today, that's great; otherwise I will plan to add it myself tomorrow morning before I push the 3.11 release.

swinslow:
Hi @emmanueltournier, I see now that `LGPL-2.1-or-later` and `LGPL-2.1+` weirdly have different markup for that line, which already handles the alt match (in fact it matches against anything at all in that line).

I don't recall why it is different from LGPL-2.1-only, but I'm not going to change it now  :)  So I'll just submit a separate PR for LGPL-2.1 to mirror what you did for LGPL-2.1-only. Thanks again!

emmanueltournier:
Thanks @swinslow ! Sorry, I thought I would have had the time to fix the other licenses yesterday.

PR #1137, opened 2020-11-12
Add MIT Open Group Variant License

### LINKED ISSUE #1026: Open Group License
Submitting issue on behalf of request sent to mailing list, see https://lists.spdx.org/g/Spdx-legal/message/2806:

= = = = =

Hi,

while adding identifiers to licenses used in the BSP (note: i am a
legal trainee, so wording may not be technical perfect), i stumbled
upon the The Open Group License, which does not have an identifier yet.
It is used in: xorg-app-iceauth, xorg-app-rgb, xorg-app-xauth, xorg-
app-xdm, xorg-app-xinit, xorg-app-xprop, xorg-app-xrdb, xorg-app-xset,
xorg-app-xsetroot, xorg-app-xvinfo (and maybe some more).
The wording is consistent the following (copyright holder is always The
Open Group):

COPYING

- - -

Copyright <yyyy, yyyy> <Copyright Holder>

Permission to use, copy, modify, distribute, and sell this software and
its documentation for any purpose is hereby granted without fee,
provided that the above copyright notice appear in all copies and that
both that copyright notice and this permission notice appear in
supporting documentation.

The above copyright notice and this permission notice shall be included
in all copies or substantial portions of the Software.

THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
IN NO EVENT SHALL <Copyright Holder> BE LIABLE FOR ANY CLAIM, DAMAGES
OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR
OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR
THE USE OR OTHER DEALINGS IN THE SOFTWARE.

Except as contained in this notice, the name of <Copyright Holder>
shall not be used in advertising or otherwise to promote the sale, use
or other dealings in this Software without prior written authorization
from <Copyright Holder>.

- - -

The first paragraph is different from X11-License, therefore i do not
want to use the X11-identifier:
> Permission to use, copy, modify, distribute,
as in curl; no "including without limitation" like X11
> and sell this software and its documentation
software and documentation are not together defined as "the software"
and therefore more divided than in X11-License
> for any purpose is hereby granted without fee, provided
> that the above copyright notice appear in all copies and that both
> that copyright notice and this permission notice appear in supporting
> documentation.
X11 does not require appearance in documentation, but in substantial
parts of software.

I am hoping that you can add the License to the license list so it may
gain an identifier.
Thank you!

Felicitas Jung

RishabhBhatnagar:
I'm not sure about whitespaces (newlines). 
I'm sure there might be many changes. Do let me know how did I fare.

Fixes #1026 .

### DIFF
--- a/src/MIT-open-group.xml
+++ b/src/MIT-open-group.xml
@@ -1,0 +1,33 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<SPDXLicenseCollection xmlns="http://www.spdx.org/license">
+    <license isOsiApproved="false" licenseId="MIT-open-group"
+             name="MIT Open Group variant">
+        <text>
+            <copyrightText>
+                <p>Copyright &lt;yyyy, yyyy&gt;</p>
+            </copyrightText>
+            <p>Permission to use, copy, modify, distribute, and sell this
+                software and its documentation for any purpose is hereby granted
+                without fee, provided that the above copyright notice appear in
+                all copies and that both that copyright notice and this
+                permission notice appear in supporting documentation.
+            </p>
+            <p>The above copyright notice and this permission notice shall be
+                included in all copies or substantial portions of the Software.
+            </p>
+            <p>THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+                EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES
+                OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+                NONINFRINGEMENT. IN NO EVENT SHALL BE LIABLE FOR ANY CLAIM,
+                DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+                TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH
+                THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+            </p>
+            <p>Except as contained in this notice, the name of shall not be used
+                in advertising or otherwise to promote the sale, use or other
+                dealings in this Software without prior written authorization
+                from .
+            </p>
+        </text>
+    </license>
+</SPDXLicenseCollection>
--- a/src/MIT-open-group.xml
+++ b/src/MIT-open-group.xml
@@ -10,7 +10,7 @@
         </crossRefs>
         <text>
             <copyrightText>
-                <p>Copyright &lt;yyyy, yyyy&gt; <alt match=".+" name="name1">The Open Group</alt></p>
+                <p>Copyright &lt;yyyy, yyyy&gt; The Open Group</p>
             </copyrightText>
             <p>Permission to use, copy, modify, distribute, and sell this
                 software and its documentation for any purpose is hereby granted
@@ -29,10 +29,10 @@
                 TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH
                 THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
             </p>
-            <p>Except as contained in this notice, the name of <alt match=".+" name="name2"> The Open Group</alt>
+            <p>Except as contained in this notice, the name of <alt match=".+" name="name1"> The Open Group</alt>
                 shall not be used in advertising or otherwise to promote the
                 sale, use or other dealings in this Software without prior
-                written authorization from <alt match=".+" name="name3">The Open Group</alt>.
+                written authorization from <alt match=".+" name="name2">The Open Group</alt>.
             </p>
         </text>
     </license>
--- a/src/MIT-open-group.xml
+++ b/src/MIT-open-group.xml
@@ -1,10 +1,16 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <SPDXLicenseCollection xmlns="http://www.spdx.org/license">
     <license isOsiApproved="false" licenseId="MIT-open-group"
-             name="MIT Open Group variant">
+             name="MIT Open Group variant" listVersionAdded="3.11">
+        <crossRefs>
+            <crossRef>https://gitlab.freedesktop.org/xorg/app/iceauth/-/blob/master/COPYING</crossRef>
+            <crossRef>https://gitlab.freedesktop.org/xorg/app/xvinfo/-/blob/master/COPYING</crossRef>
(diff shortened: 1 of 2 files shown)


swinslow:
Hi Rishabh, thanks for this! I am reviewing now and will add some comments.

src/MIT-open-group.xml
```diff
@@ -0,0 +1,33 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<SPDXLicenseCollection xmlns="http://www.spdx.org/license">
+    <license isOsiApproved="false" licenseId="MIT-open-group"
+             name="MIT Open Group variant">
```
swinslow, line 4: As an additional property here between the `name="MIT Open Group variant"` and the `>` closing tag, you'll want to add the property: `listVersionAdded="3.11"`

This is to document that this license was added in the version 3.11 release of the license list.

src/MIT-open-group.xml
```diff
@@ -0,0 +1,33 @@
...
+            </p>
+            <p>THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+                EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES
+                OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+                NONINFRINGEMENT. IN NO EVENT SHALL BE LIABLE FOR ANY CLAIM,
+                DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+                TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH
+                THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+            </p>
+            <p>Except as contained in this notice, the name of shall not be used
```
swinslow, line 26: In this sentence, there are two places where there are words missing from the example submitted by the submitter. You can see that there are missing words between: `the name of ____ shall not be used` and also at the end: `without prior written authorization from ____.`

Looking at an example, for instance at https://gitlab.freedesktop.org/xorg/app/iceauth/-/blob/master/COPYING, you can see that "The Open Group" is used in these locations.

Ideally we don't want to just hard-code "The Open Group" into the template, because then it wouldn't match if somebody else puts their own name there. But we also don't want to leave it empty, because then it will not match under the matching guidelines.

What we can do instead is to insert "The Open Group" but include markup for an alt match. This would mean, in place of the first blank, put the following:

`<alt match=".+" name="name1">The Open Group</alt>`

and in place of the second blank at the end of the sentence, put the following (only the "name" property is changed):

`<alt match=".+" name="name2">The Open Group</alt>`

src/MIT-open-group.xml
```diff
@@ -0,0 +1,33 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<SPDXLicenseCollection xmlns="http://www.spdx.org/license">
+    <license isOsiApproved="false" licenseId="MIT-open-group"
+             name="MIT Open Group variant">
+        <text>
```
swinslow, line 5: Before the `<text>` block, I would add a `<crossRefs>` list with at least one entry with a URL showing use of this license in the wild. Here is an example of what this can look like -- you can add this one as-is:

```
<crossRefs>
  <crossRef>https://gitlab.freedesktop.org/xorg/app/iceauth/-/blob/master/COPYING</crossRef>
</crossRefs>
```

test/simpleTestForGenerator/MIT-open-group.txt
```diff
@@ -0,0 +1,23 @@
...
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
+OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+IN NO EVENT SHALL BE LIABLE FOR ANY CLAIM, DAMAGES
+OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR
+OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR
+THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+Except as contained in this notice, the name of
```
swinslow, line 20: Here in the test file, you will also want to add `The Open Group` in place of the two blank bits. Note that here in the test file, it will __not__ get the alt tags around it -- just the literal text `The Open Group`.

swinslow (changes requested):
Hi @RishabhBhatnagar, thanks for this -- great job on your first one! I added a few review comments. If you can make these changes prior to the legal team call, that's fantastic -- if not, no worries, we can update during or after the call. Feel free to respond in comments if anything is unclear.

src/MIT-open-group.xml
RishabhBhatnagar: Thank You for taking out time to explain it in a detailed manner.
Having seen `alt` in other licenses, I quite didn't understand the exact use-case.

This cleared my confusion. Thanks!

src/MIT-open-group.xml
RishabhBhatnagar: Upon searching for the entire license using some search engines, I couldn't find any other matches for the license which were exactly the same (except for the alt part of course).

Will it work if I don't add any other cross-references?

test/simpleTestForGenerator/MIT-open-group.txt
RishabhBhatnagar: Got it!

RishabhBhatnagar: Do I have to replace <yyyy, yyyy> too?

swinslow (approved):

swinslow:
Hi @RishabhBhatnagar, thanks for the updated edits. The reason this failed is not obvious, so no worries 😄 

Under the SPDX matching guidelines, the `<copyrightText>` tag is interpreted such that it can contain anything at all. Because of that, I believe the XML schema is not intended to contain any `<alt>` tags inside of a `<copyrightText>` section.

To fix this, I've just removed that first `<alt>` tag section. I believe it should build correctly now, though the Travis tests have been taking a very long time to initialize. After they run I'll go ahead and merge this. Thanks again for preparing this one!

PR #1143, opened 2020-11-18
Add HTML Tidy License

johnmhoran:
Add an `.xml` and `.txt` file for the new `HTML Tidy License`.

### DIFF
--- a/src/HTMLTIDY.xml
+++ b/src/HTMLTIDY.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <SPDXLicenseCollection xmlns="http://www.spdx.org/license">
-   <license isOsiApproved="false" licenseId="HTMLTIDY" name="HTML Tidy License">
+   <license isOsiApproved="false" licenseId="HTMLTIDY" name="HTML Tidy License" listVersionAdded="3.11">
       <crossRefs>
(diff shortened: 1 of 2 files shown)


johnmhoran:
@swinslow @jlovejoy Neglected to add you as reviewers when I opened this PR -- looks like the 2 Travis CI jobs finally started and succeeded.  Please let me know if you have any comments, edit requests etc.

src/HTMLTIDY.xml
```diff
@@ -0,0 +1,44 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<SPDXLicenseCollection xmlns="http://www.spdx.org/license">
+   <license isOsiApproved="false" licenseId="HTMLTIDY" name="HTML Tidy License">
```
swinslow, line 3: @johnmhoran, can you update this `license` tag to include `listVersionAdded="3.11"` ?

swinslow:
Thanks @johnmhoran! This looks great, I think the only update needed is to add the `listVersionAdded` property that I mentioned in the comment above. After that it should be all set to merge.

src/HTMLTIDY.xml
johnmhoran: Will do @swinslow -- thank you.  👍

swinslow:
Thanks @johnmhoran! I'll merge after the tests pass.

Issue #1146, opened 2020-11-22
GPL-3.0-or-later template has extra spaces


alandtse:
This appears to be an issue converting the xml to the template text.

The template for GPL-3.0-or-later includes extra spaces around the alternatives (philosophy/licenses) for the url:
<https://www.gnu.org/licenses/why-not-lgpl.html>.

The source file looks fine:
https://github.com/spdx/license-list-XML/blob/master/src/GPL-3.0-or-later.xml#L838

But the generated template has the extra spaces:
https://github.com/spdx/license-list-data/blob/master/template/GPL-3.0-or-later.template.txt#L251

You can also see it on the main html page.
https://spdx.org/licenses/GPL-3.0-or-later.html

goneall:
You can add a `spacing="none"` attribute value to the `<optional>` element (e.g. `http<optional spacing="none">s</optional>`).

goneall:
Issue #1076 is also requesting this feature.

alandtse:
I'm not sure what the attribute does for downstream users of the generated template. They don't see that tag at all so they can't manipulate it. It also makes the template incorrect as it won't match the canonical license. The issue is whatever is parsing the xml file and adding in spaces to the generated files.

goneall:
> I'm not sure what the attribute does for downstream users of the generated template. 

It is a hint to the licenseListPublisher which generates the template.  The downstream will have the correct spacing based on the hints.

The problem is that when an element is added in the XML, you can not reliably determine what the spacing should be.  I fear removing the space for all occurrences would re-introduce a previous formatting bug.

For reference, here is the code that interprets the XML and generates the template: https://github.com/spdx/LicenseListPublisher/blob/23ff8ef1a8bf5aff20944b2e2571c28f44056f14/src/org/spdx/licensexml/LicenseXmlHelper.java#L502

There is a BUNCH of code in there to handle spaces and newlines which was developed somewhat through trial and error.  If you have any suggestions on a better approach, please add suggestions to the [LicenseListPublisher issues list](https://github.com/spdx/LicenseListPublisher/issues).

I'll add an issue soon to the LicenseListPublisher to track resolving the spacing issue described here.

goneall:
Added issue https://github.com/spdx/LicenseListPublisher/issues/87 to track the tool changes.

alandtse:
[...]

PR #1148, opened 2020-11-23
Change alt pattern for GPL-3.0-or-later

goneall:
This change works around the formatting issues described in issue #1146 

Note - there are very likely other licenses that the same issue.

The approach to resolving the formatting is to make sure the `<alt`... element is on a word boundary where it would be natural to have a space.


swinslow:
Thanks @goneall, appreciate the help with fixing this one. Merging now.

PR #1152, opened 2020-11-25
Add release notes for 3.11

swinslow:
Signed-off-by: Steve Winslow <steve@swinslow.net>

PR #1153, opened 2020-11-28
Add missing item for 3.11 release notes

swinslow:
The ANTLR-PD-fallback license was inadvertently omitted
from the list of added licenses in the release notes for 3.11.
This commit reinserts it.


waldyrious:
@swinslow while you're at it, WDYT about also making the change suggested [here](https://github.com/spdx/license-list-XML/pull/1152#discussion_r530556395)?

swinslow:
Great, thanks @waldyrious, I've added that change.

PR #1160, opened 2020-12-19
Update contributing recommendations for the test license text

goneall:
Add recommendation that plain text test files should match the text and formatting of the original license.

The test license text is used for test and is also used as the original formatting of the license text.  This helps resolve several issues where users of the license list would like to have a reference to the license text which include the original line breaks and formatting.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -31,7 +31,7 @@ The SPDX Legal Team appreciates proposals for new free and open source licenses 
 1. The SPDX Legal Team will discuss any submissions for new licenses or exceptions via comments in the Github issue. Please follow the comments and respond accordingly if there are questions or additional information requested.
 2. New licenses may be approved if 3 SPDX-legal team members (2 lawyers, 1 non-lawyer) sign-off that the license is acceptable AND there is no objection raised from the greater SPDX-legal community within the Github issue comments. If there are objections, then the issue will be labelled "discuss on legal call" and will be discussed on an upcoming bi-weekly call.
 3. Issues will be labelled either "new license/exception: Accepted" or "new license/exception: Not Accepted" as appropriate with an explanation and the Issue closed for the latter case.
-4. If accepted, the new license or exception will be need to be prepared in the proper XML format and plain text test file via a Pull Request and reviewed as appropriate.This is ideally done by the license submitter, with help from experienced members of the SPDX legal team, as needed.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -31,7 +31,7 @@ The SPDX Legal Team appreciates proposals for new free and open source licenses
...
 3. Issues will be labelled either "new license/exception: Accepted" or "new license/exception: Not Accepted" as appropriate with an explanation and the Issue closed for the latter case.
-4. If accepted, the new license or exception will be need to be prepared in the proper XML format and plain text test file via a Pull Request and reviewed as appropriate.This is ideally done by the license submitter, with help from experienced members of the SPDX legal team, as needed.
+4. If accepted, the new license or exception will be need to be prepared in the proper XML format and plain text test file via a Pull Request and reviewed as appropriate.  The plain text test file should match the text and formatting of the original license.  The Pull Request is ideally done by the license submitter, with help from experienced members of the SPDX legal team, as needed.
```
sschuberth: Some nits:
- The commit message has a typo in "reccomendation" (and the title is too long to common Git rules; text should be continued in the body).
- New sentences start with a double-space after `.` here. I know some people prefer that style, but it is not inline with with the style already used in this file.

goneall: Thanks for the detailed review.  I fixed the title and double space after `.`.

Let's squash the commits when we do the Github merge to fix the bad spelling in the commit message.

sschuberth: > Let's squash the commits when we do the Github merge to fix the bad spelling in the commit message.

How does that actually fix the spelling? Doesn't GitHub simply create a squashed message quoting all original messages? IMO the right way to fix this would be to amend the commit in this PR (and force-push a new version of this branch under the same name).

goneall: I squashed the commits locally and pushed back with the corrected spelling.

waldyrious: > > Let's squash the commits when we do the Github merge to fix the bad spelling in the commit message.
>
> How does that actually fix the spelling? Doesn't GitHub simply create a squashed message quoting all original messages?

GitHub actually allows editing the commit message when doing a squash merge.

goneall:
@jlovejoy @swinslow While you're cleaning up the other issues and PR's can you review and, if OK, merge this PR?

swinslow (approved):

swinslow:
Thanks @goneall!

PR #1165, opened 2020-12-31
Propose adding Ethical Source Definition

Mentors4EDU:
I am proposing the addition of the Ethical Source Definition as well within the inclusion guidelines. 
See [here](https://ethicalsource.dev/definition/). This hopefully broadens the range in regards to what is acceptable but still within the range of being open source.
This is just a suggestion and unrelated to my license pull requests.


jlovejoy:
The license inclusion guidelines have only been updated once since 2013  (see Historical Background section at https://github.com/spdx/license-list-XML/blob/master/DOCS/license-inclusion-principles.md). The recent update went through substantial discussions on our calls and mailing lists, and then review before being adopted. Further changes would need the same level of community involvement and review 
 :). Adding this definition was not included as per these discussions.

Also note, in the SPDX spec, you can always use a LicenseRef- string (see section 6 of the current spec) for licenses not on the SPDX License List.

PR #1167, opened 2021-01-06
Show the new (and source) URL

JJ:
Coming here from  spdx/LicenseListPublisher#91
The URL shown has been redirected, and it's been marked as unexistent.


swinslow:
Hi JJ, thanks for this.

For cross-references on existing licenses, we do not remove old cross-references even if they are no longer valid (this is mentioned [here](https://github.com/spdx/license-list-XML/blob/master/DOCS/license-fields.md) under section C, and in other issues when it comes up from time to time). That's because old URLs, even if inactive, can still be useful as pointers to the license ID when they appear in comments in source code.

For your PR, if you want to change it so that it adds the updated URL (while also retaining the old one), that would be great and very welcome. Thanks!

PR #1179, opened 2021-02-04
updates to MPL-2.0-nocopyleft-exception-

jlovejoy:
update Notes re: Exhibit B to make more sense

add standardLicenseHeader tag to Exhibit A and B (it was missing entirely)

### DIFF
--- a/src/MPL-2.0-no-copyleft-exception.xml
+++ b/src/MPL-2.0-no-copyleft-exception.xml
@@ -404,15 +404,19 @@
       </list>
       <optional>
         <p>Exhibit A - Source Code Form License Notice</p>
+         <standardLicenseHeader>
         <p>This Source Code Form is subject to the terms of the Mozilla Public License, v. 2.0. If a copy of the MPL
          was not distributed with this file, You can obtain one at http://mozilla.org/MPL/2.0/.</p>
+         </standardLicenseHeader>
         <p>If it is not possible or desirable to put the notice in a particular file, then You may include the
          notice in a location (such as a LICENSE file in a relevant directory) where a recipient would be
          likely to look for such a notice.</p>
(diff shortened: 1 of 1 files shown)


jlovejoy:
@goneall or @zvr or @swinslow - I'm not sure I did the standardLicenseHeader right - there are two bits to be included (text in ExhibitA and text in ExhibitB) - if I use two sets of standardLicenseHeader will it put that text together in the field? Or should I create one standardLicenseHeader tag with it together, outside from the exhibits?

goneall:
@jlovejoy I think we only support 1 standardLicenseHeader, but it looks like it passed verification which surprises me.  I need to verify my understanding.  I'll look into it and add a comment this afternoon.

goneall:
@jlovejoy I just checked and it looks like the different standardLicenseHeaders all get concatenated into a single standardLicenseHeader.  So, this should be fine.

Here's the resulting text.  I noticed a possible issue where an extra linefeed is added between the 2 concatenated licenseheaders:

```
This Source Code Form is subject to the terms of the Mozilla Public License, v. 2.0. If a copy of the MPL was not distributed with this file, You can obtain one at http://mozilla.org/MPL/2.0/.


This Source Code Form is "Incompatible With Secondary Licenses", as defined by the Mozilla Public License, v. 2.0.
```

jlovejoy:
Thanks @goneall ! I don't think the extra line is a major issue (and I'm not sure how to fix it) - agree?

goneall:
> I don't think the extra line is a major issue (and I'm not sure how to fix it) - agree?

Agree - I think this should work fine.

swinslow:
LGTM, thanks @jlovejoy and @goneall! Merging now.

PR #1181, opened 2021-02-05
update full names for ODC licenses

jlovejoy:
ODC should be Open Data Commons in full name

add updated URL


swinslow:
LGTM, thanks @jlovejoy! Merging now.

PR #1183, opened 2021-02-05
Add DRL-1.0

jlovejoy:
add new license

Fixes #1183


swinslow:
@jlovejoy I think this one failed because the test text file is missing the `.txt` at the end, so it couldn't find the test file...

swinslow:
I renamed the file -- we'll see if the tests pass now.

swinslow:
Thanks @jlovejoy! Tests are passing now, will merge.

PR #1188, opened 2021-02-14
Add NAIST-2003

swinslow:
Fixes #1136

Signed-off-by: Steve Winslow <steve@swinslow.net>

PR #1199, opened 2021-02-19
Moving to Github Actions

goneall:
This pull request builds on the work from @iamwillbar to move us to Github actions.

I am still not able to get this to work, but I did make progress on 2 fronts:
- Modify the Makefile to only run the license list publisher on the files changed - this is an OOM performance improvement for this use
- Change the environment variable for the GITHUB token to comply with the GitHub recommended environment variable names


goneall:
This PR no longer hangs, but it does fail with the following message:

```
fatal: could not read Password for 'https://***@github.com': No such device or address
```

goneall:
@swinslow @iamwillbar The verification works fine, it is only the publish that is an issue.  I'm wondering if I should remove the testing parameters and merge this so we can at least get the testing working.

Let me know your thoughts.

goneall:
Reporting a bit of progress - I generated a new secret and got past the error.  Fixing a few new problems - hoping to have something committed tomorrow.

goneall:
I got the actions to work on my local repository.

Unfortunately, the improvement to only process the changed files didn't work.

I made changes based on https://github.community/t/find-what-files-changed-in-a-pushed-commit/17037

When a list the changed files using `git diff-tree --no-commit-id --name-only -r $(SHA_FOR_GITHUB)`, no files were listed.

There is a SHA being passed in, but it doesn't seem to reflect the changes per the community post.

@iamwillbar let me know if you have any suggestions on this - if we could get this to work, it would speed up most commits by an order of magnitude

goneall:
Even though I couldn't get the single file improvement to work, I think this will re-enable the CI using github actions.

@iamwillbar @swinslow If you could do a review of the PR, I can merge tomorrow if it checks out.

swinslow (commented):
Thanks @goneall! I'm not able to weigh in substantively on the changes to the CI configuration, but I am fine with you merging this so that we can try it out in practice.

Once it merges, I can try submitting an intentionally-failing PR now to see if it gets blocked.

And assuming it does, then we can merge one of the pending PRs that has already been tested as working (e.g. #1188) so that we can confirm that the publisher is working.

iamwillbar:
@goneall great to see you were able to make progress here. I'll be able to review this tonight.

amCap1712:
Hi!, @goneall I think `git diff-tree` may not be working as desired due because by default the checkout action retrieves only the latest commit info. I tried to test this in my repo but I do not know from the output of the makefile whether it worked or not. Setting `fetch-depth` to 0 will fetch entire git history and data during the clone,
```
- uses: actions/checkout@v2
  with:
     fetch-depth: 0
```
This is what we do in the setup of `spdx-maven-plugin` since SonarLint checks  requires detailed git history to annotate with git blame.

goneall:
Thanks @amCap1712 for the review.  I'll look into the `fetch-depth: 0` option.  I did some experiments on my local machine.  If I set the commit env. variable to a valid commit it works - but only for that commit.  If that single commit only changes a single file, it behaves as expected.  When I merge a PR with a single file changed, it doesn't seem to pick up the right commit.

BTW - I would really like to move away from the Makefile - perhaps using native Github actions.  @iamwillbar @amCap1712 do you think this is feasible?

I'll do some more experimenting early next week.

goneall:
I think I found a solution: https://github.community/t/check-pushed-file-changes-with-git-diff-tree-in-github-actions/17220/10

I got it to list the changed files in an action step.  I just need to figure out how to export and import that information to the makefile which shouldn't be too difficult.

goneall:
I just realized that with the performance improvements on single file commits, we will loose the comparison against the other licenses in the repo.

I filed [this issue](https://github.com/spdx/LicenseListPublisher/issues/95) to resolve in the LicenseListPublisher.

goneall:
Changing back to a WIP - Testing out a solution to the single file fix

goneall:
Some additional issues with this PR:

- The publish is not optimized for a single file.  It currently purges the entire directory before running the tool to handle deletions, so the entire directory needs to be rebuilt.
- The first commit of a branch is not optimized for a single file.  I can't figure out how to get the commit hash for the commit before the start of the branch (it may not be available), so I currently pass in the commit for the change itself which causes the entire source to be processed
- I suspect deleting a single source XML file will fail, but it has not been tested

@swinslow I propose merging this in even with these issues - we can add issues to be addressed later.  I've spent over 20 hours on this so far and I really can't afford to spend much more time this week.

goneall:
@iamwillbar @swinslow Let me know if you have a chance to review - check through the comments for known problems.  If OK with merge, I'll manually merge and push to master - it appears there is some issue with the merge conflict.

iamwillbar:
@goneall overall this looks reasonable, I agree that pushing with those issues and then iterating on them (potentially as part of moving away from the Makefile) makes sense.

On the practicality of moving away from Makefile, at first glance it does seem practical with native GitHub Actions. 

There are community GitHub Actions for generating a list of changed files, using one of those may help with the problem you're encountering, here's one that produces a JSON file containing the changes:
https://github.com/marketplace/actions/get-changed-files

iamwillbar (approved):

swinslow:
Hi @goneall -- thanks again. To the extent I follow the details, I'm good with merging as-is if you and @iamwillbar are aligned (and it sounds like you are). I am fine with the caveats you mentioned above. Thank you both for all your help!

swinslow (approved):

goneall:
I just squashed and merged to master - the action is running right now - keeping my fingers crossed.

goneall:
Success :)

PR #1209, opened 2021-03-04
New License: GD License

jlovejoy:


swinslow:
@goneall will take a look at why this is failing!

goneall:
@swinslow @jlovejoy As suspected this is caused by the copyright text being too long.  As a temporary workaround, you can replace the `<copyright>` with `<optional>`.

I'm working on a solution, but it involves redesigning part of the publisher since just increasing the size will cause a performance issue.

goneall:
@swinslow @jlovejoy I have a fix in progress - I should be able to update the tools version by tomorrow morning if you want to wait on the workaround

goneall:
@jlovejoy @swinslow I'm planning on changing the maximum number of characters in a copyright text from 1,000 to 5,000.  Let me know if you think it should be more.  The tradeoff is the performance in doing comparisons (e.g. when running the duplicate checks).

swinslow:
Thanks @goneall! 5,000 sounds like it should be plenty (seems we've been okay with <1000 until now!)  Thanks again for taking care of this.

goneall:
@jlovejoy You may need to merge the latest from License-List-XML master back to your local to clear this error.  I did test it on my local machine and it passes with the new version of the license list publisher.  I thought I could just restart this after the merge, but it doesn't look like it is picking up the changes.

swinslow:
@goneall I tried merging the master branch back into this branch, but it pulled in hundreds of files -- I'm guessing the branch that this one was based on was quite old. So I'm closing this PR in place of #1214 which is just a re-submission of the same files that @jlovejoy prepared here. Thank you both for your help on this one!

PR #1226, opened 2021-03-26
Fix NCGL-UK-2.0 metadata typos

hsteptoe:
Fixing some typos in the XML metadata:

- Correcting <crossRef> link
- Correcting version `listVersionAdded`


jlovejoy:
@hsteptoe - good catch on the URL!
but the license list version is correct, see - https://github.com/spdx/license-list-XML/releases

Could you update this PR to put that as it was?

jlovejoy:
nevermind, looks like I was able to commit to your branch and fix it (not sure if I should do that, but this is kind of minor so...)

@swinslow - can you give it a once-over and then merge?

hsteptoe:
@jlovejoy Thanks for actioning this - sorry, it slipped off my radar...

Issue #1225, opened 2021-03-26
Expose equivalentwords.txt to license-list-data


m1kit:
I think it would be great if https://github.com/spdx/license-list-data includes [equivalentwords.txt](https://github.com/spdx/license-list-XML/blob/master/equivalentwords.txt) and we can access it on `https://spdx.org/licenses/equivalentwords.txt`.
This repository is mainly used for internal XML representation and is not intended for public use, right?

goneall:
This could be implemented by updating the [Makefile](https://github.com/spdx/license-list-XML/blob/master/Makefile) to copy the file to the license list data website subdirectory when `deploy-license-data` goal is executed.

PR #1227, opened 2021-03-27
Expose equivalentwords to license-list-data

m1kit:
This will close https://github.com/spdx/license-list-XML/issues/1225.

I renamed the file to `equivalentwords.csv` to make it clear the file is in a form of `csv`, but I can revert it if there is a problem.


m1kit:
Oh, the author of commits seem to be wrong... I'm fixing

Makefile
```diff
@@ -51,7 +56,7 @@ licenseListPublisher-%.jar.asc:
 
 .PHONY: licenseListPublisher-%.jar-valid
 licenseListPublisher-%.jar-valid: licenseListPublisher-%.jar.asc licenseListPublisher-%.jar goneall.gpg
-	gpg --verify --no-default-keyring --keyring ./goneall.gpg $<
```
m1kit, line 54: I forgot to uncomment this... fixing.

goneall (approved):
Changes look good to me - I'm OK with changing the extension to csv - but we should check with other members of the legal team before merging.

@swinslow @jlovejoy  Let us know if you agree with this proposed change - both the renaming of the file and adding the file to the website for public access

m1kit:
I noticed we need to update 8.1.1 in https://spdx.dev/license-list/matching-guidelines/ if we change the extension. I guess it's not  easy, so should I revert this chang,,,?

goneall:
> I noticed we need to update 8.1.1 in https://spdx.dev/license-list/matching-guidelines/ if we change the extension. I guess it's not easy, so should I revert the change...?

Good catch - although it isn't too difficult to update 8.1.1, there may be other references and possibly even programmatic access to the current file location.  So, perhaps we should leave the extension at .txt.

m1kit:
I reverted the change.

Although this is not mandatory, we should update 8.1.1 anyway to refer `http://spdx.org/licenses/equivalentwords.txt` since [license-list-XML](https://github.com/spdx/license-list-XML) is not public API. I'll make a PR in [spdx/spdx-spec](https://github.com/spdx/spdx-spec) for it, if needed.

goneall:
Thanks @m1kit I'll go ahead and merge this.

@swinslow I don't think this will change any of the deployment operations, but just wanted you to be aware it will now be included on the website.

PR #42, opened 2016-03-30
Automated PR for CECILL-2.1

myndzi:

PR #1229, opened 2021-04-04
fix: Commas in MulanPSL-2.0

nschonni:
Consistent full width comma in Chinese sections, and plain comma in English sections

### DIFF
--- a/src/MulanPSL-2.0.xml
+++ b/src/MulanPSL-2.0.xml
@@ -6,9 +6,9 @@
       </crossRefs>
       <notes>Both the Chinese and English translations of this license have been included in the markup, because current examples of use appear to include both translations in license documents.</notes>
     <text>
-        <titleText><p>木兰宽松许可证, 第2版</p></titleText>
+        <titleText><p>木兰宽松许可证，第2版</p></titleText>
 
-        <titleText><p>木兰宽松许可证， 第2版</p>
+        <titleText><p>木兰宽松许可证，第2版</p>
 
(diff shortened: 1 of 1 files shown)


goneall:
I think the test is failing due to the license compare implemented in the [SPDX tools library](https://github.com/spdx/Spdx-Java-Library) does not normalize the commas for compare.

This can be worked around by changing the text in the [MulanPSL-2.0 test file](https://github.com/spdx/license-list-XML/blob/master/test/simpleTestForGenerator/MulanPSL-2.0.txt) to match the changes in this license XML.

I added[ issue #42](https://github.com/spdx/Spdx-Java-Library/issues/48) to the SPDX library to update the comparison tools to resolve the underlying issue.

nschonni:
@goneall I flipped this to Draft because when I looked at the "source" published license it has these inconsistencies in the comma/space.

jlovejoy:
given that these instances look to occur in the TitleText tag - which is not pertinent for matching purposes, I think we can just close this?
@goneall  - agree?

goneall:
> given that these instances look to occur in the TitleText tag - which is not pertinent for matching purposes, I think we can just close this?

It does impact how the license is displayed (although you have to look very closely to see the difference).

@nschonni what do you think?  For matching purposes, the changes shouldn't have an impact - only the display.

jlovejoy:
I'm going to close this now, since it doesn't make a difference and there hasn't been any activity

Issue #1224, opened 2021-03-19
ISC.txt incorrectly includes a specific copyright notice


bsdimp:
The ISC license, as reported by the ISC.txt file and published in the sister license-list-data repo

starts off:
```
ISC License:

Copyright (c) 2004-2010 by Internet Systems Consortium, Inc. ("ISC")
Copyright (c) 1995-2003 by Internet Software Consortium
```
Instead, it should start
```
<copyright notice>
```

This makes it needlessly difficult to write a program that will generate a full license from the indirect license. Consider:
```
/*
 * Copyright 2021 The Dude Bowling, LLC
 *
 * SPDX-License-Identifier: ISC
 * /
```
Ideally the program could replace `<copyright notice>` with `Copyright 2021 The Dude Bowling, LLC` and generate:
```
Copyright 2021 The Dude Bowling, LLC

Permission to use, copy, modify, and/or distribute this software for any purpose with or without fee is hereby granted, provided that the above copyright notice and this permission notice appear in all copies.

THE SOFTWARE IS PROVIDED "AS IS" AND ISC DISCLAIMS ALL WARRANTIES WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL ISC BE LIABLE FOR ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
```

There are many examples of this. MIT-CMU gets it right, all the BSD licenses have this issue. This is more important for files that have multiple authors that evolve over time. The following example wouldn't be out of place in one of the BSD projects that had gotten all the copyright holders to switch to ISC (ignore how for this example):
```
/*
 * Copyright 1994 NetBSD Foundation
 * Copyright 1995-1999 Charles M Hummings
 * Copyright 1998-2013 George W Poindexter, III
 * Copyright 2004-2020 Great Scott Productions
 * 
 * SPDX-License-Identifier: ISC
 */
```
should generate:
```
Copyright 1994 NetBSD Foundation
[...]

PR #10, opened 2016-03-30
Automated PR for AFL-2.1

myndzi:

PR #1237, opened 2021-04-15
ISC: Make copyright notice descriptive

bsdimp:
replace ISC with COPYRIGHT HOLDER as appears to be common practice to make it more generic.


bsdimp:
@jlovejoy this is the pull request from https://github.com/spdx/license-list-XML/issues/1224 that shows what I'm asking for specifically.

jlovejoy:
Thanks @bsdimp - that helps and was what I thought you meant.

I guess I'd consider this change okay or even 'nice to have' but not necessary b/c by way of matching guideline #10 and the XML markup that implements that <copyrightText> - this text would not be required for a match. Thus, in my mind, it doesn't matter if the copyright notice displays a name or just "author" on the page. I'm slightly inclined in these cases to leave these things as is for the main reason of bigger/more fish to fry, but now that you've done the pr... 

@goneall @swinslow  - any reason not to accept this? I mean, making this change doesn't change anything from a matching perspective, but more of a display aspect, agree?

bsdimp:
I'm happy to make whatever changes like this in a controlled way. It's needed for my application more than the main thrust of SPDX for matching, so I should pay the freight since there's other higher priorities. I didn't want to do too many of these until the initial notion of what's needed was settled to avoid having to redo work.

bsdimp:
Also, this doesn't have to be in the next quarterly release if there's bigger fish to fry and limited time.

goneall:
This doesn't impact matching when using the SPDX libraries for matching.  It may impact other tools that don't use the templates or the matching guidelines when doing matches.  Personally, I think this is OK and we can accept it - but it is possible we may get some feedback from tools that use a simple (and faster) license matching algorithm.

mlinksva:
FWIW `AUTHOR` is typically found in ISC license files in the wild, rather than `COPYRIGHT HOLDER`, for example the top 5 most popular ISC licensed packages according to https://libraries.io/licenses/ISC all use `AUTHOR`, as does [OpenBSD](https://cvsweb.openbsd.org/src/share/misc/license.template?rev=HEAD), as does 0BSD which removes the condition from the ISC license text.

bsdimp:
> FWIW `AUTHOR` is typically found in ISC license files in the wild, rather than `COPYRIGHT HOLDER`, for example the top 5 most popular ISC licensed packages according to https://libraries.io/licenses/ISC all use `AUTHOR`, as does [OpenBSD](https://cvsweb.openbsd.org/src/share/misc/license.template?rev=HEAD), as does 0BSD which removes the condition from the ISC license text.

I'd personally be happier using `AUTHOR` as well, to be honest, since that's what the original BSD license used before it was slightly mangled by OSI when they change `AUTHOR` to `COPYRIGHT HOLDER`.  For my use, I don't care so much so long as there's a consistent set of words across all the licenses in the tree, even if some use one and others use another, just so long as they aren't all different.

mlinksva:
I did not realize that about OSI and BSD licenses. 🤔   However for ISC even OSI [publishes](https://opensource.org/licenses/ISC) with `AUTHOR`.

bsdimp:
> I did not realize that about OSI and BSD licenses. 🤔 However for ISC even OSI [publishes](https://opensource.org/licenses/ISC) with `AUTHOR`.

Yea, I started with what was on file here...  I'll defer to the lawyers in the group as to which words to use as I'm not qualified to judge on a legal basis if there's a significant difference between them or what policies the group has had to date as to what to select.... In the future, I'll check with OSI as part of any revisions...

swinslow:
I'm also +1 on this change.

I agree that it isn't required in light of the SPDX matching guidelines. But I regularly come across ISC licenses used by parties other than ISC in the wild, and I generally like the idea of our "more commonly reused" license templates signaling more visibly that they're commonly used that way.

I'll leave this open for a few days to give others a chance to respond, and if no one objects then I'll plan to merge it. Thanks @bsdimp!

src/ISC.xml
```diff
@@ -10,17 +10,15 @@
...
-        <br/>Copyright (c) 1995-2003 by Internet Software Consortium
-         </p>
+         <p>&lt;copyright notice&gt;</p>
       </copyrightText>
 
       <p>Permission to use, copy, modify, and<optional>/or</optional> distribute this software for any purpose with or without fee is
          hereby granted, provided that the above copyright notice and this permission notice appear in all
          copies.</p>
-      <p>THE SOFTWARE IS PROVIDED "AS IS" AND <alt match=".+" name="copyrightHolder">ISC</alt> DISCLAIMS ALL
+      <p>THE SOFTWARE IS PROVIDED "AS IS" AND <alt match=".+" name="copyrightHolder">COPYRIGHT HOLDER</alt> DISCLAIMS ALL
```
mlinksva, line 19:
```suggestion
      <p>THE SOFTWARE IS PROVIDED "AS IS" AND <alt match=".+" name="author">THE AUTHOR</alt> DISCLAIMS ALL
```

src/ISC.xml
```diff
@@ -10,17 +10,15 @@
...
          WARRANTIES WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF MERCHANTABILITY AND
-         FITNESS. IN NO EVENT SHALL <alt match=".+" name="copyrightHolderLiability">ISC</alt> BE LIABLE FOR ANY
+         FITNESS. IN NO EVENT SHALL <alt match=".+" name="copyrightHolderLiability">COPYRIGHT HOLDER</alt> BE LIABLE FOR ANY
```
mlinksva, line 21:
```suggestion
         FITNESS. IN NO EVENT SHALL <alt match=".+" name="authorLiability">THE AUTHOR</alt> BE LIABLE FOR ANY
```

mlinksva (changes requested):
Just registering my preference in the form of commit suggestions for `THE AUTHOR`, which is most common use in the wild, as mentioned in a [previous comment](https://github.com/spdx/license-list-XML/pull/1237#issuecomment-820832739) (in which I inadvertently omitted `THE `).

jlovejoy:
@swinslow - does the two place where author is used need to have different names? (i.e., "author" and "authorLiability") as @mlinksva has revised it? I would think it could be the same??

bsdimp:
> @swinslow - does the two place where author is used need to have different names? (i.e., "author" and "authorLiability") as @mlinksva has revised it? I would think it could be the same??

He followed the pattern I used which followed the pattern from either the MIT or BSD license XML files that I looked at before submitting the change initially. Logistically for me, having one name for the pair would be ideal in general. However, aa quick survey suggests that's not the common practice. And it isn't clear to me that the notion I have in mind is even sane and/or needed. Having the text itself be identical is all I think I care about so a definition for 'AUTHOR' could be giving in our project's policy doc. I'm fine with either wording (mine or mlinksva's).

PR #738, opened 2018-12-18
OBSD: Add <alt> tag around the copyright text

waldyrious:
Addresses #737.

See also #423 and #657 for similar PRs.

Issue #737, opened 2018-12-17
BSD0 license cites someone's name


techknowlogick:
The BSD0 license cites the name of "Copyright (C) 2006 by Rob Landley rob@landley.net".

Issue copied from: go-gitea/gitea#5397

zvr:
What is the issue?
The SPDX matching guidelines clearly state that copyright notices should be ignored (section **10** in https://spdx.org/spdx-license-list/matching-guidelines).

techknowlogick:
@zvr Thanks 😄

waldyrious:
Still, for completeness, shouldn't it be wrapped in an `<alt match=".+" name="copyright">` tag, as is done e.g. [for the ISC](https://github.com/spdx/license-list-XML/blob/3.3/src/ISC.xml#L10)?

zvr:
PRs always welcome ;)

PR #1246, opened 2021-04-29
Update 0BSD.xml

yearski:
Replace year and copyright holder specific example with placeholder. This method is consistent with BSD-2 and MIT files.

The root issue is that some software may (incorrectly?) consume the license file and generate a license citing Rob Landley, causing confusion. This has come up before, #737 for example. It was addressed by #738, but regressed by c6ac6f10087e66b5858fb8f072bba2cb326f5928.

### DIFF
--- a/src/0BSD.xml
+++ b/src/0BSD.xml
@@ -6,7 +6,7 @@
       </crossRefs>
     <text>
       <copyrightText>
-         <p>Copyright (C) 2006 by Rob Landley &lt;rob@landley.net&gt;</p>
+         <p>Copyright (C) &lt;year&gt; &lt;copyright holder&gt;</p>
       </copyrightText>
 
       <p>Permission to use, copy, modify, and/or distribute this software for any purpose with or without fee is


jlovejoy:
as per the SPDX Matching Guidelines, the copyright notice for the software is considered replaceable - meaning, if there is a different name, it does not make it a different license. #738 was regressed b/c the variable markup is not needed due to the matching guidelines broader rule and <copyrightText> tag

I'm not sure what you mean by "some software may (incorrectly?) consume the license file" ?

yearski:
> I'm not sure what you mean by "some software may (incorrectly?) consume the license file" ?

For example, Gitea uses the SPDX database as the source of licenses. A user selects a license for their repo and then the software pulls it in from SPDX. I was working on a project and checked the license which was used 0BSD but the copyright was "Rob Landley", this caused quite a bit of confusion as no one knew Rob Landley nor understood his relationship to the project. This has occurred to others (https://github.com/go-gitea/gitea/issues/5397 was the root issue for #738)

jlovejoy:
@yearski - thanks for that explanation! Just as a follow-up and curiosity - so, then would a Gitea user than be able to update/change the copyright notice to the right name? Because having just "author" still wouldn't be correct.  Either way, the Gitea user still needs to update the copyright notice to their own name.

Thinking about this - one thing to consider would be for Gitea to pull from the XML files and then ignore the <copyrightText> tag. @seabass-labrax is looking into this more and will comment over there.

I also want to check / confirm with @landley that this is, indeed, intended to be the copyright notice for the software and not the license itself (I've assumed the former, but if it's the latter, this would change my comment re: matching guidelines!)

yearski:
> so, then would a Gitea user than be able to update/change the copyright notice to the right name? Because having just "author" still wouldn't be correct. Either way, the Gitea user still needs to update the copyright notice to their own name.
Correct. The issue for our project was confirming that Rob Landley was not a copyright holder and could be safely deleted; if instead the license had "<copyright holder>", it would have been obvious that it was an incomplete template. (In many ways, this same issue is echoed in your final question about the authority to remove Rob Landley's name from the license text)

> Thinking about this - one thing to consider would be for Gitea to pull from the XML files and then ignore the tag. @seabass-labrax is looking into this more and will comment over there.
Agreed, but given https://github.com/go-gitea/gitea/issues/5397 it's seems that the Gitea team wants to simply copy the text from the source, which is why my proposed patch makes the license text of 0BSD consistent with other licenses here (BSD-2 and MIT for example)

bsdimp:
I think this gets back to the observation I made that we should have GENERIC copyright statements in the license that are obvious how to fill in (eg from the copyright just above/below the SPDX-License-Identifier: line). Many of the BSD licenses have this issue as well. It gets back to my quest as well for 'how does one know what license one is agreeing to when one sees:
```
/* Copyright 2020 M Warner Losh SPDX-License-Identifier: BSD-2clause */
```
and who would one go to if one needed to litigate something about license compliance? If the copyrights are all generic, like in the GPL-family, then it's obvious. If there's specific names filled into the templates, then it's less clear (common sense says those names are replaced, literal reading says they aren't: this could be a dispute due to the ambiguity).

So I'd be in favor of merging it.

mlinksva:
👍 I'd repeat the same comment I made at https://github.com/spdx/license-list-XML/issues/1224#issuecomment-803041240

landley:
Yes, the (suggested) copyright notice line is for the material being copyrighted, not for the license text itself. It's nice to know who is licensing the material, and how to contact them.

landley:
FYI when I first submitted 0BSD to SPDX in 2015, I had "your name here" in the license text: https://lists.linux-foundation.org/pipermail/spdx-legal/2015-June/001443.html

Alas, SPDX's approval cut and pasted the license text from https://landley.net/toybox/license.html and github pulled it in from their database.

jlovejoy:
thanks all for the comments and @yearski  - great to hear about another use of the SPDX License List!

jlovejoy:
this failed due to lack of .txt file - so, I just went ahead and made a new PR that passed checks as that was easier (for me)
see PR #1263

Issue #1255, opened 2021-05-11
FTL.TXT is available on freedesktop.org officially


fu7mu4:
The description of Freetype Project License in https://spdx.org/licenses/FTL.html says:
```
Other web pages for this license 
http://freetype.fis.uniroma2.it/FTL.TXT [no longer live]
http://git.savannah.gnu.org/cgit/freetype/freetype2.git/tree/docs/FTL.TXT
```

But [the freetype project](https://www.freetype.org/) says that "The git repositories at Savannah will stay as mirrors" .
I recommend bellow URL instead of http://freetype.fis.uniroma2.it/FTL.TXT written in above page: 
https://gitlab.freedesktop.org/freetype/freetype/-/blob/master/docs/FTL.TXT

The above URL is linked from [License](https://www.freetype.org/license.html) in the new Freetype Project's License page.

jlovejoy:
thanks!

seabass-labrax:
The new URL has been added to the XML file, @fu7mu4. The link will be on the Licence List website soon! Thanks again for opening this issue :)

PR #1256, opened 2021-05-12
Add new URL for the FreeType Project License (FTL)

seabass-labrax:
Fixes issue #1255


seabass-labrax:
Thanks for merging! :)

Issue #1261, opened 2021-05-14
New license request: NLOD-2.0 [SPDX-Online-Tools]


evancharlton:
**1.** License Name: Norwegian Licence for Open Government Data (NLOD) 2.0
**2.** Short identifier: NLOD-2.0
**3.** License Author or steward: Norwegian Digitalisation Agency
**4.** Comments: The first version of this (NLOD-1.0) is already in the SPDX database; this is merely a request to include version 2.0 as a separate entity.
**5.** Standard License Header: 
**6.** License Request Url: http://tools.spdx.org/app/license_requests/109
**7.** URL(s): https://data.norge.no/nlod/en/2.0/
**8.** OSI Status: Not Submitted
**9.** Example Projects: https://github.com/NVE/regObs4/
**8.** License Text Diff: https://github.com/spdx/licenseRequestImages/blob/master/5b2db5a6-2c5c-4889-814f-4c221a6eacc5.png

**Note:**
The license closely matched with the following license ID(s): NLOD-1.0

DennisClark:
I cannot find a plain text version of this license without interspersed Comments.  If anyone can find that, please post a comment here.

evancharlton:
I wasn't able to find it either -- I had to manually reconstruct the XML file 😞

seabass-labrax:
Welcome to this repository, @evancharlton! Thank you for opening this issue, and especially for taking the time to create an associated pull request with the XML and plain text. This will be most useful if the license submission is accepted :)

Following the [inclusion principles](https://github.com/spdx/license-list-XML/blob/master/DOCS/license-inclusion-principles.md) for the SPDX License List:

#### Definitive factors:

- [x] The license does not match another license already on this list.
- [x] The license does not apply only to software executables.
- [x] The license has identifiable and stable text.
- [x] The license steward appears to be committed to not modifying the license after addition to this list.

#### Other factors (roughly in order of importance):

[...]

PR #1262, opened 2021-05-14
Add NLOD-2.0

evancharlton:
The Norwegian Licence for Open Government Data (NLOD) has a new version;
add NLOD-2.0 alongside NLOD-1.0.

Closes: #1261

### DIFF
--- a/src/NLOD-2.0.xml
+++ b/src/NLOD-2.0.xml
@@ -1,0 +1,300 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<SPDXLicenseCollection xmlns="http://www.spdx.org/license">
+   <license isOsiApproved="false" licenseId="NLOD-2.0"
+            name="Norwegian Licence for Open Government Data">
+      <crossRefs>
+         <crossRef>http://data.norge.no/nlod/en/2.0</crossRef>
+      </crossRefs>
+      <notes>Norwegian translation available here:
+         http://data.norge.no/nlod/no/2.0</notes>
+    <text>
+      <titleText>
+         <p>Norwegian Licence for Open Government Data (NLOD)</p>
+      </titleText>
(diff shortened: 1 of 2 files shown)


src/NLOD-2.0.xml
```diff
@@ -0,0 +1,300 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<SPDXLicenseCollection xmlns="http://www.spdx.org/license">
+   <license isOsiApproved="false" licenseId="NLOD-2.0"
+            name="Norwegian Licence for Open Government Data">
```
swinslow, line 4: Two comments on the `<license>` tag:

* Can you include the version number (2.0) in the title?
* Can you also include a ` listVersionAdded="3.14"` property? See https://github.com/spdx/license-list-XML/blob/master/src/BSD-3-Clause-No-Military-License.xml for an example

evancharlton: Done!

Would you like me to update the NLOD-1.0 license definition as well? I notice that one has a "full" name of "Norwegian Licence for Open Government Data (NLOD) 1.0" as well.

swinslow: Thanks @evancharlton!

Yes, if you want to submit an update for NLOD-1.0, that would be great since the "full name" on the license list doesn't currently include the version number. Appreciate if you can submit it as a separate PR from this one so that we can track them separately. Thank you!

evancharlton: You got it -- I'll send a follow-up PR.

evancharlton:
The tests should pass now if you want to run them again 😁

src/NLOD-2.0.xml
evancharlton: https://github.com/spdx/license-list-XML/pull/1279

Issue #464, opened 2017-10-30
Test for license D-FSL-1.0 failed due to alt tag embedded in optional tag


goneall:
The SPDX tools currently does not support this embedded, although it is allowed by the schema.

See discussion in https://github.com/spdx/license-list-XML/issues/462

goneall:
Resolved when issue #462 was resolved.

PR #1265, opened 2021-05-18
mark entire "Creative Commons notice at bottom" as optional

TimidRobot:
Per [Legal Code Defined - Creative Commons](https://creativecommons.org/legal-code-defined/), the *Creative Commons notice at bottom* is one of the *Elements of the license page that are not part of the legal code*.

This not only improves the accuracy of the data here, but helps with [Does not detect "trimmed down" CC-BY-4.0 · Issue #464 · licensee/licensee](https://github.com/licensee/licensee/issues/464)--ex. helps end users with a trimmed down CC Licenses in their repository.

### DIFF
--- a/src/CC-BY-4.0.xml
+++ b/src/CC-BY-4.0.xml
@@ -428,17 +428,19 @@
                the legal processes of any jurisdiction or authority.
           </item>
         </list>
-      <p>Creative Commons is not a party to its public licenses. Notwithstanding, Creative Commons may elect to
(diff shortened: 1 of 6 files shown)


mlinksva:
I don't think there's anything to prevent there being nested optionals, which may be useful here. The currently optional "The text of the Creative Commons public licenses is dedicated to the public domain under the CC0 Public Domain Dedication." is sometimes present or not in the wild, and still probably should be optionally present in the larger optional.

PR #1272, opened 2021-06-04
Update MPL-2.0 crossRef to use https

jcoyne:
The http:// address redirects to an https:// address, so this avoids a redirect.


jlovejoy:
thanks for the PRs - we are generally not worrying updating these, as they generally redirect. 
FYI - where links are outdated, we leave the old ones and simply add new ones as per https://github.com/spdx/license-list-XML/blob/master/DOCS/license-fields.md section C

jlovejoy:
if you are willing to help, please join our fortnightly calls!

PR #1276, opened 2021-06-18
Add Open Publication License (OPUBL-1.0)

seabass-labrax:


swinslow (commented):
Thanks for this @seabass-labrax! 

Overall LGTM, just one question -- and apologies if we discussed this already elsewhere :)

For the appendix at the end, I see you marked it as optional text, which is great. I'm just wondering because it explicitly says "This is not considered part of the license" whether to include it at all. 

In examples of real-world use, have you seen that appendix text showing up in LICENSE or COPYING files? I'm sort of inclined to omit it -- but if it shows up in those sorts of files then probably best to handle it the way you have here, as optional text that can be matched if it's present.

src/OPUBL-1.0.xml
```diff
@@ -0,0 +1,79 @@
+<?xml version="1.0" encoding="utf-8"?>
+<SPDXLicenseCollection xmlns="http://www.spdx.org/license">
+  <license licenseId="OPUBL-1.0" isOsiApproved="false" name="Open Publication License v1.0" listVersionAdded="3.14">
+    <crossRefs>
+      <crossRef>http://opencontent.org/openpub/</crossRef>
+      <crossRef>https://www.debian.org/opl</crossRef>
+      <crossRef>https://www.ctan.org/license/opl</crossRef>
+    </crossRefs>
+    <notes>Users of this license may wish to take care to identify which, if any, of the 'license options' described in section VI apply to the licensed work. The Free Software Foundation considers only the form with no license options applied to be a free documentation license.</notes>
+    <text>
```
jlovejoy, line 10: we should remove the sentence about what the FSF thinks - we don't ever include that kind of thing as it's a interpretation.

jlovejoy:
> For the appendix at the end, I see you marked it as optional text, which is great. I'm just wondering because it explicitly says "This is not considered part of the license" whether to include it at all.

I'd say just leave it as optional - that will make it fine either way and it's what we've done for other licenses that have this kind of thing

src/OPUBL-1.0.xml
seabass-labrax: Right'o - patch added to this pull request.

Issue #1232, opened 2021-04-07
New license request: CC-BY-NC-SA-3.0-DE


wgresshoff:
1. License Name: Creative Commons Namensnennung - Nicht kommerziell - Weitergabe unter gleichen Bedingungen 3.0 Deutschland
2. Short identifier: CC-BY-NC-SA-3.0-DE
3. License Author or steward: Creative Commons Corporation
4. Comments: See below
5. Standard License Header:
6. URL: https://creativecommons.org/licenses/by-nc-sa/3.0/de/legalcode
7. OSI Status: Not Submitted

Comments:
This license is often used in german universities document servers like miami (https://miami.uni-muenster.de), edoc (https://edoc.hu-berlin.de/) and others. I hope this is sufficient to be added to SPDX licence list.

swinslow:
Closed by #1283

PR #1283, opened 2021-06-28
Add CC-...-3.0-DE licenses

wgresshoff:
This is for closing issues #1230 #1231 #1232 #1233 #1234 and #1235 .


swinslow:
@wgresshoff, thanks so much for preparing these! I'm starting the CI checks now -- will review after they run and will circle back. Appreciate all your help with preparing the XML and text files for these.

swinslow:
Hi @wgresshoff, I pushed a5ee301 which fixed two minor bugs and one cosmetic issue:

- The CI checks were failing due to the template texts (in the .txt files) including "Creative Commons" before "Namensnennung" in the first line, but the XML templates (in the .xml files) _not_ including "Creative Commons" in the `<titleText>` field. I've added "Creative Commons" to `<titleText>` now which should fix this.
- In CC-BY-SA-3.0-DE.xml, the `licenseId` property in the `<license>` tag was accidentally mislabeled as `CC-BY-NC-ND-3.0-DE`. I've fixed that so that the matches will run correctly.
- Finally, more cosmetic than anything, but: For other CC licenses that are ported to other jurisdictions, we have (for better or worse) listed the official "title" on the [SPDX License List](https://spdx.org/licenses/) using the English title. See for example [CC-BY-3.0-AT](https://spdx.org/licenses/CC-BY-3.0-AT.html) and [CC-BY-SA-2.1-JP](https://spdx.org/licenses/CC-BY-SA-2.1-JP.html). Whether or not this is the best approach is probably a separate question, but just to ensure alignment with current practices, I've switched the `name` property in the `<license>` tag to use the English name instead. (The `<titleText>` tag in the actual license text uses the German language version, as noted above.)

Looks like the tests are now passing. Do you have any questions or thoughts about the above? If not, I think this is good to go. Thanks again for all the hard work in putting the templates together!

wgresshoff:
Hi @swinslow , thanks for fixing the minor bugs. Would have done myself, but was off duty due to illness.

swinslow (approved):

roru78:
> This is for closing issues #1230 #1231 #1232 #1233 #1234 and #1235 .

Issue #1285, opened 2021-06-29
New license request: CC-BY-NC-SA-2.0-UK


larsgw:
1. License Name: Creative Commons Attribution - Non-Commercial - Share-Alike 2.0 England and Wales
  2. Short identifier: CC-BY-NC-SA-2.0-UK
  3. License Author or steward: Creative Commons Corporation
  4. Comments: See below
  5. Standard License Header:
  6. URL: https://creativecommons.org/licenses/by-nc-sa/2.0/uk/legalcode
  7. OSI Status: Not Submitted

Comments:
This license is used by the Royal Entomological Society for their out-of-print handbooks. Example: https://www.royensoc.co.uk/sites/default/files/Vol06_Part03b_Richards.pdf

swinslow:
From some quick Google searches, I'm seeing some other real-world uses of this particular CC variant as well, e.g.:
* https://www.doitpoms.ac.uk/vidlib/terms.php
* https://web.archive.org/web/20200217133701/http://www.flageolets.com/about/licensing.php
* https://cdn2.assets-servd.host/excited-hare/production/uploads/KS2-Soil-Exploration-Live-lesson.pdf

Given that and our usual approach to CC license flavors, I'm a +1 to approve this.

swinslow:
Noting also that [CC-BY-SA-2.0-UK](https://spdx.org/licenses/CC-BY-SA-2.0-UK.html) is already on the list; this would add another that additional includes the -NC property.

jlovejoy:
discussed on July 8th call - accept to add

PR #1291, opened 2021-07-09
Add CC-BY-NC-SA-2.0-UK license

larsgw:
Closes https://github.com/spdx/license-list-XML/issues/1285

### DIFF
--- a/src/CC-BY-NC-SA-2.0-UK.xml
+++ b/src/CC-BY-NC-SA-2.0-UK.xml
@@ -1,0 +1,281 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<SPDXLicenseCollection xmlns="http://www.spdx.org/license">
+   <license isOsiApproved="false" licenseId="CC-BY-NC-SA-2.0-UK" name="Creative Commons Attribution - Non-Commercial - Share-Alike 2.0 England and Wales">
+      <crossRefs>
+         <crossRef>https://creativecommons.org/licenses/by-nc-sa/2.0/uk/legalcode</crossRef>
+      </crossRefs>
+    <text>
+      <titleText>
+         <p>Creative Commons Attribution - Non-Commercial - Share-Alike 2.0 England and Wales</p>
+      </titleText>
+      <optional>
+        <p>CREATIVE COMMONS CORPORATION IS NOT A LAW FIRM AND DOES NOT PROVIDE LEGAL SERVICES. DISTRIBUTION OF THIS LICENCE DOES NOT CREATE AN ATTORNEY-CLIENT RELATIONSHIP. CREATIVE COMMONS PROVIDES THIS INFORMATION ON AN "AS-IS" BASIS. CREATIVE COMMONS MAKES NO WARRANTIES REGARDING THE INFORMATION PROVIDED, AND DISCLAIMS LIABILITY FOR DAMAGES RESULTING FROM ITS USE.</p>
+
+        <p>Licence</p>
+      </optional>
+
+      <p>THE WORK (AS DEFINED BELOW) IS PROVIDED UNDER THE TERMS OF THIS CREATIVE COMMONS PUBLIC LICENCE ("CCPL" OR "LICENCE"). THE WORK IS PROTECTED BY COPYRIGHT AND/OR OTHER APPLICABLE LAW. ANY USE OF THE WORK OTHER THAN AS AUTHORIZED UNDER THIS LICENCE OR COPYRIGHT LAW IS PROHIBITED. BY EXERCISING ANY RIGHTS TO THE WORK PROVIDED HERE, YOU ACCEPT AND AGREE TO BE BOUND BY THE TERMS OF THIS LICENCE. THE LICENSOR GRANTS YOU THE RIGHTS CONTAINED HERE IN CONSIDERATION OF YOUR ACCEPTANCE OF SUCH TERMS AND CONDITIONS.</p>
+
+      <p>This Creative Commons England and Wales Public Licence enables You (all capitalised terms defined below) to view, edit, modify, translate and distribute Works worldwide, under the terms of this licence, provided that You credit the Original Author.</p>
+
+      <p>'The Licensor' [one or more legally recognised persons or entities offering the Work under the terms and conditions of this Licence]</p>
+
+      <p>and</p>
+
+      <p>'You'</p>
+
+      <p>agree as follows:</p>
+
+      <list>
+        <item>
+          <bullet>1.</bullet> Definitions
+          <list>
+            <item>
+              <bullet>a.</bullet> "Attribution" means acknowledging all the parties who have contributed to and have rights in the Work or Collective Work under this Licence.
+            </item>
+
+            <item>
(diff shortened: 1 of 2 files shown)


src/CC-BY-NC-SA-2.0-UK.xml
```diff
@@ -0,0 +1,281 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<SPDXLicenseCollection xmlns="http://www.spdx.org/license">
+   <license isOsiApproved="false" licenseId="CC-BY-NC-SA-2.0-UK" name="Creative Commons Attribution - Non-Commercial - Share-Alike 2.0 England and Wales">
```
seabass-labrax, line 3:
```suggestion
   <license isOsiApproved="false" licenseId="CC-BY-NC-SA-2.0-UK" name="Creative Commons Attribution Non Commercial Share Alike 2.0 England and Wales">
```
(the other Creative Commons licenses in the SPDX License List don't have dashes in the license name field)

seabass-labrax (changes requested):

seabass-labrax:
Wow, that was fast! Thanks for taking the time to follow up on the license request by making this PR, @larsgw :)

I've added an inline suggestion to change the value of the name field to be more consistent with other licenses in the SPDX License List.

Also, there probably ought to be `<p>` tags surrounding some text - the license matching tests pass at the moment, but when we release a new version of the list the generated HTML might look strange without the paragraph tags.

Again, thank you for your interest in SPDX and your contribution!

src/CC-BY-NC-SA-2.0-UK.xml
larsgw: I just meant to highlight the difference between the titles in the legal code ("Attribution - Non-Commercial - Share-Alike 2.0 England and Wales" vs "Attribution-NonCommercial-ShareAlike 2.0"). I am relatively okay with changing that though, since it looks like it does not affect the text file or the `<titleText>` part of the XML file?

larsgw:
> Also, there probably ought to be `<p>` tags surrounding some text - the license matching tests pass at the moment, but when we release a new version of the list the generated HTML might look strange without the paragraph tags.

Do you mean the separate paragraphs within list items?

seabass-labrax:
> Do you mean the separate paragraphs within list items?

Exactly - although, on closer inspection, it looks as if the [LicenseListPublisher](https://github.com/spdx/LicenseListPublisher) tool can insert the `<p>` tags automatically when generating the HTML. @goneall, as the author of the program, would you mind sharing your thoughts on whether or not to add the paragraph tags, please?

src/CC-BY-NC-SA-2.0-UK.xml
seabass-labrax: > I am relatively okay with changing that though, since it looks like it does not affect the text file or the `<titleText>` part of the XML file?

Indeed, it doesn't affect either of those: the `<titleText>` element contains the name of the license as it's actually shown in the license text, whereas the `name` attribute of the `license` tag is used for the 'full name' column of the table on the [SPDX License List webpage](https://spdx.org/licenses/).

The text file is only used as part of the automated tests, and isn't published into the [license-list-data repository](https://github.com/spdx/license-list-data/) or the aforementioned webpage :)

larsgw:
I was waiting for the final word on whether `<p>` tags in list items are necessary. The new license workflow says they are not which is why I omitted them, but that might only be the case when each list item only contains a single paragraph (+ an optional nested-list).

larsgw:
@seabass-labrax I have tried running `LicenseRDFAGenerator` (licenseListPublisher-2.2.2) myself and I notice two things. First, when I run it on the current file I do not see it inserting `<p>` tags inside list items, and as a result the formatting is not entirely correct. Second, when I insert `<p>` myself I get the following error:

```
org.xml.sax.SAXParseException: cvc-complex-type.2.4.a: Invalid content was found starting with element 'p'. One of '{"http://www.spdx.org/license":item, "http://www.spdx.org/license":list}' is expected.
    at com.sun.org.apache.xerces.internal.util.ErrorHandlerWrapper.createSAXParseException(ErrorHandlerWrapper.java:203) ~[?:1.8.0_222]
    # etc.
```

So I think my interpretation of the docs was correct in the sense that `<p>` should not be used inside `<item>` elements. Unless I am running the wrong scripts, of course.

goneall:
I checked the code and the generated HTML for list items inserts a `<li>` tag, but does not insert any `<p>` tags.

From looking at the [License List XSD file](https://github.com/spdx/license-list-XML/blob/master/schema/ListedLicense.xsd), the `<p>` tags should be allowed in the list items.

This may be a bug in LicenseListPublisher - I'll investigate and update this issue on my findings.

goneall:
I was able to add the `<p>` tags within an item (e.g. `<list> ... <item> <p> ... </p> </item> </list>`) without error.

You can not, however, add a paragraph tag after `<list>` - only `<item>` and `<list>` are allowed.

goneall:
In terms of the formatting, I don't think you need the paragraph tags around list items - Attached is the generated HTML for this PR for reference.
[CC-BY-NC-SA-2.0-UK.html.txt](https://github.com/spdx/license-list-XML/files/6913606/CC-BY-NC-SA-2.0-UK.html.txt)

larsgw:
> I checked the code and the generated HTML for list items inserts a `<li>` tag, but does not insert any `<p>` tags.
> 
> From looking at the [License List XSD file](https://github.com/spdx/license-list-XML/blob/master/schema/ListedLicense.xsd), the `<p>` tags should be allowed in the list items.
> 
> This may be a bug in LicenseListPublisher - I'll investigate and update this issue on my findings.

> I was able to add the `<p>` tags within an item (e.g. `<list> ... <item> <p> ... </p> </item> </list>`) without error.
> 
> You can not, however, add a paragraph tag after `<list>` - only `<item>` and `<list>` are allowed.

Oh I see, I think I made a mistake on my end then. In section 2 there's a line ("Additional Provisions for third parties making use of the Work") that I think acts as a header for sections 2.2–2.5, so I removed it from the end of the `<item>` of section 2.1 and put it right between the `<item>`s of sections 2.1 and 2.2. However, that would be inside the `<list>`, which I did not realise.

> In terms of the formatting, I don't think you need the paragraph tags around list items - Attached is the generated HTML for this PR for reference.
> [CC-BY-NC-SA-2.0-UK.html.txt](https://github.com/spdx/license-list-XML/files/6913606/CC-BY-NC-SA-2.0-UK.html.txt)

The HTML I got was slightly different, but I think the (slight) formatting problems are the same. "HOWEVER, You must not:" are on one line in the HTML for example, while they are two different paragraphs in the original text.

larsgw:
I can add the `<p>` tags, but I am not entirely sure what to do about the header in section 2. I can keep it as it is, which works visually but is not entirely semantic. I can also split the list into two, one for section 2.1 and one for 2.2–2.5, and put the header in between in a `<p>` tag, which would be in the `<item>` of section 2 and therefore valid, but also not entirely semantic if you consider the sections as being one list.

larsgw:
Okay, than this commit updates the formatting.

seabass-labrax (approved):

PR #1297, opened 2021-07-15
0BSD: add markup for (optional) title

waldyrious:
Markup for the title added as an optional component, in a similar fashion to what was done for the [ISC license title](https://github.com/spdx/license-list-XML/blob/e0b157f45e17ff1a7b6aa8ad46fbfab05912a014/src/ISC.xml#L9-L11) (see #497 for relevant discussion).

It's also worth noting that, [unlike for ISC](https://opensource.org/licenses/0BSD), opensource.org does include the title in the text for the [0BSD license](https://opensource.org/licenses/0BSD).

The match expression in the `<alt>` element supports the following variations:
- "BSD Zero Clause" or "Zero Clause BSD"
- Both with and without the "License" suffix
- Both with and without the "(0BSD)" suffix
- Both "Zero Clause" (separated with space) and "Zero-Clause" (separated with hyphen)

For reference, this was triggered by the discussion at https://github.com/github/choosealicense.com/issues/837 (/cc @x13a and @mlinksva).

### DIFF
--- a/src/0BSD.xml
+++ b/src/0BSD.xml
@@ -5,6 +5,9 @@
          <crossRef>http://landley.net/toybox/license.html</crossRef>
       </crossRefs>
     <text>
+      <titleText>
+         <p><alt name="title" match="(BSD Zero[ -]Clause|Zero[ -]Clause BSD)( License)?( \(0BSD\))?">BSD Zero Clause License</alt></p>
+      </titleText>
       <copyrightText>
          <p>Copyright (C) YEAR by AUTHOR EMAIL</p>
       </copyrightText>


jlovejoy:
Considering the license matching guidelines explicitly state the titles are not matchable, I don't think this is worth doing markup for - see #11 at https://spdx.dev/license-list/matching-guidelines/
I don't think we've ever added markup with the title tag before.

(I believe we also put "BSD" first to make it easier to find alphabetically)

is this causing a problem elsewhere because people are not using the matching guidelines?

goneall:
When generating the [license templates](https://spdx.github.io/spdx-spec/appendix-II-license-matching-guidelines-and-templates/#spdx-license-list-template-access) from the License XML, the LicenseListPublisher creates `<optional>` tags around the title.  

However, I noticed that the `alt` regular expression matches more than a simple optional text would match, so it may be beneficial to keep the `alt` expression if we believe there will be variability in the title text.

waldyrious:
> is this causing a problem elsewhere because people are not using the matching guidelines?

Yes – in fact, this PR was created due to github/choosealicense.com#837, where @x13a [reported](https://github.com/github/choosealicense.com/issues/837#issuecomment-880496569) he tried to use the title as indicated by [opensource.org](https://opensource.org/licenses/0BSD), and that led the license to stop being recognized.

x13a:
Github's [choosealicense.com](https://choosealicense.com/) builds the license from SPDX. Cause SPDX does not have a `titleText` tag in `0BSD.xml` the output license contains only text without a title. It is ok for machines, but is not ok for humans. I assume that it is nice to have a title.

waldyrious:
I'd rather add the crossref in a separate PR, unless the maintainer's prefer it to be added here.

swinslow:
I'm generally in favor of adding this, given @goneall's comments.

Discussed on legal team call 2021-08-05, no objections to adding, so merging.

PR #1294, opened 2021-07-13
Assorted improvements to documentation

seabass-labrax:
## Improvements to `DOCS/new-license-workflow.md`

- Remove references to Travis CI now that the License List uses GitHub Actions for CI/CD

- Fix spelling and formatting

- Correct dead link to the SPDX Online Tools

PR #1300, opened 2021-07-26
Improve `DOCS/new-license-workflow.md`

seabass-labrax:
Further to pull request #1294:

* Fix Markdown list numbering

* Update links

### DIFF
--- a/DOCS/new-license-workflow.md
+++ b/DOCS/new-license-workflow.md
@@ -45,8 +45,8 @@ The following steps assume you're working from a clone of the repo, but we have 
 ### Add the test .txt file
 
 1. The test files are in the `test/simpleTestForGenerator/` directory of your clone of the license-list-XML repo.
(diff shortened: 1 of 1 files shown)


swinslow (approved):

swinslow:
LGTM, thanks @seabass-labrax!

Out of curiosity, for the list numbering, is there a particular preference for numbering them in order vs. "1., 1., 1., ..."? My understanding is that MarkDown handles either format when it's rendered. (Personally I prefer actually writing them out in order, the way you're doing here, but just curious if there's a generally accepted preference...)

jlovejoy:
@swinslow - can I merge this? 
(don't know what the generally accepted preference is for numbering in markup, but so long as it renders correctly...)

seabass-labrax:
The CommonMark specification (which appears to be what GitHub Flavoured Markdown is based on) [doesn't make any distinction](https://spec.commonmark.org/0.30/#lists) between ordering the lists' numbers and not doing so. I often read and write source Markdown as well as the generated HTML, so, like you @swinslow, I tend to prefer the ordered approach :)

PR #1308, opened 2021-08-05
Add CC-BY-NC-SA-3.0-IGO

tigerswell:
Closes #1210.

PR #1313, opened 2021-08-08
Add CC-BY-NC-SA-3.0-IGO

swinslow:
This is a replacement PR for #1308, due to some issues with follow-up edits that accidentally modified a separate file.

All credit and thanks to @tigerswell for the hard work on preparing this addition to the list!


tigerswell:
thanks @swinslow, i read through your comments and saw that the error was due to an accidental error somewhere else? Just making sure that I know if I made an error...don't want to repeat mistakes!

PR #1329, opened 2021-09-05
Add osiApproved=true to wxWindow

goneall:
The OSI website lists this license as OSI approved (see https://opensource.org/licenses/alphabetical).

Even though this license is deprecated, adding this flag removes a warning while running the license generation tool.


vmbrasseur:
lgtm!

PR #1319, opened 2021-08-22
Add 'Verbatim-man-pages' license

alejandro-colomar:
Fixes https://github.com/spdx/license-list-XML/issues/1310

@jlovejoy 
@seabass-labrax 
@swinslow

Issue #1310, opened 2021-08-06
New license request: verbatim


alejandro-colomar:
Hi,

I'm trying to migrate the Linux man-pages to use SPDX markings and be REUSE compliant.  However, the most used license in the man-pages project, the "verbatim" license, doesn't appear in the list.

See <https://www.kernel.org/doc/man-pages/licenses.html>.

Could you please add it to the list?

Thanks,

Alex

seabass-labrax:
@alejandro-colomar, welcome to SPDX!

We'd need some more information about the license to see if it fulfils the [License Inclusion Principles](https://github.com/spdx/license-list-XML/blob/master/DOCS/license-inclusion-principles.md) for the SPDX License List. Please could you include in this GitHub issue the same information that is asked for in our [online tools form](https://tools.spdx.org/app/submit_new_license/)? That would be very much appreciated; thanks! :)

You can, in this case, omit the 'example project URLs' field - this request is unusual in that such examples of use can be readily found by `zcat`-ing the contents of `/usr/share/man`!

alejandro-colomar:
> @alejandro-colomar, welcome to SPDX!

:-)

> 
> We'd need some more information about the license to see if it fulfils the [License Inclusion Principles](https://github.com/spdx/license-list-XML/blob/master/DOCS/license-inclusion-principles.md) for the SPDX License List. Please could you include in this GitHub issue the same information that is asked for in our [online tools form](https://tools.spdx.org/app/submit_new_license/)? That would be very much appreciated; thanks! :)

License's full name: `Verbatim`
Short identifier: `Verbatim`
Source / URL: <https://www.kernel.org/doc/man-pages/licenses.html>
OSI status: I don't know; probably not submitted.  DFSG-compatible (no doubt man-pages ships with Debian), although not listed on Debian's website either.
Standard license header:  None; the full text of the license is used.
License text:


```
.\" Copyright (c) <year>, <copyright holder>
.\" 
.\" %%%LICENSE_START(VERBATIM)
.\" Permission is granted to make and distribute verbatim copies of this
.\" manual provided the copyright notice and this permission notice are
.\" preserved on all copies.
.\"
.\" Permission is granted to copy and distribute modified versions of
.\" this manual under the conditions for verbatim copying, provided that
.\" the entire resulting derived work is distributed under the terms of
.\" a permission notice identical to this one.
.\"
.\" Since the Linux kernel and libraries are constantly changing, this
[...]

Issue #1323, opened 2021-08-26
Update version of the LicenseListPublisher


goneall:
A new version of the LicenseListPublisher is under development which is planned to address the following issues:

- issue #1321 
- https://github.com/spdx/spdx-java-rdf-store/issues/16
- https://github.com/spdx/Spdx-Java-Library/issues/61

swinslow:
Merged #1353 -- closing issue.

PR #1331, opened 2021-09-11
Match erroneous punctuation in Verbatim-man-pages

seabass-labrax:
The trailing characters in the license text are likely to be removed upstream. However, as users of SPDX may wish to run automated license matchers on older files, this commit adds `alt` elements to match license texts containing the erroneous punctuation.

Related to issue #1310 and pull request #1319.

### DIFF
--- a/src/Verbatim-man-pages.xml
+++ b/src/Verbatim-man-pages.xml
@@ -1,41 +1,22 @@
-<?xml version="1.0" encoding="UTF-8"?>
+<?xml version="1.0" encoding="utf-8"?>
 <SPDXLicenseCollection xmlns="http://www.spdx.org/license">
-    <license isOsiApproved="false"
-             licenseId="Verbatim-man-pages"
-             listVersionAdded="3.15"
-             name="The &#34;Verbatim&#34; License (man-pages)">
-        <crossRefs>
-            <crossRef>https://www.kernel.org/doc/man-pages/licenses.html</crossRef>
-        </crossRefs>
-        <text>
-            <copyrightText>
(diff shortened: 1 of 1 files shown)


seabass-labrax:
@goneall, I've checked the test output locally, and I'm a little concerned that the `spacing="both"` attribute doesn't seem to affect the whitespace of the generated HTML. Would you mind taking a look at this, please? Thanks :)

---
verbatim-alt-text

goneall:
@seabass-labrax Looks like an issue with the LicenseListPublisher - I'll look into it after I finish up some of my Docfest tasks - may take a couple days.

goneall:
I created a pull request which resolves this issue: https://github.com/spdx/LicenseListPublisher/pull/122

I just added simple spaces to the HTML rather than using `&nbsp` - I'm not sure if there is any advantage/disadvantage to this approach but it seems to resolve the issue.

I plan on updating the version used to publish the website before the next license list release - this is tracked in issue #1323 

Here's the output from the fixed LicenseListPublisher:

fixspacing

seabass-labrax:
Thank you, @goneall! I had looked in that file and found https://github.com/spdx/LicenseListPublisher/blob/cdb270cb95f2e1bb4517c2191d81f969aa4fb43c/src/org/spdx/licensexml/LicenseXmlHelper.java#L350-L353, but that turned out to be just for the optional text and not for replaceable text! Maybe we should combine the code into a single function that takes Element (be it `optional` or `alt`) and returns a string with the spacing added. What do you think? If you think it's a good idea I can have a go implementing that on the LicenseListPublisher repository :)

@swinslow, @jlovejoy - this pull request is now ready to merge! :)

Issue #1268, opened 2021-05-21
BSD 1 Clause licence text change


sameer1046:
The licence text is not generic for bsd 1 Clause. And contains bsd copyright.

Should be changed according to https://opensource.org/licenses/BSD-1-Clause

seabass-labrax:
Welcome to this repository, @sameer1046!

Looking at the [XML file for BSD-1-Clause](https://raw.githubusercontent.com/spdx/license-list-XML/master/src/BSD-1-Clause.xml), both the copyright statement at the beginning and the name of the software provider in the disclaimer are marked as 'replaceable'. Thus, when following the [SPDX License Matching Guidelines](https://spdx.github.io/spdx-spec/appendix-II-license-matching-guidelines-and-templates/), the license would still be considered BSD-1-Clause even if those parts were different.

For the [*text file*](https://raw.githubusercontent.com/spdx/license-list-XML/master/test/simpleTestForGenerator/BSD-1-Clause.txt), these parts are deliberately not generic: the text files are intended for testing that the automated license matchers can succeed, so using a real copyright notice in these test files can help us catch errors in the XML or matching software.

Of course, we could still modify the XML to make the replaceable parts generic, as was done for another license in #1246. Please tell us if that is something that you would find useful.

Thank you for your interest in SPDX and for taking the time to open this issue. Please tell me if I've missed anything or if you have any further questions! We're having a Legal Team meeting in just under an hour (see the [mailing list archives](https://lists.spdx.org/g/Spdx-legal/message/2947) for more information), so feel free to join that if you happen to be available at such short notice :)

swinslow:
Thank you @sameer1046 and @seabass-labrax! 

Given that BSD-1-Clause is on OSI's list, and since it appears OSI now lists it in the "generic" form (with "[Name of Organization]" instead of "Berkeley Software Design, Inc."), I'd be in favor of making the changes to the template to align with how OSI is displaying it.

I agree with @seabass-labrax that it doesn't have a substantive impact since it is already marked as replaceable text in the template, but no harm in using the generic form particularly where OSI has done so :)

bsdimp:
Using generic replaceable text makes the text product that's generated from the XML make more sense to people looking at it since the replaceability tags are gone from it. Many projects ship the text version in their LICENSE directory, so making it less confusing there is a big win, if nothing else in terms of extra silly questions about who $EXAMPLE_FOR_THIS_LICENSE is.

jlovejoy:
[...]

PR #1342, opened 2021-10-15
Make author name generic for BSD-1-Clause

seabass-labrax:
Resolves issue #1268


bsdimp:
Other licenses use "Author" or "Copyright Holder" rather than "Name of Organization".

seabass-labrax:
The term that I initially used in this pull request was from the OSI's website, but I too would prefer not to assume that the licensor is an organisation. @bsdimp, @jlovejoy - would you concur that we should make it consistent with the other BSD licenses in the SPDX License List?

PR #1351, opened 2021-11-02
Add Community-Spec-1.0

### LINKED ISSUE #1344: New license request: Community-Spec-1.0
1. License Name: Community Specification License 1.0
2. Short identifier: Community-Spec-1.0
3. License Author or steward: [Joint Development Foundation](http://www.jointdevelopment.org)
4. Comments: Used for spec writing, including SPDX :-)
5. Standard License Header:
6. License Request Url: _manually submitted as in issue_
7. URL(s): https://github.com/spdx/governance/blob/main/1._Community_Specification_License-v1.md
8. OSI Status: Not Submitted (I think)
9. Example Projects: SPDX spec

zvr:
This PR adds Community-Spec-1.0 (and the corresponding test text file).
Closes #1344.

Things to note:

- I've added an entry in the `notes` section, pointing to the supporting material, since they were mentioned in #1344.
- The licence text refers to different `.md` files (which are in the supporting material above). Most references seem to have no effect to the license itself, like references to `Notices.md` which lists patents.  However, *Scope* in the license is essentially defined as "look at the other file, `Scope.md`" and *Approved Specification* is also defined as "look at `Governance.md`." Maybe we should have a note that this license text cannot be used standalone, as it _requires_ extra files?
- When transcribing the text, I found two syntax errors. I've opened a [PR](https://github.com/CommunitySpecification/1.0/pull/5) towards the upstream repo about them. If it's merged, the text here must also be changed.
- Four-level deep nested lists are not fun... :unamused:

I apologize for misconstructions in this PR; I'm using the `gh` CLI for all the workflow and still learning...

### DIFF
--- a/src/Community-Spec-1.0.xml
+++ b/src/Community-Spec-1.0.xml
@@ -1,0 +1,429 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<SPDXLicenseCollection xmlns="http://www.spdx.org/license">
+   <license isOsiApproved="false" licenseId="Community-Spec-1.0" name="Community Specification License 1.0">
+      <crossRefs>
+         <crossRef>https://github.com/CommunitySpecification/1.0/blob/master/1._Community_Specification_License-v1.md</crossRef>
+      </crossRefs>
+      <notes>Supporting materials are available at https://github.com/CommunitySpecification/1.0.</notes>
+    <text>
+      <titleText>
+         <p>Community Specification License 1.0</p>
+      </titleText>
+      <p>
+	The Purpose of this License.
+	This License sets forth the terms under which
+	1) Contributor will participate in and contribute to the development of specifications,
+	standards, best practices, guidelines, and other similar materials under this Working Group, and
+	2) how the materials developed under this License may be used.
+	It is not intended for source code.
+	Capitalized terms are defined in the License's last section.
+      </p>
+      <list>
+	<item>
+	  <bullet>1.</bullet>
+          Copyright.
+          <list>
+	    <item>
+	      <bullet>1.1.</bullet>
+	      Copyright License.
+	      Contributor grants everyone a non-sublicensable, perpetual, worldwide, non-exclusive,
+	      no-charge, royalty-free, irrevocable (except as expressly stated in this License)
+	      copyright license, without any obligation for accounting, to reproduce,
+	      prepare derivative works of, publicly display, publicly perform, and distribute
+	      any materials it submits to the full extent of its copyright interest in those materials.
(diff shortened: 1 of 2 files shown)


swinslow:
Thank you @zvr! I still need to take a look at the commits themselves, but just a couple thoughts on your comments here...

> * The licence text refers to different `.md` files (which are in the supporting material above). Most references seem to have no effect to the license itself, like references to `Notices.md` which lists patents.  However, _Scope_ in the license is essentially defined as "look at the other file, `Scope.md`" and _Approved Specification_ is also defined as "look at `Governance.md`." Maybe we should have a note that this license text cannot be used standalone, as it _requires_ extra files?

I tend to lean towards _not_ getting into this in the notes section. I'd say we don't really want the License List itself to get into the business of opining on _how_ licenses ought to be implemented, e.g. which files the license text should be placed in, etc.

For comparison, in [LGPL-3.0](https://spdx.org/licenses/LGPL-3.0-only.html) we don't have a note to say "you ought to include the GPL-3.0 text as well." I'd kind of lump this together in a similar bucket.

> * When transcribing the text, I found two syntax errors. I've opened a [PR](https://github.com/CommunitySpecification/1.0/pull/5) towards the upstream repo about them. If it's merged, the text here must also be changed.

Yup -- I saw there were a couple of typos as well. If they're merged upstream for what the license steward designates as 1.0, then we can update it here (perhaps with `<alt>` matches to catch both versions). To be honest, I kind of hope that if it's merged upstream that they would designate it with a separate version, e.g. 1.0.1 or 1.1 or something...

swinslow (changes requested):
Just one comment -- in line 3, the `<license ...>` tag should include an attribute `listVersionAdded="3.15"`.

If you want to add an additional commit to this PR to fix that, feel free. Otherwise, I'll plan to add that as a follow-up PR after we merge this one.

Otherwise, looks good to me -- thanks again @zvr! I'll wait a day or so before merging this PR, in case others have any feedback.

jlovejoy:
+1 to what @swinslow said above.
Do you want to go head an merge this, or @zvr - are you going to make the minor update on the listVersionAdded="3.15" tag first?

swinslow:
Just merged the new License List Publisher tools in #1353; after those finish running, I'll plan to merge this next and then update the `listVersionAdded` bit.

zvr:
The `listVersionAdded` is already there, people: check the commit after @swinslow's comment above.
Sorry for not commenting about it.

swinslow:
Ah, sorry @zvr -- moving too quickly this morning and I didn't read closely enough. My fault.

Thanks for this! Reviewing and will merge momentarily.

swinslow (approved):

PR #1360, opened 2021-11-15
Spencer-94 - add markup

jlovejoy:
in https://metacpan.org/dist/File-MMagic - this license appears but adds one word. 

I used the <optional> tag, but it'd be nice to be more exact perhaps, if that's possible with the replaceable tag?


swinslow:
I'm +1 to make this change, I think adding "freely" here is essentially the same meaning and is fine to group into the same license ID.

swinslow (approved):

PR #1363, opened 2021-11-16
BSD-3-Clause - remove "all rights reserved"

jlovejoy:
While this language doesn't have real significance, someone asked me about it and I noticed that it is now not showing on the OSI page for BSD-3-Clause - so might as well remove it here to.

### DIFF
--- a/src/BSD-2-Clause.xml
+++ b/src/BSD-2-Clause.xml
@@ -7,7 +7,7 @@
       </crossRefs>
     <text>
       <copyrightText>
-         <p>Copyright (c) &lt;year&gt; &lt;owner&gt;. All rights reserved.
+         <p>Copyright (c) &lt;year&gt; &lt;owner&gt;. 
       </p>
       </copyrightText>
 
--- a/src/BSD-3-Clause.xml
+++ b/src/BSD-3-Clause.xml
@@ -7,7 +7,7 @@
       </crossRefs>
     <text>
       <copyrightText>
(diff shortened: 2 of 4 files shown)


swinslow:
@jlovejoy I'm on board with this change, but it looks like the change here was made to BSD-3-Clause for the XML file but BSD-2-Clause for the test text. Was that intentional?

bsdimp:
This looks good to me.

bsdimp:
I'm changing my mind...  I think it should be flagged as optional text, not deleted unless the license match tool does that automatically for the now-useless Buenos Aires Convention language.
In addition,  166 of the licenses include this phrase, including 14 other BSD licenses.

jlovejoy:
@swinslow - no that was not intentional! I have updated the PR to include the same change for both BSD-2-Clause and BSD-3-Clause, as they are both OSI approved, OSI does not include "all rights reserved" in their template, and they are both used often.

@bsdimp - re: using the optional tag - that would then mean we'd have an optional tag within the copyright tag (which is not "matchable") - not sure we've done that before?? @goneall  or @zvr - any thoughts on this flavor of tag nesting?

bsdimp:
> @bsdimp - re: using the optional tag - that would then mean we'd have an optional tag within the copyright tag (which is not "matchable") - not sure we've done that before?? @goneall or @zvr - any thoughts on this flavor of tag nesting?

If it isn't matchable, it doesn't need an optional tag. Forget I said anything.... Let's not make this harder than it needs to be.

swinslow (approved):
LGTM -- thanks @jlovejoy and @bsdimp!

Issue #1223, opened 2021-03-10
Deprecate one of the two bzip2 licenses, they are duplicates.


pombredanne:
Somehow there is a  patent commentary that has been (IMHO incorrectly) included in the SPDX text of https://spdx.org/licenses/bzip2-1.0.5.html and dropped out of the SPDX text of https://spdx.org/licenses/bzip2-1.0.6.html

Yet this statement is just that: a commentary and was never part of the actual license text. It just shows up in the doc and in some README such as here: https://sourceware.org/bzip2/1.0.5/bzip2-manual-1.0.5.html

The PATENTs statement has never been part of the LICENSE file, but is only a comment in some README and that was (imho incorrectly) taken in the SPDX text instead of using the actual correct LICENSE file text.
There is no material difference in any of the upstream official bzip2 LICENSE texts and therefore there should not be multiple bzip2 licenses at SPDX.

The attached three versions of the LICENSE texts are essentially the same.

The URLs to the original code are:
-  [1.0 of 21 March 2000.LICENSE.txt](https://github.com/spdx/license-list-XML/files/6118055/1.0.of.21.March.2000.LICENSE.txt) is from https://web.archive.org/web/*/http://www.bzip.org/1.0.1/bzip2-1.0.1.tar.gz

- [1.0.5 of 10 December 2007.LICENSE.txt](https://github.com/spdx/license-list-XML/files/6118057/1.0.5.of.10.December.2007.LICENSE.txt) is from  https://web.archive.org/web/20100405073031/http://www.bzip.org/1.0.5/bzip2-1.0.5.tar.gz

- [1.0.6 of 6 September 2010.LICENSE.txt](https://github.com/spdx/license-list-XML/files/6118056/1.0.6.of.6.September.2010.LICENSE.txt) is from https://web.archive.org/web/20161119084155/http://www.bzip.org/1.0.6/bzip2-1.0.6.tar.gz

- and the latest LICENSE is now in git with a full history at https://sourceware.org/git/?p=bzip2.git;a=history;f=LICENSE;h=81a37eab7a5be1a34456f38adb74928cc9073e9b;hb=HEAD

In all cases the bzip2 LICENSE never includes this patent commentary and the official license text has never changed since 1.0 and the switch from GPL. So there should be not two but only one license.

jlovejoy:
[...]

PR #1364, opened 2021-11-24
bzip2 1.0.6 xml - add alt language

tigerswell:
Issue #1223


seabass-labrax (approved):
Looks good to me! :)

swinslow:
+1, looks good to me too!

From https://github.com/spdx/license-list-XML/issues/1223#issuecomment-911862185 I think we also want to deprecate bzip2-1.0.5, but we can do that in a follow-up PR. We'll discuss briefly on the legal team call shortly.

PR #1396, opened 2022-01-09
Use properly formatted license text files as a test reference

sschuberth:
Fixes #30.

Signed-off-by: Sebastian Schuberth <sschuberth@gmail.com>


jlovejoy:
Hi @sschuberth - I'm not sure what this is fixing (I think maybe the issue link is wrong, perhaps?)
this PR includes changes to 299 files! Can you give a quick explanation?

sschuberth:
> I'm not sure what this is fixing (I think maybe the issue link is wrong, perhaps?)

It's indeed issue number 30, but in another repository, see https://github.com/spdx/LicenseListPublisher/issues/30.

> this PR includes changes to 299 files! Can you give a quick explanation?

I hope the details in https://github.com/spdx/LicenseListPublisher/issues/30 make that more clear now: Basically the formatting of all licenses texts is broken.

zvr:
So, essentially @sschuberth you think that the "test" files should be formatted in a specific way and disregard the SPDX matching guidelines? I see no point in this...

sschuberth:
> you think that the "test" files should be formatted in a specific way

Absolutely. Where the "specific way" should be that of the upstream original license text. For example, the SPDX license text for Apache-2.0 should be formatted *exactly* like https://www.apache.org/licenses/LICENSE-2.0.txt.

> disregard the SPDX matching guidelines

I guess you're referring to https://spdx.dev/license-list/matching-guidelines/. I'm not necessarily suggesting to disregard these; but IMO it's very important to maintain the original license text *exactly*, without any changes to formatting or even interpretation in the direction of what parts of the license text could be changed without changing the meaning of the license. It's fine to have such a "normalized" form of the original license text, but that's not what should be stored. Instead, the original text should be stored, plus an algorithm description with which anyone could create the normalized form. Or maybe that normalized form could be stored additionally, but by no means that should be the only / canonical form provided by SPDX.

> I see no point in this...

A very common use-case for this is to assemble plain text attribution documents, like `NOTICE` files. I'd *love* to be able to simply use SPDX' license texts for this, but the plain text we get back out of the XML representation looks just so insanely broken, that we simply can't do it.

zvr:
You obviously know that you can generate the NOTICE file with license text formatted in a variety of ways; there is not a single license that specifies that the formatting of the text has to be reproduced verbatim.

I understand the point that there might be uses for a canonical text representation of some licenses -- although for example I cannot see one for BSD-3-Clause.

However, the "test" files are not "canonical text" files.

I suggest this request is handled by creating another directory of "canonical texts" and then have different PRs for each one, providing references to where they are coming from.
Otherwise, reviewing changes to hundreds of files will take too long... (for example, what's the change in BSD-3-Clause-Open-MPI?)

sschuberth:
> I understand the point that there might be uses for a canonical text representation of some licenses -- although for example I cannot see one for BSD-3-Clause.

I acknowledge that there might be licenses for which no canonical plain text representation is available. My point is: *If* it's available, then that's exactly what SPDX should provide as the plain text version.

> However, the "test" files are not "canonical text" files.

Fair enough.

> I suggest this request is handled by creating another directory of "canonical texts" and then have different PRs for each one, providing references to where they are coming from.

Providing references where the texts are coming from makes sense and resembles my proposal in https://github.com/spdx/LicenseListPublisher/issues/78 that "License details should include a link to the plain text, if known". However, I don't see me creating hundreds of PRs.

Given that I currently have no time to move this PR forward, nor that it would be accepted in it's current form, I'm closing it in favor of a new approach.

goneall:
Wow - catching up on a rather long issue.

@jlovejoy Agree with some of your concerns on use of canonical texts.  I personally don't have much of an opinion on preserving the "canonical" text of licenses, but I acknowledge and respect several individuals and organizations which have requested this.

Currently, the [LicenseListPublisher](https://github.com/spdx/LicenseListPublisher) will copy the license text data to the license text which is stored in the [license-list-data](https://github.com/spdx/license-list-data) repo preserving all white space in the text files.  This was done as a "path of least resistance" fix since many of these text files were copy/pasted from what some may consider as canonical text sources.

@zvr I agree the test file are not, by definition, canonical text.  The approach I used was arguably a bit of a hack, but it was the only reasonable path I could see based on the current level of volunteer labor to implement other approaches.

@sschuberth I appreciate you making the effort for aligning the license texts.

One thought as I'm writing this, perhaps we could use the work @sschuberth has done on creating the formatted text and create a separate directory "formatted_text" or "canonical_text" which includes these files.  I could then update the LicenseListPublisher to pull those files in rather than the test file.

The downside of this approach is everyone creating a new LicenseListXML file would need to submit 2 text files instead of one.

A variation on the above theme is to see if there is a "canonical text" to use.  If not, use the test text.  This would not require additional files to be submitted.

sschuberth:
> create a separate directory "formatted_text" or "canonical_text" which includes these files. I could then update the LicenseListPublisher to pull those files in rather than the test file.

By now, I'd rather follow @zvr's proposal from over [here](https://github.com/spdx/license-list-XML/issues/1452#issuecomment-1112689639) to put canonical text into a separate repo to completely decouple them from the `LicenseListPublisher` mechanism.

goneall:
> By now, I'd rather follow @zvr's proposal from over https://github.com/spdx/license-list-XML/issues/1452#issuecomment-1112689639 to put canonical text into a separate repo to completely decouple them from the LicenseListPublisher mechanism.

I would be interested in the opinions of those requesting the canonical text.  From the previous issues, it seems that there is an expectation that the SPDX API's for accessing license text and the license-list-data repo would contain "canonical" text.

sschuberth:
> it seems that there is an expectation that the SPDX API's for accessing license text and the license-list-data repo would contain "canonical" text.

I definitely agree that'd be desirable. But it seems a bit cumbersome to implement with the current setup, or?

goneall:
> I definitely agree that'd be desirable. But it seems a bit cumbersome to implement with the current setup, or?

If there was a separate directory in the license-list-XML repo, it wouldn't be much of a change.  If it was a separate repo, we'd need to figure out the publishing mechanism when there are 2 different Github actions and the action would be more complicated having to pull from 2 repos - but I think it's possible.

PR #1397, opened 2022-01-19
MPL-2.0 and -no-copyleft-exception: Fix conspicuous indicators

### LINKED ISSUE #493: MPL-2.0 disclaimer, limitation un-conspicoufied
Probably not important for license matching, but to the extent the SPDX texts and site are treated as canonical and copied by others, possibly unfortunate.

http://lu.is/blog/2012/08/19/a-quick-note-on-conspicuous-text-also-known-as-all-caps/ describes how the Mozilla-provided HTML and text versions are highlighted without using all caps.

Maybe there is some annotation that would allow some form of highlighting to be retained in versions SPDX publishes? I don't see anything offhandedly in https://github.com/spdx/license-list-XML/blob/master/schema/ListedLicense.xsd

swinslow:
Fixes #493 

### DIFF
--- a/src/MPL-2.0-no-copyleft-exception.xml
+++ b/src/MPL-2.0-no-copyleft-exception.xml
@@ -1,9 +1,9 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <SPDXLicenseCollection xmlns="http://www.spdx.org/license">
    <license isOsiApproved="true" licenseId="MPL-2.0-no-copyleft-exception"
-            name="Mozilla Public License 2.0 (no copyleft exception)">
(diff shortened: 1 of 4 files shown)


swinslow:
This PR looks like more changes than it really is, because:
* both of the test texts for MPL-2.0 and -no-copyleft-extension are changed to be https://www.mozilla.org/media/MPL/2.0/index.txt
* the "conspicuousness indicators" discussed in #493 are inserted in sections 6 and 7
* the new file also required adding the "===" and "---" lines under the various section headers, also as optional

Should be good to go, assuming the tests pass!

PR #1399, opened 2022-01-24
Fix typo in PS/PDF font exception

c0bw3b:
The exception applies to the GNU Affero General Public License (AGPL) v3.0


swinslow:
LGTM, thank you @c0bw3b!

swinslow (approved):

PR #1408, opened 2022-02-04
Updating X11 and adding variant

### LINKED ISSUE #1389: Make x11 licence generic
The x11 license text contain x11 terms which is not replaceable.

It should be made generic just like ISC licence

capfei:
Please let me know if I need to make any changes!

Fixes #1389

### DIFF
--- a/src/X11-distribute-modifications-variant.xml
+++ b/src/X11-distribute-modifications-variant.xml
@@ -1,0 +1,29 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<SPDXLicenseCollection xmlns="http://www.spdx.org/license">
+   <license isOsiApproved="false" licenseId="X11-distribute-modifications-variant" name="X11 License Distribution Modification Variant">
+      <crossRefs>
+         <crossRef>https://github.com/mirror/ncurses/blob/master/COPYING</crossRef>
+      </crossRefs>
+      <notes>This is same as X11, but with an additional "distribute with modifications" grant and does not have the trademark at the end.</notes>
+    <text>
+      <copyrightText>
+         <p>Copyright (c) &lt;year&gt; &lt;copyright holders&gt;</p>
+      </copyrightText>
+
+      <p>Permission is hereby granted, free of charge, to any person obtaining a copy of this software and
+         associated documentation files (the "Software"), to deal in the Software without restriction,
+         including without limitation the rights to use, copy, modify, merge, publish, distribute, distribute with modifications, sublicense, and/or sell copies of the Software, and to permit persons to whom the Software is furnished to do so,
+         subject to the following conditions:</p>
+      <p>The above copyright notice and this permission notice shall be included in all copies or substantial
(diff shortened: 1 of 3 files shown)


src/X11-distribute-modifications-variant.xml
```diff
@@ -0,0 +1,29 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<SPDXLicenseCollection xmlns="http://www.spdx.org/license">
+   <license isOsiApproved="false" licenseId="X11-distribute-modifications-variant" name="X11 License Distribution Modification Variant">
```
swinslow, line 3: We'll want to add `listVersionAdded="3.16"` since this is a new license. I know it isn't in the original X11 XML since that was a much older one, but for new ID's we're trying to include that property.

src/X11.xml
```diff
@@ -22,13 +22,13 @@
...
       <p>THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT
          LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN
-         NO EVENT SHALL THE X CONSORTIUM BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN
+         NO EVENT SHALL <alt match=".+" name="nameDisclaimer">THE X CONSORTIUM</alt> BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN
          ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE
          USE OR OTHER DEALINGS IN THE SOFTWARE.</p>
-      <p>Except as contained in this notice, the name of the X Consortium shall not be used in advertising or
+      <p>Except as contained in this notice, the name of <alt match=".+" name="name2Disclaimer">the X Consortium</alt> shall not be used in advertising or
          otherwise to promote the sale, use or other dealings in this Software without prior written
          authorization from the X Consortium.</p>
```
swinslow, line 30: I think that this reference to "the X Consortium" should probably also get an alt-tag similar to the one earlier in the sentence.

swinslow (changes requested):

capfei:
@swinslow Thank you for checking these! I made the corrections. 👍

swinslow:
Hi @capfei, thanks! Sorry I wasn't clear with my comment -- on the X11 change, the issue is that "X Consortium" appears twice in that sentence, but only the first instance currently has "alt" tags around it. We'll want to add alt-tags to the second instance as well.

I'm making some updates to PRs and merging them now in preparation for the 3.16 release, so I'll go ahead and see if I can make this change. Thanks again for your help in prepping the PR!

PR #1425, opened 2022-03-10
Add GPL-3.0 text as optional for LGPL-3.0 licenses

swinslow:
As discussed on the mailing list at https://lists.spdx.org/g/Spdx-legal/topic/88334638, this PR adds the GPL-3.0 text as an "optional" addition to the end of the LGPL-3.0 licenses.

This addition is optional in the LGPL-3.0-only and LGPL-3.0-or-later templates, and also in the deprecated LGPL-3.0 and LGPL-3.0+ identifiers for compatibility. Notes have been added to LGPL-3.0-only and LGPL-3.0-or-later to explain this change. Users of the "test text" files will now receive copies of both licenses in a single file, as requested by REUSE participants in connection with the mailing list thread above.

Separately, this also adds notes to LGPL-3.0 and LGPL-3.0+ to explain why those identifiers were deprecated, which didn't appear to be present previously.

Finally, this also updates the "expected warnings" file with the full set of combinations of overlapping LGPL-3.0 license text warnings.


swinslow:
@mxmehl @seabass-labrax I think you were both interested in this -- can you take a look and confirm this fits with the requested changes, from your perspectives? Thanks!

swinslow:
@mxmehl I'm about 98% certain the answer is "yes"  :)  I believe the license-list-data text files are essentially direct copies of the "test" text files from license-list-XML, and this PR updates the test text files for the LGPL-3.0 licenses to add the optional text.

goneall:
> I believe the license-list-data text files are essentially direct copies of the "test" text files from license-list-XML, and this PR updates the test text files for the LGPL-3.0 licenses to add the optional text.

Just confirming the above statement is correct.  If you are using the the text from the text folder of license-list-data repo, it will be a copy of the "test" text file.

mxmehl:
Excellent, thank you!

pombredanne:
@swinslow @jlovejoy I voiced my concerns on list and I still think this change is a mistake until the FSF changes the official text of the LGPL; this will be adopted and this WILL generate new confusion, mistakes and ambiguities. I hope I will be proven wrong, but I very much doubt so.

armijnhemel:
This is not a good idea. LGPL3 is defined as an exception to GPL3 but it doesn't include the GPL3 text itself. It *references* the GPL3 text though. With this change you are mixing syntax and semantics, which is never a good idea.

The proper fix for this would have been in REUSE, not in the standard or the license list.

PR #1434, opened 2022-03-24
KiCad-libraries-exception: Add note on base license

swinslow:
Several of the exceptions, where applicable, indicate in their "notes" section which license(s) they are typically used with. This PR adds a corresponding note for the newly-added exception KiCad-libraries-exception to indicate that it's intended for use with CC-BY-SA-4.0.


swinslow:
@seabass-labrax Since you added the original PR for this exception, does the note here look good to you? Thanks!

seabass-labrax (approved):

swinslow:
Thanks @seabass-labrax! Discussed on 2022-03-24 legal team call, merging now.

PR #1438, opened 2022-03-24
APSL-2.0: add standard header markup

OliverFendt:
added the Standard License Header for APSL-2.0


swinslow:
Hi @OliverFendt, thanks for this!

For some reason the CI tests didn't run on your PR... if they had, it would have flagged one thing about how the `standardLicenseHeader` tag works, which is that it gets inserted inline into the actual license text rather than added as a duplicate section below it. The way that it was proposed in this PR, I think it would have resulted in adding a second copy of Exhibit A to the rendered license.

I've added a replacement PR in #1455 to address this -- also implemented your suggestion to do the optional tags around the quote marks at the start and end of the exhibit, thanks for flagging that!

PR #1446, opened 2022-04-13
SMLNJ: add markup for template #1441

johnmhoran:
This modifies the markup for SMLNJ to permit copyright holders other than those
named in the SMLNJ license to use the license.

Signed-off-by: John M. Horan <johnmhoran@gmail.com>


bobcampbell112233 (approved):

swinslow:
Hi @johnmhoran, I left a comment about the failing test here at https://github.com/spdx/license-list-XML/issues/1441#issuecomment-1099224454 -- take a look and see what you think.

johnmhoran:
Thanks @swinslow -- I replied with a comment in the issue and made the change you suggested.  If not already documented, we might want to document this error and process, and maybe even tackle it when a duplicate/deprecation is initially created?

goneall:
@johnmhoran Please add `,"Duplicates licenses: SMLNJ-NJ, StandardML"` to the expected warnings file to cover a possible case where the files are checked in the reverse order.

I agree we should document this.  It looks like the right place to add it is the [new-license-workflow.md](https://github.com/spdx/license-list-XML/blob/master/DOCS/new-license-workflow.md) file.

goneall:
> If not already documented, we might want to document this error and process, and maybe even tackle it when a duplicate/deprecation is initially created?

I'll work on a PR to the new-license-workflow this morning

johnmhoran:
Thank you @goneall .  One question -- I don't follow the syntax -- do you perhaps mean `,"Duplicates licenses: StandardML-NJ, SMLNJ"` rather than `,"Duplicates licenses: SMLNJ-NJ, StandardML"`?

Also, since this is not a new license but an edit to an existing one, I'd have thought this error would have been thrown, and resolved, when the most recent of the 2 was first added?

goneall:
I added a PR #1447 to document the duplicate licenses

goneall:
@johnmhoran 

> do you perhaps mean ,"Duplicates licenses: StandardML-NJ, SMLNJ" rather than ,"Duplicates licenses: SMLNJ-NJ, StandardML"

In addition to, not rather than. The order in which the licenses are checked isn't deterministic, so we need to add both directions in the expected warnings.

> Also, since this is not a new license but an edit to an existing one, I'd have thought this error would have been thrown, and resolved, when the most recent of the 2 was first added?

I haven't verified it, but my guess is that the additional markup added caused the match.  The CI/CD runs a license matching per the matching guidelines including processing the alt/optional tags.

johnmhoran:
@goneall I'm confused re the syntax.  If I understand correctly, the 2 identifiers are `SMLNJ` and `StandardML-NJ`.  What are `SMLNJ-NJ` and `StandardML`?  The latter are not reversing the order of the files but rather are changing the identifiers themselves.

goneall:
Just FYI - there is an issue to enhance the warnings processing to make it a bit easier to maintain: https://github.com/spdx/LicenseListPublisher/issues/117

Looking for volunteers to implement.

goneall:
> I'm confused re the syntax. If I understand correctly, the 2 identifiers are SMLNJ and StandardML-NJ. What are SMLNJ-NJ and StandardML? The latter are not reversing the order of the files but rather are changing the identifiers themselves.

Oops - that was a typo.  Just reverse the order.

goneall:
I just added an [issue for the LicenseListPublisher to ignore duplicates](https://github.com/spdx/LicenseListPublisher/issues/131) if one or both of the licenses are marked as deprecated.  It would be an easy fix - I just want to check with @jlovejoy and @swinslow to make sure they agree this is a good idea.

johnmhoran:
Thanks @goneall -- just added the reverse-order warning as discussed.

goneall (approved):
Thanks @johnmhoran - looks good

swinslow:
Hi @johnmhoran, because the expected-warnings file had changed in parallel in another PR, there was a conflict with this PR. I've resolved that now and the tests are re-running -- assuming they pass, this should be all good to go!

swinslow (approved):

Issue #1356, opened 2021-11-09
OpenSSL exceptions analysis


jlovejoy:
there are multiple issues open related to variants of the OpenSSL exception. This issue is attempt to bring them all together to one issue. (I'm going to close the other issues, so we have all info here, but for reference see: #1324 #1058 #1007 #720 )

Here is a list of the where the variants have been found (and a shortname that corresponds with attachment)
- openvpn-debian - https://tracker.debian.org/media/packages/o/openvpn/copyright-2.4.0-6%2Bdeb9u2
- psycopg - https://github.com/psycopg/psycopg2/blob/master/LICENSE#L14-L24
- deluge - https://git.deluge-torrent.org/deluge/tree/LICENSE
-  cryptsetup (GPL) - https://gitlab.com/cryptsetup/cryptsetup/-/blob/master/COPYING#L342
- siege - https://github.com/JoeDog/siege/blob/master/README.md
- naxsi - https://github.com/nbs-system/naxsi/blob/951123ad456bdf5ac94e8d8819342fe3d49bc002/naxsi_src/naxsi_raw.c
- incinga2 (Lesser) - https://github.com/Icinga/icinga2/blob/master/README.md
- cryptosetup (Lesser) - https://gitlab.com/cryptsetup/cryptsetup/-/blob/master/COPYING.LGPL#L505

I made a text file with the different variants and highlighted the relevant text in different colors - will try to attach here.

Below are the questions the legal team needs to answer in terms of what is legally substantive differences

jlovejoy:
[openssl-exceptions.pdf](https://github.com/spdx/license-list-XML/files/7507273/openssl-exceptions.pdf)

jlovejoy:
1) can we consider any of these (in the context of describing the permission to link) as equivalent / not legally substantively different:

- the code of this program with the OpenSSL Library (or with modified versions of OpenSSL that use the same license as OpenSSL), 
- the code of this program with any version of the OpenSSL library which is distributed under a license identical to that listed in the included COPYING.OpenSSL file, 
- this program with the OpenSSL library (or with modified versions of OpenSSL that use the same license as OpenSSL),
- the code of portions of this program with the OpenSSL library under certain conditions as described in each individual source file, 

Note: Deluge uses another variant - "the code of portions of this program with the OpenSSL library." and is different once again in omitting the last line of the first paragraph stating "and distribute linked combination including the two." - which all other variants use

jlovejoy:
[...]

PR #1448, opened 2022-04-28
openvpn-openssl-exception - add markup

jlovejoy:
adding markup for variations as per #1356 

there is additional variants to add, but starting with this to make sure I have markup correct!

### DIFF
--- a/src/exceptions/openvpn-openssl-exception.xml
+++ b/src/exceptions/openvpn-openssl-exception.xml
@@ -9,16 +9,23 @@
       <titleText>
          <p>Special exception for linking OpenVPN with OpenSSL:</p>
       </titleText>
-      <p>In addition, as a special exception, OpenVPN Technologies, Inc.
-         gives permission to link the code of this program with the
+      <p>In addition, as a special exception, <alt name="copyrightHolder"
+         match=".+">OpenVPN Technologies, Inc.</alt>
+         give<optional>s</optional> permission to link <alt name="thisProgram" match="(the code of
+         this program)|(this program)">the code of this program</alt> with the
          OpenSSL Library (or with modified versions of OpenSSL that use
          the same license as OpenSSL), and distribute linked
-         combinations including the two. You must obey the GNU General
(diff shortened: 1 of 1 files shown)


swinslow (approved):

swinslow:
Oh wait -- I approved this, but just saw the failing test. Taking a closer look...

goneall:
> Hmmm... from the failing test, I think it's failing because of the <optional>s</optional> occurring in the middle of a word. I'm not sure if that works.

Yup - that doesn't work.  You could call it a tooling issue, but it would be very very difficult to implement character level optional characters unless they are at the beginning or ending of a token/word.  It may even be impossible to implement such a pattern matcher with a deterministic completion time.  Even at the beginning/end it is very difficult to code for this scenario.

> `<alt name="give" match="give|gives">gives</alt>`

should work as a workaround.

jlovejoy:
thanks @swinslow - yeah, I kind of thought the optional tag was a bit weird there.

made update in my file, but still seeing "change requested" tag - not sure how or if I need to turn that off?

jlovejoy:
hmmm still failing. I think I need to add the optional text to the .txt file too?

swinslow:
Hmmm... for the updated failing test log, it looks like it's this part of the match that is now failing:

`<alt name="thisProgram" match="(the code of this program)|(this program)">the code of this program</alt>`

I'm not sure why this is failing, but I'd note that I think this would be equivalent:

`<optional>the code of</optional> this program`

so maybe that would work better / make the tests pass?

goneall:
I don't see an obvious problem with `<alt name="thisProgram" match="(the code of this program)|(this program)">` the code of this program</alt>`   The parenthesis don't add and value for this regex, so you could try `<alt name="thisProgram" match="the code of this program|this program">`.

If either approach doesn't work, I can bring this into my local debugging environment and find out why it's failing.

jlovejoy:
@goneall was right! it was the parens. Turned out I had another PR I started, so I incremented off that one with each change so I could see what the issue was (will now close this PR)

jlovejoy:
closing in favor of $1436

Issue #1442, opened 2022-04-08
New license request: "GStreamer proprietary plugins" exception


hadess:
A bit of background. GNOME Videos (né "totem") is a video player for the GNOME desktop environment licensed under the GPL v2+. It uses GStreamer (LGPLv2.1+) for video playback. As was, and still is, the case, a number of video codecs are patent encumbered and can't be shipped easily in some places.

That's the reason why there are proprietary video decoding plugins available from various vendors, and which totem would like to use without causing problems by recombining, or shipping together, a GPL totem with non-free GStreamer plugins.

In 2005, we added this note to the COPYING file in totem:
```
The Totem project hereby grant permission for non-gpl compatible GStreamer
plugins to be used and distributed together with GStreamer and Totem. This
permission are above and beyond the permissions granted by the GPL license
Totem is covered by.
```

This better version was added to the Rhythmbox music player in 2008 and is a bit clearer:
```
The Rhythmbox authors hereby grant permission for non-GPL compatible
GStreamer plugins to be used and distributed together with GStreamer and 
Rhythmbox. This permission is above and beyond the permissions granted by
the GPL license by which Rhythmbox is covered. If you modify this code, you may 
extend this exception to your version of the code, but you are not obligated 
to do so. If you do not wish to do so, delete this exception statement from 
your version.
```

This exception is also used by GNOME Music, and mentioned in GStreamer's licensing documentation:
https://gstreamer.freedesktop.org/documentation/frequently-asked-questions/licensing.html?gi-language=c#licensing-of-applications-using-gstreamer

seabass-labrax:
# New submission review

## Has the license been approved by the OSI?
*All OSI-approved license will be included on the SPDX License List, regardless of other factors*

No; I see no evidence that the OSI has approved this exception.

## Definitive factors
*These must all be satisfied to allow inclusion in the license list*

### Is the submitted license unique, that is, it does not match another license already on the License List as per the matching guidelines?

Yes, it is not already on the SPDX License List.

### If a software license, does it apply to source code and not only to executables?

Yes.

### Does the license have identifiable and stable text, and is not in the midst of drafting?

The submitter expressed two distinct exceptions, but both seem to be stable.

### Has the license steward, if any, committed to versioning new versions in the future and to not modify it after addition to the list?

I think we'll ask @hadess to answer this one :smiley:

## Other factors for inclusion
*Roughly in order of descending importance*

### 1. Does the license substantially comply with one of the free/open content definitions?
[...]

PR #1449, opened 2022-04-28
Add GStreamer exception to the GPL

hadess:
Closes: #1442


hadess:
A couple of things:
- I don't know if the "listVersionAdded" value is correct
- If there is a way to add the [original 2005 totem version of the exception](https://github.com/spdx/license-list-XML/issues/1442#issue-1197054035) for it to be "detected", a pointer to a license doing something similar would be great

seabass-labrax:
>* I don't know if the "listVersionAdded" value is correct

Yup, that's right! We're about to make the 3.17 release, which this can be part of.

>* If there is a way to add the [original 2005 totem version of the exception](https://github.com/spdx/license-list-XML/issues/1442#issue-1197054035) for it to be "detected", a pointer to a license doing something similar would be great

I think that's sufficiently different that it would need a different SPDX identifier, but we can add it too (the Inclusion Principles would be satisfied for that as well, in my opinion).

hadess:
> I think that's sufficiently different that it would need a different SPDX identifier, but we can add it too (the Inclusion Principles would be satisfied for that as well, in my opinion). Maybe something like `GPL-GStreamer-exception-old`?

I've added it as a separate commit, and fixed up the XML to be valid.

jlovejoy:
oh, also - @hadess - can you remove the "DEPRECATED" note in the Note field? "deprecated" has specific meaning within the context of SPDX. 

@swinslow @goneall - are we using <obsoletedBys> tag? I'm not familiar with that one!

swinslow:
Thanks @hadess! Looking at the comments in #1442, it looks like there might be an open question on this one, so I'm going to move it forward to 3.18 (since I'm about to push out 3.17 this evening).

That said, after this is merged (which I expect will be early in 3.18), we typically see projects starting to use new IDs in code even prior to the formal release -- so hopefully that shouldn't be a blocker to your projects looking to use the new IDs. Thanks!

hadess:
> Thanks @hadess! Looking at the comments in #1442, it looks like there might be an open question on this one, so I'm going to move it forward to 3.18 (since I'm about to push out 3.17 this evening).

I'm really disappointed that I would get asked this question moments before you do a release and miss out on the release because of that when the issue and pull request were opened for all to see for the past week and a half.

jlovejoy:
I'm sorry, I'll take responsibility on that - should have caught/questioned about the short ids choice on the call. Put a longer comment in #1442 to try and keep (most) of discussion in one place

Issue #1520, opened 2022-06-04
KICad exception is recognized as license


ecki:
This exception starts with a <license> tag not with a <exception> tag

https://github.com/spdx/license-list-XML/blob/65dd995755eba6076e895e020a7c7f3a1bcbe06a/src/exceptions/KiCad-libraries-exception.xml#L2

goneall:
`<license ` should be replaced with `<exception `

Good catch @ecki

swinslow:
Great catch. Intriguing that it passed the tests as a `license` despite being in the `src/exceptions/` directory. We'll have to watch out for this in the future!

goneall:
> Intriguing that it passed the tests as a license despite being in the src/exceptions/ directory. 

I was thinking of adding a check for this in the LicenseListPublisher.  I'm a little reluctant to make assumptions about the directory structure in the general utility that creates the license data output - even though it is probably only used for listed licenses.  

The cleanest approach would be to check in the Github action similar to the ValidateSchema.js.  We could also add the check to ValidateSchema.js.

Namdar21:
> `<license ` should be replaced with `<exception `
> 
> Good catch @ecki

256

swinslow:
Merged the specific KiCad fix in #1521 -- thank you both!

@goneall do you want to leave this issue open for the broader fix needed, or should we start a separate thread?

PR #1521, opened 2022-06-04
Chance KiCad-libraries-exception from a license to an exception

goneall:
Fixes issue #1520


src/exceptions/KiCad-libraries-exception.xml
```diff
@@ -1,5 +1,5 @@
 <SPDXLicenseCollection xmlns="http://www.spdx.org/license">
-  <license isOsiApproved="false" licenseId="KiCad-libraries-exception" listVersionAdded="3.17" name="KiCad Libraries Exception">
+  <exception isOsiApproved="false" licenseId="KiCad-libraries-exception" listVersionAdded="3.17" name="KiCad Libraries Exception">
```
ecki, line 2: Maybe also remove isOsiApproved and listVersionAdded?

goneall: It was the `isOsiApproved` - looks like that attribute isn't allowed for exceptions.

swinslow (approved):

swinslow:
Thank you both!

PR #1531, opened 2022-07-13
https://t.me/+Hc-_Kw-Z0J8zYWIy

477899844:
Secure the copied fragments so that they do not disappear after an hour


goneall:
Spam?

PR #1430, opened 2022-03-17
Add LZMA licenses

testworksau:
Closes https://github.com/spdx/license-list-XML/issues/1429

I manually checked the license text from various SDK packages and thought it might be best to add all variants in.

I wasn't sure if there was a standard approach to name licenses that change over versions so feel free to suggest changes in naming / convention etc.

PR #1538, opened 2022-07-28
Add LZMA licenses

### LINKED ISSUE #1429: What license identifier can be used for LZMA SDK?
For:
https://www.7-zip.org/sdk.html

I can find an LZMA-Exception identifier, which looks like its purpose is for NSIS rather than the LZMA SDK itself:
https://spdx.org/licenses/LZMA-exception.html

I've taken a look at:
https://wiki.spdx.org/view/Legal_Team/Decisions/Dealing_with_Public_Domain_within_SPDX_Files_(DRAFT)

However it is unclear (to me) as to whether we should be creating a new identifier for the LZMA SDK license text, which reads:

```
LICENSE
-------

LZMA SDK is written and placed in the public domain by Igor Pavlov.

Some code in LZMA SDK is based on public domain code from another developers:
  1) PPMd var.H (2001): Dmitry Shkarin
  2) SHA-256: Wei Dai (Crypto++ library)

Anyone is free to copy, modify, publish, use, compile, sell, or distribute the 
original LZMA SDK code, either in source code form or as a compiled binary, for 
any purpose, commercial or non-commercial, and by any means.

LZMA SDK code is compatible with open source licenses, for example, you can 
include it to GNU GPL or GNU LGPL code.
```

swinslow:
Fixes #1429 

Replacement for #1430


swinslow:
Discussed on 2022-07-28 legal team call, agreed to add two most recent versions (9.11-to-9.20 and 9.22) for the 3.18 release -- I'll update the PR to reflect that.

For the earlier ones (4.x), there were questions about whether these versions are still in any sort of substantial use? Particularly because the first one is a pretty odd license, being a combination of other existing licenses together with several special exceptions.

jlovejoy:
@swinslow - I think you need to re-do this to only have the two licenses we are going to add, then I can merge

swinslow:
@jlovejoy OK -- this one should be all set now, I revised the commit to leave just the two the legal team agreed to keep.

Issue #905, opened 2019-07-18
matching guidelines location / update Appendix / remove from website


jlovejoy:
to ensure people from the legal team see this, opening issue to cross-reference here:
see https://github.com/spdx/spdx-spec/issues/134

swinslow:
Submitted PR to add this in Appendix II at: https://github.com/spdx/spdx-spec/pull/139

swinslow:
This has now been added to Appendix II in the v2.2 version of the spec: see discussion at https://github.com/spdx/spdx-spec/issues/134

However, since the v2.2 spec hasn't yet been released, I don't think we should remove the website version of the matching guidelines. So I'm inclined to keep this issue open until after the v2.2 spec is released. After it is, we can replace the website copy with a pointer to the spec appendix.

I'm changing this from 3.8 to "later release" so that we don't close it and forget, but so that it isn't pending in case the v2.2 spec is not released in the meantime.

jlovejoy:
I've had a task to raise the location of the matching guidelines generally on the tech/legal team mailing lists (and then update them!!) especially with 3.0 pending... updating tags to raise the visibility of this

jlovejoy:
see https://github.com/spdx/spdx-spec/pull/761

once 2.3 is released, then we can:

- [ ]  once that is pushed, remove from website
- [ ]  update all links on website to point to spec appendix (direct link issue?
- [ ] remove page in DOCS for html version

jlovejoy:
see https://github.com/spdx/LicenseListPublisher/pull/139

swinslow:
@jlovejoy Based on https://github.com/spdx/LicenseListPublisher/pull/139 it looks like this has been completed for the updates to the License List website itself.

Was there anything else to be updated under this issue, or can we mark this as complete? Thanks!

jlovejoy:
once the license list is published, the last step is to takedown the matching guidelines webpage (and check website for other links to be updated), but I didn't want to do that until the license list is published b/c that would create a dead lnk in the meantime. I'll close issue once that's done after release :)

swinslow:
@jlovejoy Now that 3.19 is live, it looks like the links on the License and Exception pages are correctly going to the spec appendix now. Do you want to go ahead and take down the old page, so we can close this issue? Thanks!

jlovejoy:
ah, good reminder. I will do so (and then I'll lose this issue).

jlovejoy:
[...]

PR #1542, opened 2022-07-30
Delete matching-guidelines

jlovejoy:
this was just a file for convenience to track the HTML used for this page on the website. Due to the matching guidelines being in the Spec and the pending goal to remove the webpage (since it's a duplication) and update links to point to the Annex, we don't need this file any longer


jlovejoy:
relates to #905

swinslow:
LGTM, but I'm going to hold on merging this until 3.18 is out -- so that we can keep this connected with the 3.19 docs release cycle.

swinslow (approved):

Issue #1200, opened 2021-02-19
New license request: Python-3.something and maybe others


ockham:
Some Python-related projects (such as https://github.com/nodeca/argparse) use a [license](https://github.com/nodeca/argparse/blob/master/LICENSE) that they refer to as `Python-2.0` in their `package.json`'s (SPDX compliant) [`license` field](https://github.com/nodeca/argparse/blob/70fc26eb5a9a829d6ddcceb6b4c87802218227fd/package.json#L17).

However, that license isn't identical to what SPDX lists as [Python-2.0 license](https://spdx.org/licenses/Python-2.0.html). Specifically, SPDX has the following:

> 7. This License Agreement shall be governed by and interpreted in all respects by the law of the State of Virginia, [...]

That part is found in the "CNRI OPEN SOURCE LICENSE AGREEMENT (for Python 1.6b1)" section. (I'm not even sure if that part is legally binding or just there for historic purposes.) Anyway, I think that that part is what [is mentioned by the GNU project's license list as the reason for that license _being GPL-incompatible_](https://www.gnu.org/licenses/license-list.en.html#PythonOld):

> This is a free software license but is incompatible with the GNU GPL. The primary incompatibility is that this Python license is governed by the laws of the State of Virginia, in the USA, and the GPL does not permit this.

Anyway, `argparse`'s `LICENSE` has the following instead:

> 7. This License Agreement shall be governed by the federal
intellectual property law of the United States, including without
limitation the federal copyright law, and, to the extent such
U.S. federal law does not apply, by the law of the Commonwealth of
Virginia, excluding Virginia's conflict of law provisions. [...]

Note that this is identical to what the GNU Project calls the [Python-2.0.1 license](https://directory.fsf.org/wiki/License:Python-2.0.1) (which unfortunately currently _isn't_ an SPDX-listed license), and [lists as GPL-compatible](https://www.gnu.org/licenses/license-list.en.html#Python).

The latter license is also found on the Python website (which should probably be considered the canonical source): https://www.python.org/download/releases/2.0.1/license/

Would you consider adding the `Python-2.0.1` license? The apparent difference in GPL compatibility seems relevant enough for (FL)OSS projects -- this is basically what prompted me to file this issue :slightly_smiling_face: 

---

Based on my findings at https://github.com/WordPress/gutenberg/pull/28890#issuecomment-777799177).
Also discussed at https://github.com/nodeca/argparse/issues/160.

capfei:
[...]

PR #1549, opened 2022-08-02
New license: Python-2.0.1

jlovejoy:
see lengthy discussion at #1200 

But it seems that this "version" appears in a number of places and thus would be helpful to add. 
Most notably:
- https://www.python.org/download/releases/2.0.1/license/ (which is same text that FSF recreates and claims as FSF-free here: https://www.gnu.org/licenses/license-list.en.html#Python
- https://docs.python.org/3/license.html
- https://github.com/python/cpython/blob/main/LICENSE


jlovejoy:
@swinslow  or @zvr - I tried to do markup throughout to accommodate where sometimes it says: "Python 2.0.1" or "Python 3.10.6" or "Python" (no version number)

I don't think I did it right... maybe I should have had the word Python in there as well, not just the version number? 

It'd be nice to also proof this against other version numbers...

jlovejoy:
btw - I could go either way on short id of:
Python-2.0.1 (which is what I put)
or
PSF-2.0.1 which @brettcannon suggested

The reason I leaned towards Python-2.0 was to be "consistent" with the use of that prior for a license stack, versus PSF-2.0 which was for just the first license of the stack. 

FYI @ockham
@capfei 
@ariel11

As for 0BSD - I did not included that, as I take @VanL comment (through @brettcannon) to mean that one can choose 0BSD or Python/PSF 

Would really like to get this merged asap and be done with it (for now or until some other historical version surfaces...) :)

richardfontana:
PSF-2.0.1 seems confusing - I already have the notion in my mind that "the PSF license Is the top of the Python License stack".

capfei:
Agree with @richardfontana; PSF = top, Python = full.

Thank you @jlovejoy!!!!

gotmax23:
``` xml
and otherwise use Python <alt match="2.0.1|3.10.6| " name="version">2.0.1</alt> alone or in any derivative
```

This is not correct. It should match *any* Python version.

jlovejoy:
@gotmax23 - yeah.... I only found those two examples of version numbers and in other cases there simply isn't a version number at all (Please, PSF, do it this way from now on!)  I need a bit of help on the markup to express, it can be a version or no version number... once I sort that out, then this should be ready to go.

src/Python-2.0.1.xml
```diff
@@ -0,0 +1,284 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<SPDXLicenseCollection xmlns="http://www.spdx.org/license">
+   <license licenseId="Python-2.0.1" name="Python Software
+   Foundation License 2.0.1" listVersionAdded="3.18">
```
swinslow, line 4: Should the license name be "Python License 2.0.1" rather than "Python Software Foundation License 2.0.1"? I think we've used the latter name for the `PSF` version, not for the full `Python` stack.

src/Python-2.0.1.xml
```diff
@@ -0,0 +1,284 @@
...
+               accessing and otherwise using <alt match="this software ("Python")|Python 3.10.6|Python 2.0.1" name="software"> this software ("Python")</alt>
+               in source or binary form and its associated documentation.
+            </item>
+            <item>
+               <bullet>2.</bullet>
+               Subject to the terms and conditions of this License Agreement,
+               PSF hereby grants Licensee a nonexclusive, royalty-free,
+               world-wide license to reproduce, analyze, test, perform
+               and/or display publicly, prepare derivative works, distribute,
+               and otherwise use Python <alt match="2.0.1|3.10.6| " name="version">2.0.1</alt> alone or in any derivative
```
swinslow, line 33: Here and below, I think this would be:

`match="(([0-9]+)\.([0-9]+)\.([0-9]+))?"`

I think this would match either nothing (no version given) or a `1.2.3` style of version numbering.

Side note, for the `name=""` attribute, I don't know whether these need to be different names for each instance here and below... try it and see, I guess  :shrug:

src/Python-2.0.1.xml
```diff
@@ -0,0 +1,284 @@
...
+<SPDXLicenseCollection xmlns="http://www.spdx.org/license">
+   <license licenseId="Python-2.0.1" name="Python Software
+   Foundation License 2.0.1" listVersionAdded="3.18">
+      <crossRefs>
+         <crossRef>https://www.python.org/download/releases/2.0.1/license/</crossRef>
+         <crossRef>https://docs.python.org/3/license.html</crossRef>
+         <crossRef>https://github.com/python/cpython/blob/main/LICENSE</crossRef>
+      </crossRefs>
+      <notes>
+         This is the stack of licenses used for Python version 1.6.1 onward.
```
swinslow, line 11: Just confirming -- this is for 1.6.1 onwards? If so, should the ID and name of the license be `Python-1.6.1` instead of `Python-2.0.1`? Apologies if this was already sorted out...

swinslow (commented):
Take a look at my comments inline -- see if these help resolve the issues?

src/Python-2.0.1.xml
```diff
@@ -0,0 +1,284 @@
...
+            <p>
+               PSF LICENSE AGREEMENT FOR PYTHON 3.10.6
+            </p>
+         </titleText>
+         <list>
+            <item>
+               <bullet>1.</bullet>
+               This LICENSE AGREEMENT is between the Python Software Foundation
+               ("PSF"), and the Individual or Organization ("Licensee")
+               accessing and otherwise using <alt match="this software ("Python")|Python 3.10.6|Python 2.0.1" name="software"> this software ("Python")</alt>
```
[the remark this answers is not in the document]
swinslow: This is probably subsumed by my next comment just below here...

src/Python-2.0.1.xml
jlovejoy: that would be logical... I'm not entirely sure on the "used for version onwards" to be honest, I think I got that from some of the research done above.
But 2.0.1 seems to be the "name" used in some places and it seems to also have come after the license stack we have as 2.0
I think the license version number and python version number don't necessarily match up. (numbering seems to be a bit "loose" in this context overall)

src/Python-2.0.1.xml
jlovejoy: good point, fixed

src/Python-2.0.1.xml
```diff
@@ -30,37 +30,37 @@
...
+               THE USE OF PYTHON <alt match="(([0-9]+)\.([0-9]+)\.([0-9]+))?" name="version">2.0.1</alt> WILL NOT INFRINGE ANY THIRD PARTY RIGHTS.
             </item>
             <item>
                <bullet>5.</bullet>
                PSF SHALL NOT BE LIABLE TO LICENSEE OR ANY OTHER
-               USERS OF PYTHON <alt match="2.0.1|3.10.6| " name="version">2.0.1</alt> FOR ANY INCIDENTAL, SPECIAL, OR
+               USERS OF PYTHON <alt match="(([0-9]+)\.([0-9]+)\.([0-9]+))?" name="version">2.0.1</alt> FOR ANY INCIDENTAL, SPECIAL, OR
                CONSEQUENTIAL DAMAGES OR LOSS AS A RESULT OF MODIFYING,
-               DISTRIBUTING, OR OTHERWISE USING PYTHON <alt match="2.0.1|3.10.6| " name="version">2.0.1</alt>, OR ANY DERIVATIVE
+               DISTRIBUTING, OR OTHERWISE USING PYTHON <alt match="(([0-9]+)\.([0-9]+)\.([0-9]+))?"" name="version">2.0.1</alt>, OR ANY DERIVATIVE
```
swinslow, line 62: @jlovejoy looks like it's two quotes by mistake at the end of the `match` attribute, this might be causing the error?

src/Python-2.0.1.xml
jlovejoy: added better note on this

src/Python-2.0.1.xml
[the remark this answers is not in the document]
jlovejoy: updated

src/Python-2.0.1.xml
jlovejoy: tried other solution below

src/Python-2.0.1.xml
jlovejoy: fixed

src/Python-2.0.1.xml
```diff
@@ -0,0 +1,283 @@
...
+            <p>
+               PSF LICENSE AGREEMENT FOR PYTHON 3.10.6
+            </p>
+         </titleText>
+         <list>
+            <item>
+               <bullet>1.</bullet>
+               This LICENSE AGREEMENT is between the Python Software Foundation
+               ("PSF"), and the Individual or Organization ("Licensee")
+               accessing and otherwise using <alt match="this software \(\"Python\"\)|Python(( [0-9]+)\.([0-9]+)\.([0-9]+))?" name="software"> this software ("Python")</alt>
```
swinslow, line 23: Ah -- so this is failing because the quotation marks within the `match` attribute, even though they're escaped for purposes of the regex, the XML itself interprets them as ending the string. I'm digging to see if I can figure out an alternative approach here  :)

swinslow:
OK -- I'm hopeful on this one...

swinslow:
@jlovejoy Solved the quote marks issue, but some other bits are failing. I'll clean it up before merging prior to releasing 3.18.

swinslow:
So, even though under the matching guidelines the `titleText` should be disregarded, the license-list-XML validator requires it to be the same as the test text file. Including the row-of-dashes lines, as separate `<p>` paragraphs.  :)

brettcannon:
> Please, PSF, do it this way from now on!

Do which specifically? Leave off a version number or specify one?

jlovejoy:
> > Please, PSF, do it this way from now on!
> 
> Do which specifically? Leave off a version number or specify one?

leave off version number :)

jlovejoy:
@swinslow - it's passing now!! yeah! want to merge then :)

swinslow (approved):

swinslow:
@jlovejoy Yup -- merging! Thanks everyone for your help in getting this one over the line.

ockham:
Amazing, thanks all! 🎉

PR #1554, opened 2022-08-05
mpich2 and variants

jlovejoy:
add markup to mpich2 in author section

in response to #1292 

Signed-off-by: Jilayne Lovejoy

PR #1564, opened 2022-08-11
Add markup to mpich2

swinslow:
Replaces mpich2 portion of #1554 

Something about the original notice language didn't like being included within an `<alt>` tag for some reason, so replacing it with another example notice from use in the wild.


swinslow:
@jlovejoy After the tests pass, can you +1 and/or merge this? Thanks!

jlovejoy (approved):

PR #1573, opened 2022-08-11
Update License-fields.md

### LINKED ISSUE #1451: documentation: short identifier field - add bullet about "exception"
discussed on legal call 4/28 - @karsten-klein asked about short ids for exceptions. We looked at descriptions of short names at https://github.com/spdx/license-list-XML/blob/master/DOCS/license-fields.md and it doesn't mention this.

We should probably add a bullet saying that in most cases, the exception id should include the word "exception" since that is how we've been operating.

Any other suggestions for updates??

jlovejoy:
fixes #1451


jlovejoy:
There may be other more general wording improvement that could be made here, now that I'm looking at this.

swinslow (approved):

PR #1581, opened 2022-08-22
MIT-CMU: Adding optional "prior" permission

### LINKED ISSUE #1578: MIT-CMU/PIL - "prior" not included
I found the conversations about adding PIL, but the markup for MIT-CMU was updated instead based on similarities. However, "prior" was left out in `without specific, written prior permission`, which is in PIL (in the link used as an example). It was pointed out https://lists.spdx.org/g/Spdx-legal/message/2519 but not included. Is there a reason why? If not, can it be added as optional text so it will match?

capfei:
Added "prior" with optional tag.

Fixes #1578


swinslow (approved):

swinslow:
LGTM, thank you @capfei!

PR #1584, opened 2022-08-24
HPND-sell-variant: add XML markup

jlovejoy:
adding XML markup to accommodate "appear" or "appears"

PR #1588, opened 2022-08-24
Fix alt regex pattern in HPND-sell-variant

goneall:
The current `<alt match ... name="period">` has an incorrect regex that will not match the pattern `.,` which is present in the test file.  This will cause test failures once the LicenseListPublisher is updated with https://github.com/spdx/Spdx-Java-Library/pull/97 

This is the same fix as https://github.com/spdx/license-list-XML/pull/1587


jlovejoy:
@goneall - I have a PR on this same license at #1584 - is merging that going to mess this up?

@swinslow - FYI

goneall:
@jlovejoy 

> I have a PR on this same license at https://github.com/spdx/license-list-XML/pull/1584 - is merging that going to mess this up?

Maybe - depends on how smart git/github is ;)  If your PR is approved, go ahead and merge that first and I'll fix this one up if needed.

PR #1593, opened 2022-08-25
Fix APSL-2.0 exhibit text to match test code

goneall:
This was not caught previously due to an issue with the LicenseListPublisher.

### DIFF
--- a/src/APSL-2.0.xml
+++ b/src/APSL-2.0.xml
@@ -404,7 +404,7 @@
       <optional>
          <p>EXHIBIT A.</p>
          <standardLicenseHeader>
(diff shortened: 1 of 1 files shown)


jlovejoy:
@goneall - I noticed this tag, <optional spacing="none"> - which I don't think I've seen before. what is the spacing="none" for?

goneall:
@jlovejoy 

> what is the spacing="none" for?

this is just to control the display - the default is to put a space before and after optional text.  If you use `spacing="none"` it will remove any spacing.

We have to use this type of attribute/tagging since XML kind of ignores most white space.

BTW - I didn't modify the spacing - just adding a missing word for this PR.

PR #1594, opened 2022-08-25
Fix matching problem for APL-1.0.xml

goneall:
Match the number of underscores to the test file for the form included in this license.

Note - we may want to update the license matching guidelines and the license matching software to cover this case. 

As an alternative to this PR, we could create an `alt` tag with any number of underscores.  If someone wants to take this on, please submit a PR and I'll close this one.

Note: once we update the licenseListPublisher for the Github actions, this license will fail the comparison test if this PR (or alternative) is not merged.

### DIFF
--- a/src/APL-1.0.xml
+++ b/src/APL-1.0.xml
@@ -812,7 +812,7 @@
 	           <item>
 	              <bullet>PART 1:</bullet> INITIAL CONTRIBUTOR AND DESIGNATED WEB SITE
 <p>The Initial Contributor is: <br/>
-________________________________________________ <br/>
+____________________________________________________ <br/>
 [Enter full name of Initial Contributor]<br/>
                </p>
                <p>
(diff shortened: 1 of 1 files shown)


jlovejoy:
added new issue re: underscores, but merging this in meantime

jlovejoy:
discussed on 10/27 legal call. agreed this is a good thing to add in terms of matching. Came up with some implementation questions to discuss with @goneall 
* question as to how to implement matching - needs to be more than on dash, for example, to ensure it's not confused with an actual hyphen
* do we need to cover these 3 variants: ----  ====== ______  or may something broader such as a sequence of non-characters?
* we have done this using the optional tag in some cases, so is goal to have a standalone rule, so we don't have to do optional tag (or any tag) but simply a guidelines that is implemented in tool to "if you see 2 or more non-letter characters"

PR #1615, opened 2022-08-29
Removing 0BSD license from Python 2.0.1 test file

### LINKED ISSUE #1583: 0BSD license in Python-2.0.1.txt
Using the SPDX License Diff, I noticed that it is including the 0BSD license as part of the match, even though it is not included in the xml file. I did see that it is in the test file. Is it supposed to be there?

https://github.com/spdx/license-list-XML/blob/15b3ed7b817b39005d307072d9becb51d6f7b624/test/simpleTestForGenerator/Python-2.0.1.txt#L195-L207

capfei:
Removing 0BSD license from Python 2.0.1 test file.

Closes #1583


bsdimp:
I'd retitle to add 'from Python 2 license' or something.

PR #1620, opened 2022-09-02
Added FPA License

William-McGonagle:


jlovejoy:
Hi - I think you are intending to submit a license for consideration for the SPDX License List via the PR? Can you please have a look at our process document for this here: https://github.com/spdx/license-list-XML/blob/main/DOCS/request-new-license.md

You don't need to use the Online Tool at this point, but we do still need an issue with all the required information as listed in that form. Thanks!

PR #1621, opened 2022-09-02
Fix bad match pattern GFDL-1.1-invariants-or-later

goneall:
There is an extra linefeed in the middle of a match pattern which
will cause the GFDL-1.1-invariants-or-later license to no longer
match once the next version of the licenseListPublisher is released


goneall:
@swinslow Could you review and merge - this is the last issue for fixing the license list publisher to properly test the last optiona text.

zvr (approved):

Issue #1600, opened 2022-08-25
Add docs re: recommendation on using license IDs


swinslow:
Placeholder issues to discuss:

Consider whether to add documentation recommending e.g. "SPDX encourages the use of identifiers in SPDX short-form identifiers after a PR has been merged, but prior to the quarterly release" -- if this is a position we want to take.

swinslow:
Note that I've added [documentation on the release process](https://github.com/spdx/license-list-XML/blob/main/DOCS/release-process.md) for 3.19.

The "Overview" part at the top includes some generally neutral details on license IDs during the interim period between merging a PR and the next actual release, so this might be a good place to update if there is additional "guidance" for use of IDs during this interim.

jlovejoy:
@swinslow 
I just realized I had added this to the bottom of https://github.com/spdx/license-list-XML/blob/main/DOCS/request-new-license.md

> NOTE: Once a license has been accepted and to be added to the SPDX License List and the associated files merged, it is fine to start using the SPDX identifier, even if the official release has not yet occurred.

should I update that to be more consistent with what you said in the release doc?
or can we call this good/covered and close this issue?

swinslow:
I guess I don't love "it is fine to start using the SPDX identifier" as the text... it really kind of depends on the user's use case and what they're wanting to convey. If they want to use it in an SPDX document, then it isn't going to be fine for that purpose.

Can we maybe replace that with a link to https://github.com/spdx/license-list-XML/blob/main/DOCS/release-process.md#overview, with text something like the following? "See the [license list release process documentation](https://github.com/spdx/license-list-XML/blob/main/DOCS/release-process.md) for more details about the timing of releases and use of new IDs prior to a release."

That way we at least aren't repeating ourselves and we're just saying it in one place...?

jlovejoy:
@swinslow - see #1842

swinslow:
Thanks -- merged!

PR #1625, opened 2022-09-08
Add CI job to comment on accepted license requests

seabass-labrax:
This commit adds a GitHub Actions workflow file to leave comments on any 'new license/exception request' issues which have been finalised (that is, the issue has been closed and labelled as accepted).

Please see <https://lists.spdx.org/g/Spdx-legal/message/3225> for further context.

### DIFF
--- a/auto-comment.yaml
+++ b/auto-comment.yaml
@@ -1,7 +1,7 @@
 # SPDX-License-Identifier: 0BSD OR CC0-1.0
 
 repo: spdx/license-list-XML
-startAt: 2022-12-01
+startAt: 2023-02-23
 pullRequests: false
 filters:
   - type: issue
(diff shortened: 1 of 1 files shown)


auto-comment.yaml
```diff
@@ -0,0 +1,28 @@
...
+
+actions:
+  - type: add_comment
+    body: >
+      This new license/exception request has been accepted and the information
+      for the license/exception has been added to the repository. Thank you to
+      everyone who has participated!
+
+      The license/exception will be published at https://spdx.org/licenses/ as
+      part of the next SPDX License List release, which will occur no longer
```
goneall, line 23: Since we sometimes don't make the release dates precisely, could we soften this statement a bit?  Something like `... which is targeted to be released in 3 months or less`.  Not a big deal since the notice will go out on average 6 weeks before a release.

goneall (commented):
Thanks @seabass-labrax for pulling this together!

Just one minor suggestion.

auto-comment.yaml
seabass-labrax: @goneall, thank you for the suggestion; I've pushed an additional commit.

seabass-labrax:
@swinslow, it would be nice to have this running sooner rather than later, so would you mind merging this PR? #1600 is still open for discussion, so #1696 should also be left open; we can iterate on the contents of the message to reflect what's decided later on.

swinslow:
Discussed on 2023-02-23 legal team call, agreed to add -- will merge after the tests finish re-running!

PR #1628, opened 2022-09-14
Add mark-up to BSD-3-Clause

jlovejoy:
for grammar difference - see "BSD-3-clause-author" listed here: https://github.com/freedesktop/libbsd/blob/master/COPYING

### DIFF
--- a/src/BSD-3-Clause.xml
+++ b/src/BSD-3-Clause.xml
@@ -10,7 +10,7 @@
         <p>Copyright (c) &lt;year&gt; &lt;owner&gt;. </p>
       </copyrightText>
 
-        <p>Redistribution and use in source and binary forms, with or without modification, are permitted provided
+        <p>Redistribution and use in source and binary forms, with or without modification, <alt match="are|is" name="tobe"> permitted provided
          that the following conditions are met:</p>
         <list>
             <item>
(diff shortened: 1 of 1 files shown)


bsdimp:
Despite some of the checks failing for what appears to be compile reasons, this looks good to me. I've seen both variations. I can't imagine this one word difference changing anything at all, but ianal.

bsdimp:
I've seen these variations, and this looks good now.

jlovejoy:
I see it as a matter of poor grammar :) 

@swinslow - do you want to weigh in? do we need any further discussion beyond in this issue for this one?

richardfontana:
What's perplexing is that someone must have deliberately chosen to change "are" to "is" (because the "are" version had been in wide use for many years by the time this variant started being used, presumably), perhaps with a sincere belief that "is" was actually more grammatically correct. I think that's only true if you think of "redistribution and use" as being like "ham 'n' eggs" or "rock 'n' roll".

bsdimp:
There were a number of well meaning people that liked to 'fix' grammar that tried to do these things from time to time in the various BSD projects (and likely outside of them too, people being what they are). It's a completely irrational fetish that the different projects have had to police from time to time over the years... I've also seen some come from well-meaning corporate lawyers who think the disclaimer needs an extra word or three (or has a few extra words), though it's been a long time since I saw that asked for (most recently by Cavium with their community Octeon toolkit which had rewritten it)...  Though this is getting a bit far away from the markup, which still looks good to my eye :)

swinslow:
@jlovejoy I see this fixes #1632, I think we're good to merge given the discussion there and here, thanks!

PR #1641, opened 2022-09-22
faq: update to lates identifiers

bsdimp:
GPL-2.0 is depreated, move to GPL-2.0-only or GPL-2.0-or-later instead (these are examples, so use both consistently in different examples).


swinslow (approved):
LGTM, thanks @bsdimp :)

PR #1677, opened 2022-10-13
Add documentation on XML templates and fields

### LINKED ISSUE #1289: document regular expression examples
example with GCC exception where could be "give(s)?" or "give|gives" to get same outcome

might be good to have list of example of markup with explanation on where used (instead of having to find a license with similar markup to copy)

### LINKED ISSUE #1369: Review documentation re: XML formatting
@swinslow to review existing docs to see what we've got re: XML formatting for licenses in the list. This is a spinout of the discussion at https://github.com/spdx/license-list-XML/issues/576

swinslow:
Fixes #1369 

Fixes #1289

### DIFF
--- a/DOCS/xml-fields.md
+++ b/DOCS/xml-fields.md
@@ -49,7 +49,8 @@ There are two mandatory attributes for every `<license>` and `<exception>` tag:
 
 There are two additional attributes which are optional, but highly encouraged:
 * `isOsiApproved` (for licenses, not exceptions): either "true" or "false" based on whether this license has been approved by the [Open Source Initiative](https://opensource.org/licenses/alphabetical)
-* `listVersionAdded`: in which release version of the SPDX License List was the license first added, e.g. "3.19"
+* `listVersionAdded`: in which release version of the SPDX License List was the license first added, e.g. "3.19".
+  * Typically you'll check the currently-released version at https://spdx.org/licenses/ and increment the minor version by 1 for a new license.
 
 Finally, if the license ID has been deprecated, two additional attributes should be included:
 * `isDeprecated`: should be "true" if the license ID is deprecated
@@ -57,12 +58,14 @@ Finally, if the license ID has been deprecated, two additional attributes should
 
 Note that the deprecated tags refer to whether the ID has been deprecated _by SPDX_ -- in other words, if the SPDX Legal Team no longer recommends that the identifier should be used. Licenses that have been described as "deprecated" or "superseded" by the _license author_ might no longer be recommended to use for new code, but their identifiers remain valid License IDs on the SPDX License List.
 
+Deprecated license identifiers will be listed at the bottom of the SPDX License List page at https://spdx.org/licenses/.
+
 ### License metadata
 
 There are three optional first-level metadata fields that can be nested immediately under `<license>` or `<exception>`, each of which is optional:
 
-* **`<crossRefs>`**: Defines cross-references to URLs where the license can be found in the wild
-* **`<notes>`**: Describes general comments about the license
+* **`<crossRefs>`**: Defines cross-references to URLs where the license can be found in use by one or a few projects, and (if applicable) where posted by the license steward
+* **`<notes>`**: Describes general comments about the license; if deprecated, also briefly explain the reason for deprecating the license identifier
 * **`<obsoletedBys>`** (if license is deprecated): Lists the license(s) that should be used instead of this deprecated license
 
 ### License text
@@ -163,7 +166,7 @@ The **<alt>** tag includes two attributes:
 * `name`: specifies a unique name for this matching element
 * `match`: specifies a [POSIX extended regular expression (ERE)](http://pubs.opengroup.org/onlinepubs/9699919799/) for what text will match
 
-Note that an example of actually-matching text should be included between the `<alt>` and `</alt>` tags, as the specified example is what will be rendered on the SPDX License List website page.
+Note that an example of actually-matching text should be included between the `<alt>` and `</alt>` tags, as the specified example is what will be rendered on the SPDX License List website page and in license text files provided by SPDX. To the extent a license has a "canonical" or "most commonly used" standard language, use that in the match between the `alt` tags.
 
 Learning regular expressions is outside the scope of this document. However, here are a few of the most common patterns, taken from [BSD-3-Clause](https://github.com/spdx/license-list-XML/blob/main/src/BSD-3-Clause.xml) which is a good example of using several `<alt>` tags effectively:
 * `<alt match="are|is" name="tobe">are</alt>`: Matches if the text is either the word "are" or the word "is"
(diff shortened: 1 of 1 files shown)


DOCS/xml-fields.md
```diff
@@ -0,0 +1,177 @@
...
+</SPDXLicenseCollection>
+```
+
+### `<license>` or `<exception>` tag
+
+The main tag used to define the license or exception is, unsurprisingly, **`<license>`** or **`<exception>`**. All of the remaining content will be enclosed within a `<license></license>` or `<exception></exception>` pair of tags.
+
+There are two mandatory attributes for every `<license>` and `<exception>` tag:
+* `licenseId`: the unique SPDX Identifier for the license; should be identical to the filename
+* `name`: the longer "real name" of the license
```
bsdimp, line 48: Does the 'real name' also need to be unique?

DOCS/xml-fields.md
```diff
@@ -0,0 +1,177 @@
...
+
+There are two additional attributes which are optional, but highly encouraged:
+* `isOsiApproved` (for licenses, not exceptions): either "true" or "false" based on whether this license has been approved by the [Open Source Initiative](https://opensource.org/licenses/alphabetical)
+* `listVersionAdded`: in which release version of the SPDX License List was the license first added, e.g. "3.19"
```
bsdimp, line 52: Do we need to say what to put here for new submissions?

DOCS/xml-fields.md
```diff
@@ -0,0 +1,177 @@
...
+
+Finally, if the license ID has been deprecated, two additional attributes should be included:
+* `isDeprecated`: should be "true" if the license ID is deprecated
+* `deprecatedVersion`: in which release version of the SPDX License List was the license first marked as deprecated, e.g. "3.19"
+
+Note that the deprecated tags refer to whether the ID has been deprecated _by SPDX_ -- in other words, if the SPDX Legal Team no longer recommends that the identifier should be used. Licenses that have been described as "deprecated" or "superseded" by the _license author_ might no longer be recommended to use for new code, but their identifiers remain valid License IDs on the SPDX License List.
```
bsdimp, line 58: Do we need / want to have (invent?) details on (a) why spdx deprecated its recommendation to use it and (b) what to use instead?

DOCS/xml-fields.md
```diff
@@ -0,0 +1,177 @@
...
+
+### License metadata
+
+There are three optional first-level metadata fields that can be nested immediately under `<license>` or `<exception>`, each of which is optional:
+
+* **`<crossRefs>`**: Defines cross-references to URLs where the license can be found in the wild
```
bsdimp, line 64: "in the wild" is a bit informal.
"the license can be found in use by an open source project" or similar maybe?

DOCS/xml-fields.md
```diff
@@ -0,0 +1,177 @@
...
+* **`<notes>`**: Describes general comments about the license
```
bsdimp, line 65: Should we start adding the reason for deprecation here? or suggest that...

DOCS/xml-fields.md
```diff
@@ -0,0 +1,177 @@
...
+### Section annotation markers
+
+Some portions of the license text can be surrounded by annotation tags, to denote certain sections of the license text:
+
+* **`<titleText>`**: indicates the title of the license, if specified in the license text itself
+* **`<copyrightText>`**: indicates where a copyright notice would be placed for the licensed code
+  * Note that `<copyrightText>` should _not_ be used for copyright notices that apply to the copyright in the license text itself!
+* **`<standardLicenseHeader>`**: indicates a standard way that the license recommends specifying the license, e.g. in code comments
+  * Note that `<standardLicenseHeader>` should _only_ be used if it is defined by the license text itself!
+  * Typically, this would be located in an appendix titled something like "How to apply the license to your work"
```
bsdimp, line 100: Maybe add 'Do not use it if the license text is included verbatim in the file' perhaps?

DOCS/xml-fields.md
```diff
@@ -0,0 +1,177 @@
...
+
+### Replaceable text
+
+Some licenses contain portions of language that can match to a variety of alternative texts. This is called "replaceable text", and is denoted with the **`<alt>...</alt>`** tag.
+
+The **<alt>** tag includes two attributes:
+* `name`: specifies a unique name for this matching element
+* `match`: specifies a [POSIX extended regular expression (ERE)](http://pubs.opengroup.org/onlinepubs/9699919799/) for what text will match
+
+Note that an example of actually-matching text should be included between the `<alt>` and `</alt>` tags, as the specified example is what will be rendered on the SPDX License List website page.
```
bsdimp, line 166: We should specify that the <alt match="RE">Some text</alt> "Some Text" will be rendered on the web pages, text files, etc and should be the canonical text, as much as is possible with some licenses....

bsdimp:
Despite my many nits and picky questions, I like this generally.

DOCS/xml-fields.md
johnmhoran, line 58: Maybe note that the deprecated info is available at the bottom of the `SPDX License List` page?

johnmhoran:
LGTM!  👍   (I added 1 minor comment/question.)

DOCS/xml-fields.md
swinslow: Technically we don't require it, e.g. some of the deprecated licenses share the same "name" with new identifiers that were added to replace them. `GPL-3.0` and `GPL-3.0-only` both have the "name" `GNU General Public License v3.0 only`.

DOCS/xml-fields.md
swinslow: Usually I think the "why deprecated" details get included in the "Notes" section of the license, if anywhere. E.g. https://spdx.org/licenses/wxWindows.html and https://spdx.org/licenses/StandardML-NJ.html

DOCS/xml-fields.md
swinslow: Yes, good call -- I'll update to highlight that also.

DOCS/xml-fields.md
swinslow: Sure -- I'll update to make it clearer.

DOCS/xml-fields.md
swinslow: I'm inclined to say that if we want to add more clarity about what SPDX considers to be a "standard license header", that would probably be more appropriate to add to other documentation / FAQ and then point to it from here. I'm hesitant to expand the scope of this document too much (I'm pretty wordy already as it is...)

DOCS/xml-fields.md
swinslow: Yup, good call. I'll update to note that too.

DOCS/xml-fields.md
swinslow: Good call -- can flag that briefly so folks know where to look.

swinslow:
@bsdimp and @johnmhoran Thanks for your feedback! I've just pushed 897bd28 which I think should address your comments. If you're good with this, grateful for your +1 or let me know any further concerns / tweaks.

swinslow:
Discussed on Legal Team call 2022-10-27, merging as-is and can iterate later. Need to link into README also.

PR #1678, opened 2022-10-13
New exception: x11vnc-opensel-exception

### LINKED ISSUE #1603: New license request: x11vnc-openssl-exception [SPDX-Online-Tools]
**1.** License Name: x11vnc OpenSSL Exception
**2.** Short identifier: x11vnc-openssl-exception
**3.** License Author or steward: Karl Runge (possibly)
**4.** Comments: This is a variety of OpenSSL GPL linking exception that is used by at least one project packaged in Fedora.
**5.** Standard License Header: 
**6.** License Request Url: http://tools.spdx.org/app/license_requests/156
**7.** URL(s): https://github.com/LibVNC/x11vnc/blob/master/src/8to24.c#L22
**8.** OSI Status: Not Submitted
**9.** Example Projects: https://github.com/LibVNC/x11vnc

jlovejoy:
fixes #1603

### DIFF
--- a/src/exceptions/x11vnc-openssl-exception.xml
+++ b/src/exceptions/x11vnc-openssl-exception.xml
@@ -1,0 +1,25 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<SPDXLicenseCollection xmlns="http://www.spdx.org/license">
+   <license isOsiApproved="false" licenseId="x11vnc-openssl-exception"
+   name="x11vnc OpenSSL Exception" listVersionAdded="3.19">
+      <crossRefs>
+         <crossRef>https://github.com/LibVNC/x11vnc/blob/master/src/8to24.c#L22</crossRef>
+      </crossRefs>
+     
+      <notes>This is very similar to openvpn-openssl-exception with a couple key differences.
+      </notes>
+      <text>
+         <p>
+            In addition, as a special exception, Karl J. Runge gives permission
+            to link the code of its release of x11vnc with the OpenSSL project's
+            "OpenSSL" library (or with modified versions of it that use the
+            same license as the "OpenSSL" library), and distribute the linked
+            executables. You must obey the GNU General Public License in all
+            respects for all of the code used other than "OpenSSL". If you
+            modify this file, you may extend this exception to your version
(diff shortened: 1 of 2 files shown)


src/exceptions/x11vnc-openssl-exception.xml
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<SPDXLicenseCollection xmlns="http://www.spdx.org/license">
+   <license isOsiApproved="false" licenseId="x11vnc-openssl-exception"
```
swinslow, line 3: I think exceptions need the tag to be "exception" rather than "license"...

swinslow: I'm also not sure whether `isOsiApproved` can (or should) be an attribute on an `exception` tag...?

src/exceptions/x11vnc-openssl-exception.xml
```diff
@@ -0,0 +1,25 @@
...
+            "OpenSSL" library (or with modified versions of it that use the
+            same license as the "OpenSSL" library), and distribute the linked
+            executables. You must obey the GNU General Public License in all
+            respects for all of the code used other than "OpenSSL". If you
+            modify this file, you may extend this exception to your version
+            of the file, but you are not obligated to do so. If you do not
+            wish to do so, delete this exception statement from your version.
+         </p>
+      </text>
+   </license>
```
swinslow, line 24: This would also change from `</license>` to `</exception>`.

swinslow (changes requested):
Hi @jlovejoy, I think that since this is an exception it needs the `<exception>` tag rather than `<license>` -- see my comments below...

src/exceptions/x11vnc-openssl-exception.xml
jlovejoy: good catch, thanks!

src/exceptions/x11vnc-openssl-exception.xml
jlovejoy: done

swinslow (approved):

PR #1685, opened 2022-10-27
License Inclusion Principles - add ordered list

jlovejoy:
add ordered list using letters to "definitive factors"  
move historical comment re: OSI to background section


jlovejoy:
LGTM now!

PR #1686, opened 2022-10-27
Add AFWL-1.0

MichaelAgarkov:
I would like to add my license if it can be added.


MichaelAgarkov:
Why is this pull request taking so long for review?

PR #1699, opened 2022-11-08
Remove spacing around philicenses alt text

### LINKED ISSUE #1697: GPL-3.0-only template text: extraneous whitespace around philicenses var near end
On [line 255 of GPL-3.0-only.template.txt](https://github.com/spdx/license-list-data/blob/master/template/GPL-3.0-only.template.txt#L255) (v3.18) there's a variable element within a URL, but the `<<var>>` is surrounded by whitespace, causing matches to fail:

**GPL-3.0-only.template.txt:**

`But first, please read <https://www.gnu.org/ <<var;name="philicenses";original="licenses";match="philosophy|licenses">> /why-not-lgpl.html>.` (note whitespace around the \<\<var\>\>)

[**Canonical License text:**](https://www.gnu.org/licenses/gpl-3.0.txt)

`But first, please read <https://www.gnu.org/licenses/why-not-lgpl.html>.` (note no whitespace around "licenses")

Note: it appears that this may be an issue with the logic that converts the source license list XML to the template text format, since this whitespace is not apparent in the [source XML](https://github.com/spdx/license-list-XML/blob/main/src/GPL-3.0-only.xml#L855).

goneall:
Fixes #1697


jlovejoy:
not sure what I'm reviewing here... but ready to merge?

swinslow (approved):

PR #1700, opened 2022-11-09
Add "and" and "&" to the equivalent words list

### LINKED ISSUE #1663: MS-LPL XML should accommodate 'and' instead of '&' in last section
Section 3(F) of the SPDX version of MS-LPL says: "(F) Platform Limitation- The licenses granted in sections 2(A) & 2(B) extend only to the software or derivative works that you create that run on a Microsoft Windows operating system product."

However, a number of uses of this license, including by Microsoft, use "2(A) and 2(B)" rather than "2(A) & 2(B)". 
https://gitlab.freedesktop.org/pulseaudio/webrtc-audio-processing/-/blob/v0.3/webrtc/LICENSE_THIRD_PARTY#L388
https://kb.acronis.com/system/files/content/2012/12/38416/mspublic.txt
https://marketplace.visualstudio.com/items/MikeBMcL.WindowsStoreDirectXGameTemplate/license

As I understand it, "&" and "and" are not equivalent under the SPDX matching guidelines.

swinslow:
Fixes #1663

### DIFF
--- a/equivalentwords.txt
+++ b/equivalentwords.txt
@@ -1,5 +1,6 @@
 acknowledgement,acknowledgment
 analog,analogue
+and,&
 analyze,analyse
 artifact,artefact
 authorization,authorisation


swinslow:
@jlovejoy and @goneall If/when the tests pass, grateful to get your +1 for "and" and "&"  :)

goneall (approved):
Seems like a reasonable update to me - I will need to update the tools - it doesn't pull from this file automatically.

jlovejoy (approved):

PR #1704, opened 2022-11-10
Beerware: add markup to email address

### LINKED ISSUE #1633: Beerware variant with different email address
I found this license:

```
"THE BEER-WARE LICENSE" (Revision 42):
 * <phk@login.dkuug.dk> wrote this file.  As long as you retain this notice you
 * can do whatever you want with this stuff. If we meet some day, and you think
 * this stuff is worth it, you can buy me a beer in return.   Poul-Henning Kamp
```
Am I correct that this does not match to SPDX Beerware because the email address is different?

jlovejoy:
fixes #1633 

not sure if can use <> within alt tag, match quotes or need to use &gt; and &lt;

### DIFF
--- a/src/Beerware.xml
+++ b/src/Beerware.xml
@@ -7,8 +7,8 @@
       </crossRefs>
 
     <text>
-      <p>"THE BEER-WARE LICENSE" (Revision 42):
-      <br></br>&lt;phk@FreeBSD.ORG&gt; wrote this file. As long as you retain
+       <p>"THE BEER-WARE LICENSE" (Revision 42):</p>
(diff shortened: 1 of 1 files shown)


jlovejoy:
@swinslow - can you merge this one?

swinslow:
@jlovejoy I admit I'm confused about the `&lt;` and `&gt;` being included within the quotation-marks string that's used to define the regular expression for the match.

That said, it's passing the tests, which appears to mean that it isn't breaking the tools or the match. So I'm going to go ahead and merge it. Thanks!

PR #1705, opened 2022-11-11
XPP: add markup for spelling error

jlovejoy:
markup for University or Univeristy, plus and/or


swinslow (changes requested):
I'm not certain why this is failing. I suspect it _might_ have something to do with embedding the `<alt>` tags in the middle of a quoted string in the XML file.

If so, then this might be fixable by changing each `"` mark to `&quot;` in the relevant bits surrounding the "Indiana University" parts of the text.

jlovejoy:
thanks @swinslow - trying that...

swinslow:
Hmm... still appears to be failing. I think we're going to need more time to sort this out, so I'm moving it to 3.20.

swinslow:
OK, so I think this one has been failing because of the same problem with the Validate schema tests that we fixed last week. I'll update this PR so we can re-run those updated tests...

swinslow (approved):

PR #1709, opened 2022-11-16
Fix formatting of Artistic-1.0-cl8

mgorny:
Fix the formatting of Artistic-1.0-cl8 to list the 8th clause as a bullet point rather than an additional paragraph to the 7th clause.

Signed-off-by: Michał Górny <mgorny@gentoo.org>


jlovejoy:
Thanks @mgorny !

Issue #1676, opened 2022-10-13
Inconsitency in file license-list-data/json/licenses.json


vargenau:
In file ```license-list-data/json/licenses.json``` you have:

```
    {
      "reference": "https://spdx.org/licenses/GPL-3.0.html",
      "isDeprecatedLicenseId": true,
      "detailsUrl": "https://spdx.org/licenses/GPL-3.0.json",
      "referenceNumber": 242,
      "name": "GNU General Public License v3.0 only",
      "licenseId": "GPL-3.0",
      "seeAlso": [
        "https://www.gnu.org/licenses/gpl-3.0-standalone.html",
        "https://opensource.org/licenses/GPL-3.0"
      ],
      "isOsiApproved": true,
      "isFsfLibre": true
    },

```

```
    {
      "reference": "https://spdx.org/licenses/GPL-3.0-only.html",
      "isDeprecatedLicenseId": false,
      "detailsUrl": "https://spdx.org/licenses/GPL-3.0-only.json",
      "referenceNumber": 274,
      "name": "GNU General Public License v3.0 only",
      "licenseId": "GPL-3.0-only",
      "seeAlso": [
        "https://www.gnu.org/licenses/gpl-3.0-standalone.html",
        "https://opensource.org/licenses/GPL-3.0"
      ],
      "isOsiApproved": true,
      "isFsfLibre": true
    },
```

In short, you have the same "name" for both the deprecated license and the replacement license.

This is true for:
```
             "name": "GNU General Public License v1.0 only",
             "name": "GNU General Public License v1.0 or later",
             "name": "GNU General Public License v2.0 only",
             "name": "GNU General Public License v2.0 or later",
             "name": "GNU General Public License v3.0 only",
             "name": "GNU General Public License v3.0 or later",
             "name": "GNU Lesser General Public License v2.1 only",
             "name": "GNU Lesser General Public License v3.0 only",
             "name": "GNU Lesser General Public License v3.0 or later",
             "name": "GNU Library General Public License v2 only",
             "name": "GNU Library General Public License v2 or later",
             "name": "Standard ML of New Jersey License",
```
But it is not the case for other deprecated licenses.

For example, you have:
```
   {
      "reference": "https://spdx.org/licenses/LGPL-2.1-or-later.html",
      "isDeprecatedLicenseId": false,
      "detailsUrl": "https://spdx.org/licenses/LGPL-2.1-or-later.json",
      "referenceNumber": 31,
[...]

PR #1712, opened 2022-11-16
AGPL-1.0.xml: fix license name

vargenau:
Change license name from "Affero General Public License v1.0" to "Affero General Public License v1.0 only". Then we have the same license name for AGPL-1.0.xml and AGPL-1.0-only.xml (as it is already the case for the GPL licenses).

Signed-off-by: Marc-Etienne Vargenau <marc-etienne.vargenau@nokia.com>

### DIFF
--- a/src/AGPL-1.0.xml
+++ b/src/AGPL-1.0.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <SPDXLicenseCollection xmlns="http://www.spdx.org/license">
    <license isOsiApproved="false" licenseId="AGPL-1.0"
-            name="Affero General Public License v1.0"
+            name="Affero General Public License v1.0 only"
             isDeprecated="true" deprecatedVersion="3.1">
       <obsoletedBys>
          <obsoletedBy>AGPL-1.0-only</obsoletedBy>


swinslow (approved):

swinslow (commented):
Actually, although I approved this a moment ago, in retrospect I'm not certain about this one.

Part of the reason that `GPL-1.0`, `GPL-2.0`, etc. had `only` in their name is because originally there were separate `GPL-1.0+`, `GPL-2.0+`, etc. identifiers actually on the license list. I don't know the history but I assume those predated `+` being an operator in the license expression syntax.

Put another way, the deprecated "AGPL-1.0" identifier could have correctly had `+` appended to the end of it, so I'm not sure whether the intent was for `AGPL-1.0` to have "only" appended to the end of its name.

@jlovejoy Do you have any thoughts here? Personally I don't see a particular problem with adding "only" now, but didn't want to do something that might have already been considered in the past.

jlovejoy:
ok, so the GFDL licenses that @vargenau has updated here were deprecated as of 3.0
but you can see the family of LGPL and GPL licenses have been deprecated (and updated) at two points in the past: 2.0rc2 and 3.0
this is b/c prior to 2.0rc2 - we didn't have operators, so we had GPL-2.0 and GPL-2.0+ on the list. With the addition of operators (e..g, the + operator), GPL-2.0+ (and the like) were deprecated, but GPL-2.0 remained with the same meaning (only).

In the meantime, the GFDL didn't have this same evolution. I'd have to dig back in the archives, but I don't believe we ever had GFDL-x.y+ on the list. But as of 3.0, we updated ALL the GNU Licenses and added the "invariants" options for GFDLs as well.

In any case, while I appreciated @vargenau diligence here, I don't think we should update deprecated licenses. It's sort of irrelevant now, and in this case, it was all a bit unclear previously, so it would feel a bit like re-writing history.

swinslow:
Thanks all -- closing as described at https://github.com/spdx/license-list-XML/issues/1676#issuecomment-1364537861.

vargenau:
I agree with the reasoning.

Best wishing to all for 2023.

PR #1714, opened 2022-11-16
GFDL-1.1.xml: fix license name

vargenau:
Change license name from "GNU Free Documentation License v1.1" to "GNU Free Documentation License v1.1 only". Then we have the same license name for GFDL-1.1.xml and GFDL-1.1-only.xml (as it is already the case for the GPL licenses).

Signed-off-by: Marc-Etienne Vargenau <marc-etienne.vargenau@nokia.com>

### DIFF
--- a/src/GFDL-1.1.xml
+++ b/src/GFDL-1.1.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <SPDXLicenseCollection xmlns="http://www.spdx.org/license">
   <license licenseId="GFDL-1.1" isOsiApproved="false"
-  name="GNU Free Documentation License v1.1"
+  name="GNU Free Documentation License v1.1 only"
   isDeprecated="true" deprecatedVersion="3.0">
     <obsoletedBys>
       <obsoletedBy>GFDL-1.1-only</obsoletedBy>


swinslow (commented):
Same comment here as for https://github.com/spdx/license-list-XML/pull/1712#pullrequestreview-1220030538 -- and the same will apply for #1715 and #1716. Let's see what @jlovejoy has to say once she has a chance to review...

jlovejoy:
ugh, I put this comment in the wrong issue! fixing now...

ok, so the GFDL licenses that @vargenau has updated here were deprecated as of 3.0
but you can see the family of LGPL and GPL licenses have been deprecated (and updated) at two points in the past: 2.0rc2 and 3.0
this is b/c prior to 2.0rc2 - we didn't have operators, so we had GPL-2.0 and GPL-2.0+ on the list. With the addition of operators (e..g, the + operator), GPL-2.0+ (and the like) were deprecated, but GPL-2.0 remained with the same meaning (only).

In the meantime, the GFDL didn't have this same evolution. I'd have to dig back in the archives, but I don't believe we ever had GFDL-x.y+ on the list. But as of 3.0, we updated ALL the GNU Licenses and added the "invariants" options for GFDLs as well.

In any case, while I appreciated @vargenau diligence here, I don't think we should update deprecated licenses. It's sort of irrelevant now, and in this case, it was all a bit unclear previously, so it would feel a bit like re-writing history.

swinslow:
Given the comment at https://github.com/spdx/license-list-XML/pull/1714#issuecomment-1362375350, I'm going to close this PR. Thanks everyone.

Issue #902, opened 2019-07-14
New license request: Commons-Clause-1.0 [SPDX-Online-Tools]


bb010g:
**1.** License Name: Commons Clause License Condition v1.0
**2.** Short identifier: Commons-Clause-1.0
**3.** License Author or steward: https://fossa.com/
**4.** Comments: The Commons Clause should be included because it significantly restricts the freedom of commercial users of otherwise OSI-compliant software, including making it no longer OSI compliant.

The Commons Clause was used by Redis Labs for a period (e.g. in RediSearch from 2018-07-15 ( https://github.com/RediSearch/RediSearch/blob/c516603c0705ccd14efc1623c4702cba63b8d262/LICENSE ) to 2019-02-21 ( https://github.com/RediSearch/RediSearch/blob/532437185522a9338559034cbdc372c1cb91519c/LICENSE ) and in RedisGraph from 2018-07-16 ( https://github.com/RedisGraph/RedisGraph/blob/0bcb957f75955bef3dbcf920dae57bd741b6c472/LICENSE ) to 2019-02-21 ( https://github.com/RedisGraph/RedisGraph/blob/7c1c912d2bb0970032e5b211e8a83ce4d875036d/LICENSE )), and is currently used by various projects, including (found by searching GitHub for "Commons Clause"): https://github.com/exlskills/spf-server/blob/96c2b8fa9adbe005967cc85983269e2d01c6d665/LICENSE.md , https://github.com/cityseer/cityseer/blob/9c3a1c700f656ae6fb3aab4b3b856958ed2d081a/LICENSE , https://github.com/ITfoxtec/Sortingtime/blob/44f429070abefeee10f6894fd2014819cc3632a1/LICENSE .
**5.** Standard License Header: 
**6.** URL: https://commonsclause.com/
**7.** OSI Status: Rejected

jlovejoy:
This is not a standalone license, but written to be "added" to an existing license. Our exception list defines things that are included there as, "These exceptions grant an exception to a license condition or additional permissions beyond those granted in a license; they are not stand-alone licenses." This does not grant an additional permission or an exception to a condition in the existing license, but it a clear restriction on rights granted by license.

While our license inclusion principles do not require strict compliance with the open source definition, this doesn't even come close. 

For the reasons above, I would vote to NOT add this to the SPDX License List.

[...]

PR #1738, opened 2022-12-10
Create OSI-SPDX-history.md

jlovejoy:
beginning commit for background and logging issues that were (or were not) from when SPDX sought to include every OSI-approve license on the SPDX License List

More research and commits are needed, and will take time!

### DIFF
--- a/DOCS/OSI-SPDX-history.md
+++ b/DOCS/OSI-SPDX-history.md
@@ -1,17 +1,15 @@
 # A History of OSI-approved licenses and the SPDX License List
 
-In the early days it was decided that the SPDX License List should include all licenses that were ever OSI-approved licenses. The rationale being that once OSI-approved, always OSI-approved and that even old or deprecated licenses may still appear "in the wild". 
+Early on, the SPDX License List should include all licenses that were ever OSI-approved licenses. The rationale being that once OSI-approved, always OSI-approved and that even old or "deprecated" (by way of a newer version or other such circumstance) licenses may still appear "in the wild". (Note: the OSI and SPDX use "deprecated" in different ways.)
 
-This involved early collaboration between SPDX and OSI starting around 2011 and onward. As the SPDX-legal team sought to accurately identify all OSI-approved licenses and their corresonpding text to ensure that representation on the SPDX License List. This involved cross-collaboration between SPDX-legal and various OSI representagives, with most communication occurring on the SPDX-legal mailing list, OSI's license-disucss mailing list, or both. 
+In 2011, OSI [began using the SPDX license identifiers](https://opensource.org/minutes20110608) in their URLs and on the license pages.
 
-At that time, OSI began using the SPDX license identifiers in their URLs and on the license pages.
+The SPDX-legal team sought to accurately identify all OSI-approved licenses and their corresponding text to ensure representation on the SPDX License List. This involved cross-collaboration between SPDX-legal and various OSI representatives, with most communication occurring on the SPDX-legal mailing list, OSI's license-disucss mailing list, or both. John Cowan, Martin Michlmayr, Karl Fogel, and Luis Villa were among the OSI board members who helped Jilayne Lovejoy from SPDX-legal on researching the various issues. 
 
-This page seeks to explain some of the challenges that this reconciliation work encountered, especially where a solution to the challenge may still leave questions for people currently looking at or using both the SPDX License List and the OSI list of approved licenses. 
-
-
+This page seeks to explain some of the challenges that this reconciliation work encountered, especially where a solution to the challenge may still leave questions for people currently looking at or using both the SPDX License List and the OSI list of approved licenses. Any suggestions that had been made in the past for a possible reconciliation have been reproduced here as well. 
 
 ## Old and "deprecated" licenses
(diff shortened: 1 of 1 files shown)


DOCS/OSI-SPDX-history.md
```diff
@@ -0,0 +1,104 @@
...
+* CVW: MITRE Collaborative Virtual Workspace License
+
+still need to research issue here
+
+# from Max's list - still to address: 
+## Not in OSI list, but in SPDX's list
+
+* GPL-3.0-with-GCC-exception
+
+what is the quesiton here?
```
mxmehl, line 73: @richardfontana explained the background of the WITH IDs:

> You list some cases of 'WITH' expressions. The OSI has been reluctant to approve license exceptions, except in a few special cases where the exception (or exception coupled with a standard license) is itself thought of as a single license (e.g. LGPL version 3; ec0s-2.0 is also like this). From my recollection of my time on the OSI board, the main concern was the potential numerosity of license submissions if the OSI encouraged submission of exceptions. There's been a tendency to assume that typical types of GPL exceptions are legit (for a GPL-world notion of legit) because they conform to the model of a grant of additional permission -- I need to comment on this issue on another recent thread.

That would resolve my issue with this ID as I can assume that an official SPDX expression containing `with` is very likely OSI-approved.

DOCS/OSI-SPDX-history.md
```diff
@@ -0,0 +1,104 @@
...
+
+this is a deprecated SPDX license id, not sure what the question is here?
+
+* MIT-Modern-Variant
+
+### False-positives
+
+- Many mismatches in GPL/LGPL/AGPL because of -only/-or-later/+. I unified those
+
+what do you mean my mismatches? and "unified those"?
```
mxmehl, line 85: Yes sorry, that was unclear. Of the GPL family, I treated `-only`, `-or-later`, and `+` the same as OSI does not use these. It just means that I did not look for `GPL-2.0-or-later` on OSI's website, but just for `GPL-2.0`. Effectively I found LGPL-2.0 as missing (see the comment above).

DOCS/OSI-SPDX-history.md
richardfontana: I don't think "OSI-approved" is the correct concept here. Rather, historically, the OSI might have informally said that "normal" (additional-permission-style) GPL exceptions are _probably_ OSD-conformant. There's no formal sense in which OSI has approved anything other than the licenses it has approved through its license approval process.

Also, I don't think you can make this general assumption going forward, because in a recent mailing list thread @jlovejoy has [clarified](https://lists.spdx.org/g/Spdx-legal/message/3283) (unless I misunderstood) that SPDX is not, or no longer, strictly tied to an understanding that "exceptions" (the right side of a WITH expression) must be limited to additional permissions. This came out of some interest in exploring SPDX recognition of a currently-hypothetical exception type that would involve imposition of clearly-non-FOSS restrictions on top of FOSS license terms. In the past, SPDX has [rejected](https://github.com/spdx/license-list-XML/issues/902) inclusion of an exception identifier for the Commons Clause but I believe this was prior to SPDX's adoption of its current license incluson criteria.

DOCS/OSI-SPDX-history.md
```diff
@@ -0,0 +1,104 @@
...
+
+Note that OSI has the GPL-2.0-or-later license notice on its page for the eCos exception, which would use the SPDX license expression: `GPL-2.0-or-later WITH eCos-exception-2.0`. Did OSI _only_ approve this exception for GPL-2.0-or-later? 
+
+Suggestion: Update the [eCos](https://opensource.org/licenses/eCos-2.0) page to use the current `eCos-exception-2.0` identifier and change the URL to https://opensource.org/licenses/eCos-exception-2.0. Clarify what versions of GPL it is approved for.
+
+## Licenses that allow for variations by way of a different license notice
+These are licesnes that were approved by the OSI in full. These licenses allow legally substantive variations based on use of a different license notice. This variataion warranted distinct SPDX identifiers to indicate whether or not the allowed variation is being triggered. All of these variations include a Note in the SPDX License List entry explaining this.
```
xsuchy, line 49: s/licesnes/licenses/

xsuchy:
@jlovejoy ping, It will be nice to finish this one PR before itself become a history :)

goneall:
For reference see: https://github.com/OpenSourceOrg/licenses/issues/62

I created a utility to compare the SPDX license data with the OSI machine readable data.

Once all the issues are resolved, we can automate the OSI field in the license list publisher.

DOCS/OSI-SPDX-history.md
jlovejoy: fixed

jlovejoy:
@goneall - not sure that will resolve all of these old issues? Let's discuss so I better understand!

DOCS/OSI-SPDX-history.md
jlovejoy: seems like there is no longer a question here, so removing this from the "list"

DOCS/OSI-SPDX-history.md
jlovejoy: added note

jlovejoy:
made updates, but tests still failing here, so I'm made a new PR at #2504 and am going to close this one.

Foxsenrubas (approved):

Issue #1706, opened 2022-11-11
New license request: LOOP [SPDX-Online-Tools]


jamesjer:
**1.** License Name: Common Lisp LOOP License
**2.** Short identifier: LOOP
**3.** License Author or steward: Unknown
**4.** Comments: The license does not exactly match any license currently on the list.  The license text has been stable for many years.  The license is known to be in use by various Common Lisp implementations.

In each known use, the license text is repeated twice, once referring to MIT and once referring to Symbolics.  The license text is similar, but not identical, to the NTP license.

There is no known license name or short identifier.  I will propose both, but am happy to have somebody choose better names.
**5.** Standard License Header: 
**6.** License Request Url: http://tools.spdx.org/app/license_requests/183
**7.** URL(s): https://gitlab.com/embeddable-common-lisp/ecl/-/blob/develop/src/lsp/loop.lsp
**8.** OSI Status: Unknown
**9.** Example Projects: https://gitlab.com/embeddable-common-lisp/ecl/-/blob/develop/src/lsp/loop.lsp, http://git.savannah.gnu.org/cgit/gcl.git/tree/gcl/lsp/gcl_loop.lsp?h=Version_2_6_13pre, https://sourceforge.net/p/sbcl/sbcl/ci/master/tree/src/code/loop.lisp, https://github.com/cl-adams/adams/blob/master/LICENSE.md, https://github.com/blakemcbride/eclipse-lisp/blob/master/lisp/loop.lisp, https://gitlab.common-lisp.net/cmucl/cmucl/-/blob/master/src/code/loop.lisp

richardfontana:
This arises out of a Fedora packaging issue (@jamesjer is a Fedora package maintainer), and we would request that this be treated as a 'high Fedora priority' issue.

jlovejoy:
note, already determined to be "allowed" for Fedora, as per https://gitlab.com/fedora/legal/fedora-license-data/-/issues/93#note_1169622386

jlovejoy:
# New submission review

* **Has the license been approved by the OSI?** _All OSI-approved license will be included on the SPDX License List, regardless of other factors_
    - [ ] Yes
    - [ ] No

## Definitive Factors

_These must all be satisfied to allow inclusion in the license list_

1. **Is the submitted license unique, that is, it does not match another license already on the License List as per the [matching guidelines](https://spdx.org/spdx-license-list/matching-guidelines)?**
    - [ ] Yes
    - [ ] No
[...]

PR #1740, opened 2022-12-12
New License: LOOP

jamesjer:
fixes https://github.com/spdx/license-list-XML/issues/1706

### DIFF
--- a/src/LOOP.xml
+++ b/src/LOOP.xml
@@ -1,0 +1,68 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<SPDXLicenseCollection xmlns="http://www.spdx.org/license">
+  <license isOsiApproved="false" licenseId="LOOP"
+  name="Common Lisp LOOP License" listVersionAdded="3.20">
+    <crossRefs>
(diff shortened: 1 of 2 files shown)


swinslow (approved):
Thanks @jamesjer! Your PR passed the validation check.

For the one that failed, I don't think that's because of your PR. I think there's some issue that's popped up in a dependency for one of the CI checks (I see it's failing also in other recent PRs). But that shouldn't be due to a problem with your submission.

@goneall I'll take a look also, but if you happen to see what's going on at https://github.com/spdx/license-list-XML/actions/runs/3680847546/jobs/6243671789, feel free to jump in  :)  Either way I'm assuming that since this PR passed the "validate canonical match" check, it must also be okay for the "validate schema" check.

goneall:
@swinslow I took a look at the failure.  It looks like it is an incompatibility involving in indirect dependency of [libxml-xsd](https://www.npmjs.com/package/libxml-xsd) and Python.  

I did some searches and couldn't find any information.  It looks like we are using the latest version of libxml-xsd.  The only thing I can think of is trying to install a different version of Python or use a different XML XSD library.

swinslow:
Thanks @goneall! I'm taking a closer look at this and have filed #1741 so we can dig deeper.

@jamesjer I'm going ahead and merging this PR, thanks again for your help here!

PR #1742, opened 2022-12-14
Fix validation with switch to Python xml validator

### LINKED ISSUE #1741: validate-schema check is failing
See discussion in https://github.com/spdx/license-list-XML/pull/1740#pullrequestreview-1216297591 and thereafter

Example of failing message: https://github.com/spdx/license-list-XML/actions/runs/3673080831/jobs/6209823179

swinslow:
Fixes #1741

This replaces the now-breaking node-based XML validation script with a Python script that uses the xmlschema PyPI dependency.


swinslow:
@goneall @zvr It would be great to get your eyes on this just as a sanity check  :)

This PR removes the former (now-breaking) JavaScript-based schema validator, and replaces it with a very short Python script to do the validation.

Once this is merged, I can submit an intentionally-breaking PR to show it failing. You can also see it at https://github.com/swinslow/xml-validation/pull/3 and https://github.com/swinslow/xml-validation/actions/runs/3696312431/jobs/6259843418 which shows what happens when an invalid license is submitted.

Finally, this also updates the GitHub actions to run the Python script in place of JS / Node, and removes package.json since I don't think it is required for anything other than the removed validate-schema.js file.

goneall (commented):
This looks good - just one likely naive question: when there is a validation error and Python raises an exception, will this set the status exit code properly so that Github action will fail?

goneall (approved):
Thanks @swinslow!

validate-schema.py
```diff
@@ -0,0 +1,25 @@
...
+                paths.append(os.path.join(root, f))
+    return paths
+
+if __name__ == "__main__":
+    spdxSchema = xmlschema.XMLSchema("schema/ListedLicense.xsd")
+    paths = getAllXMLFiles("src")
+
+    print(f"Checking {len(paths)} licenses and exceptions for schema validation...")
+    for xf in paths:
+        spdxSchema.validate(xf)
```
zvr, line 22: Alternatively, call `is_valid`, so that you can check all files without raising an exception and stop checking the rest.

If it returns false, add it to a list of "bad" ones. At the end of the loop if the bad list is not empty, print them and raise an exception.

swinslow: Good call -- I'll give that a try, will update...

swinslow:
@zvr With 2debd65, when it encounters errors it now gives a list of the reasons for failures. Here's an example of what the output looks like when a couple of invalid entries are thrown in:

```
(xmlvalidation) steve@terciel: license-list-XML> python3 validate-schema.py
Checking 549 licenses and exceptions for schema validation...
Found 2 licenses and exceptions that failed schema check:
  - src/Invalid.xml
  - src/exceptions/Invalid-exception.xml
---------------------------
Displaying reasons for failures:
  - src/Invalid.xml: Unexpected child with tag 'tns:hi' at position 1. Tag ('tns:license' | 'tns:exception') expected.
  - src/exceptions/Invalid-exception.xml: mismatched tag: line 95, column 2
```

swinslow:
@zvr Also, since it exits with a failing return code, I _believe_ that should cause the build to fail if any invalid entries are detected (even though we're now catching the exceptions). But after this is merged, I'll submit a test PR so we can confirm.

swinslow:
Given the +1 from @goneall and I think I've addressed @zvr's suggestion, I'm going to go ahead and merge this. Will submit a failing PR so that you can see that it works also.  :)

PR #1801, opened 2023-02-04
Update references now that the spec has been reorganized

Jayman2000:
The language that this commit changes was first introduced by 6a12c74 (add missing URLs, 2019-05-19). At that time, the latest version of the SPDX spec was 2.1. That version of the spec had sections which use Arabic numerals and appendices which use Roman numerals. At the moment, the latest version of the SPDX spec is 2.3.0. This version of the spec has clauses which use Arabic numerals and annexes which use letters.

This change updates references to account for how the latest version of the spec is organized.


jlovejoy:
ugh, thought I caught all of these... thanks!

PR #1807, opened 2023-02-06
New License: TPL-1.0

### LINKED ISSUE #1612: New license request: Thor Public License (TPL-1.0)
License text: https://fedoraproject.org/wiki/Licensing:ThorPublicLicense

License steward: Thomas Richter

Suggested short identifier: TPL (used by Fedora under the pre-SPDX system, and also apparently favored by the license author)

This license appears in one Fedora package (atari++, upstream project: http://www.xl-project.com/ [apparently no web-visible source repo]). It is an MPL 1.1 derivative differing substantively from MPL 1.1 only in the choice of law provision.

jlovejoy:
fixes #1612 

### DIFF
--- a/src/TPL-1.0.xml
+++ b/src/TPL-1.0.xml
@@ -586,7 +586,7 @@
                   any software, hardware, or device, other than such
                   Participant's Contributor Version, directly or indirectly
                   infringes any patent, then any rights granted to You by such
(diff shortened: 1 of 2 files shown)


swinslow:
I think e3709f0 should fix the error in the tests -- will merge after it runs if so!

swinslow (approved):

PR #1809, opened 2023-02-07
Remove incorrect bullet

Jayman2000:
Before this change, the `DOCS/faq.md` said that the License List incorporated the syntax for license expressions. That’s not accurate. [The syntax for license expressions is in the SPDX Spec][1]. Also, the link in that bullet point was invalid.

[1]: <https://spdx.github.io/spdx-spec/v2.3/SPDX-license-expressions/>

### DIFF
--- a/DOCS/faq.md
+++ b/DOCS/faq.md
@@ -18,8 +18,6 @@ regarding what is eligible for inclusion on the SPDX License List
 provide parameters as to what constitutes a match to a license or exception on the SPDX License List
 * an [explanation of the fields](https://github.com/spdx/license-list-XML/blob/main/DOCS/license-fields.md) 
 used in the SPDX License List
-* [license expression syntax](LINK) 
-enable expressing composite licensing scenarios, such as when more than one license applies, there is a choice of license, or an exception or additional terms apply to the license
 
 The authoritative files for the SPDX License List are stored in XML source files 
 in a [Github repo](https://github.com/spdx/license-list-XML). These files are 


swinslow:
The license expression syntax is relevant to the use of the SPDX License List. I think it's worth including it in the FAQ (though agree with your note about the link being broken -- thanks for calling that out). But other than fixing the link, I would leave this bullet point in.

Jayman2000:
> The license expression syntax is relevant to the use of the SPDX License List.

I agree that it’s relevant to the License List, but that’s not what [`DOCS/faq.md`](https://github.com/spdx/license-list-XML/blob/cc49f38054ec5276472bb06dd1280d9ec700c6e4/DOCS/faq.md) says. It says

> # What is the SPDX License List?
> […]
> It is more than just a list, though. The SPDX License List incorporates
> […]
> * [license expression syntax](LINK) 
> enable expressing composite licensing scenarios, such as when more than one license applies, there is a choice of license, or an exception or additional terms apply to the license

In other words, this part of the FAQ says that the license expression syntax is a part of the License List, but the license expression syntax isn’t a part of the License List.

> I think it's worth including it in the FAQ

I agree. Even with this bullet removed, there’s still [two](https://github.com/spdx/license-list-XML/blob/cc49f38054ec5276472bb06dd1280d9ec700c6e4/DOCS/faq.md#why-are-exceptions-listed-separately) [different](https://github.com/spdx/license-list-XML/blob/cc49f38054ec5276472bb06dd1280d9ec700c6e4/DOCS/faq.md#what-are-license-expressions) questions that link to the Spec’s license expression annex.

> (though agree with your note about the link being broken -- thanks for calling that out). But other than fixing the link, I would leave this bullet point in.

That’s what I was going to do originally, but I couldn’t find anything on the License List or in this repo that really described the syntax for license expressions. I was able to find two explanations of license expressions:

1. On the License List website in the Deprecated License Identifiers section.
2. [In the FAQ](https://github.com/spdx/license-list-XML/blob/cc49f38054ec5276472bb06dd1280d9ec700c6e4/DOCS/faq.md#what-are-license-expressions).

but those aren’t really descriptions of the syntax. Those are just overviews.

jlovejoy:
@Jayman2000 - as the person who probably wrote this... I think the operative wording that leads this list is, "It is more than just a list, though. The SPDX License List incorporates"

This is based on my observation that many people seem to view the webpage at https://spdx.org/licenses/ - and just look at the list (literally) there, skipping over the other information at the top of the page that explains context, bigger picture, etc. Hence, "it is more than just a list" - the list is what people primarily see, but what makes the list meaningful is more than what you see on the list... 

not sure that makes sense, but if there is a better way to express that, I'm all ears!

Jayman2000:
> not sure that makes sense, but if there is a better way to express that, I'm all ears!

I just opened #1826 which tries to express that idea in a better way.

jlovejoy:
given the other PR, I'm going to close this one.

@Jayman2000 - have you joined the mailing list? would be nice to know how you found SPDX/got interested!

Jayman2000:
> @Jayman2000 - have you joined the mailing list?

I hadn’t joined before you said that, but I have now.

> would be nice to know how you found SPDX/got interested!

I’m not sure that it makes sense to post my story on the mailing list, but I will post it here.

I develop video games as a hobby. I often end up using assets from [OpenGameArt.org](https://opengameart.org/) with comes with a few different challenges:

1. It can be difficult to make sure that games include all of the required attribution and legal notices.
2. It can be difficult to communicate which files are under which licenses.

I ended up finding [REUSE](https://reuse.software/) which solved those two problems. From there, I found and got interested in SPDX.

PR #1811, opened 2023-02-10
FAQ heading improvements

Jayman2000:
See the individual commit messages for details.

### DIFF
--- a/DOCS/faq.md
+++ b/DOCS/faq.md
@@ -3,7 +3,7 @@ Tip: use the table of contents icon in the upper left hand corner of this window
 The SPDX License List
 =====================
 
-# What is the SPDX License List?
+## What is the SPDX License List?
 The SPDX License List is an integral part of the SPDX Specification. The
 SPDX License List itself is a list of commonly found licenses and
 exceptions used in free and open or collaborative software, data,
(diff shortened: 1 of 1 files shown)


jlovejoy:
interesting @Jayman2000 - we are not (currently) generating webpages from these .md pages, but... not sure if we might in the future, in which case this would be helpful

@seabass-labrax - thoughts on this?

Jayman2000:
> we are not (currently) generating webpages from these .md pages

GitHub automatically turns Markdown into Web pages for you. If you go [here](https://github.com/spdx/license-list-XML/blob/main/DOCS/faq.md) and inspect the page, you’ll be able to see that the “The SPDX License List” heading is an &lt;h1&gt; and that the “What is the SPDX License List?” heading is also an &lt;h1&gt;.

seabass-labrax (approved):

seabass-labrax:
Thanks, @Jayman2000! This looks good to me.

For those reading this who might not be aware, a line of equals signs underneath text actually makes that text an 'h1' heading, so this would be two 'h1' headings:

```
> The SPDX License List
> =====================
> 
> # What is the SPDX License List?
```

I would suggest changing that first heading to be written with the more typical hash mark method to prevent future confusion, either in this PR or a follow-up one.

Jayman2000:
> I would suggest changing that first heading to be written with the more typical hash mark method to prevent future confusion, either in this PR or a follow-up one.

OK, I just pushed a new commit to this PR that does that.

seabass-labrax:
Brilliant; thank you!

seabass-labrax (approved):

PR #1826, opened 2023-02-12
More directly draw attention to rest of list

Jayman2000:
[Before this change, there was an awkward and slightly incorrect part of the FAQ.][1] [It was trying to draw attention to specific parts of the License List Web page that typically get ignored.][2]

This replaces that part of the FAQ with a more direct call out. Hopefully, this will make people pay more attention to those parts of the License List.

This change also removes an invalid link.

[1]: <https://github.com/spdx/license-list-XML/pull/1809#issue-1574869665>
[2]: <https://github.com/spdx/license-list-XML/pull/1809#issuecomment-1426582567>

---

This PR depends on another PR. Before this one is merged, the following needs to happen:

1. [x] spdx/LicenseListPublisher#154 needs to be finished.
2. [x] spdx/LicenseListPublisher#154 needs to be merged.
3. [ ] A new version of [the LicenseListPublisher](https://github.com/spdx/LicenseListPublisher) needs to be released.
4. [ ] A commit needs to get added to license-list-XML that makes it use the new version of the LicenseListPublisher.
5. [ ] A new version of the License List needs to be released and published at <https://spdx.org/licenses/>.

Until those things happen, the links that this PR adds won’t work properly.

---

This PR is an alternative to #1809.

### DIFF
--- a/DOCS/faq.md
+++ b/DOCS/faq.md
@@ -11,15 +11,10 @@ hardware, or documentation. The SPDX License List includes a
 standardized short identifier, the full name, the license text, and a
 canonical permanent URL for each license and exception.
 
-It is more than just a list, though. The SPDX License List incorporates
-* [license inclusion principles](license-inclusion-principles.md)
-regarding what is eligible for inclusion on the SPDX License List
-* [matching guidelines](https://spdx.github.io/spdx-spec/v2.3/license-matching-guidelines-and-templates/) 
-provide parameters as to what constitutes a match to a license or exception on the SPDX License List
-* an [explanation of the fields](license-fields.md)
-used in the SPDX License List
-* [license expression syntax](LINK) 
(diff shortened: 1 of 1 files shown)


jlovejoy:
hey @Jayman2000 - thanks for this again... but I still want to keep the list of bullets... we really can't draw enough attention to the larger context in every place possible. Of course, that does mean having to update links in various places, but I think that's better than people skipping this stuff (which they do anyway...)

As to the version number on the main SPDX License List page - when the next release is pushed, that gets updated as part of that. so no added "to-do" for that.

:)

goneall:
Just a note - https://github.com/spdx/LicenseListPublisher/pull/154 is already merged.  It doesn't look like the new tags will be used, but I'll leave those tags in unless someone wants to create a PR to remove.

Also - I'm having trouble publishing a new version of the app due to some issues with the OSI website - see https://github.com/spdx/LicenseListPublisher/pull/155 for more info

PR #1850, opened 2023-02-22
New exception: Autoconf-exception-macro

### LINKED ISSUE #1849: New exception: Autoconf macro exception
<https://github.com/freedesktop/xorg-macros/blob/39f07f7db58ebbf3dcb64a2bf9098ed5cf3d1223/xorg-macros.m4.in#L1260>, also found in <https://www.x.org/releases/individual/app/xgamma-1.0.7.tar.xz>, uses an exception from GPL-3.0-or-later:

~~~~
As a special exception, the respective Autoconf Macro's copyright owner
gives unlimited permission to copy, distribute and modify the configure
scripts that are the output of Autoconf when processing the Macro. You
need not follow the terms of the GNU General Public License when using
or distributing such scripts, even though portions of the text of the
Macro appear in them. The GNU General Public License (GPL) does govern
all other use of the material that constitutes the Autoconf Macro.

This special exception to the GPL applies to versions of the Autoconf
Macro released by the Autoconf Archive. When you make and distribute a
modified version of the Autoconf Macro, you may extend this special
exception to the GPL to apply to your modified version as well.
~~~~

This exception looks distinct from <https://spdx.org/licenses/Autoconf-exception-2.0.html>.

It has already been approved for Fedora <https://gitlab.com/fedora/legal/fedora-license-data/-/issues/161>.

I need to add this exception to SPDX exception list and to determine an identifier. I propose "Autoconf macro exception" name.

ppisar:
This adds a new Autoconf-exception-macro exception.

Fixes #1849

### DIFF
--- a/src/exceptions/Autoconf-exception-macro.xml
+++ b/src/exceptions/Autoconf-exception-macro.xml
@@ -1,0 +1,27 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<SPDXLicenseCollection xmlns="http://www.spdx.org/license">
+   <exception licenseId="Autoconf-exception-macro" name="Autoconf macro exception" listVersionAdded="3.21">
+       <crossRefs>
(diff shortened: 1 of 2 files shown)


swinslow:
LGTM, thanks @ppisar!

PR #1862, opened 2023-03-06
Removed the `isDeprecated` field form the licenses

### LINKED ISSUE #1545: Clean up isDeprecated values
(currently marking for 3.19 docs/cleanup release, but might be more appropriate for 3.20 instead)

This is a follow-on from the discussion in #583. I believe @goneall has reworked the license list publisher tooling to just look for the presence of `deprecatedVersion="..."` to determine whether or not a license is deprecated. Which means that the separate `isDeprecated=true` flag is (ironically) deprecated.

Two tasks to do in parallel:
* remove `isDeprecated` from the [XML schema description file](https://github.com/spdx/license-list-XML/blob/master/schema/ListedLicense.xsd); and
* go through the deprecated licenses and exceptions to make sure that they have a `deprecatedVersion="..."` value and to remove the `isDeprecated=true` flag.

BassCoder2808:
Fixes #1545 
Compared all the licenses in the `src` folder and `src/exceptions` folder and removed the `isDeprecated` field from the XML format as each license had `deprecatedVersion` field

### DIFF
--- a/src/AGPL-1.0.xml
+++ b/src/AGPL-1.0.xml
@@ -2,7 +2,7 @@
 <SPDXLicenseCollection xmlns="http://www.spdx.org/license">
    <license isOsiApproved="false" licenseId="AGPL-1.0"
             name="Affero General Public License v1.0"
-            isDeprecated="true" deprecatedVersion="3.1">
+            deprecatedVersion="3.1">
       <obsoletedBys>
          <obsoletedBy>AGPL-1.0-only</obsoletedBy>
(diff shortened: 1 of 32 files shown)


swinslow:
Looks good to me, thank you @BassCoder2808! Merging now.

swinslow (approved):

goneall:
Just FYI - I checked and indeed the deprecated fields are being handled properly in the license list data JSON files after this PR.  It all looks good.

swinslow:
Thank you @goneall! I had left this browser tab open with the intent to go back and check later -- thanks for beating me to it  ;-)

PR #1864, opened 2023-03-08
Clean up isDeprecated values

### LINKED ISSUE #1545: Clean up isDeprecated values

BassCoder2808:
Fixes #1545 
Removed the `isDeprecated` attribute from the `XML Schema Definition` file. `xml-fields.md` contains `isDeprecated` definition which was also removed now.

### DIFF
--- a/DOCS/xml-fields.md
+++ b/DOCS/xml-fields.md
@@ -44,19 +44,16 @@ Within the .xml file, the first and last lines of the file should be the followi
 The main tag used to define the license or exception is, unsurprisingly, **`<license>`** or **`<exception>`**. All of the remaining content will be enclosed within a `<license></license>` or `<exception></exception>` pair of tags.
 
 There are two mandatory attributes for every `<license>` and `<exception>` tag:
-
-- `licenseId`: the unique SPDX Identifier for the license; should be identical to the filename
-- `name`: the longer "real name" of the license
+* `licenseId`: the unique SPDX Identifier for the license; should be identical to the filename
+* `name`: the longer "real name" of the license
 
 There are two additional attributes which are optional, but highly encouraged:
-
-- `isOsiApproved` (for licenses, not exceptions): either "true" or "false" based on whether this license has been approved by the [Open Source Initiative](https://opensource.org/licenses/alphabetical)
-- `listVersionAdded`: in which release version of the SPDX License List was the license first added, e.g. "3.19".
-  - Typically you'll check the currently-released version at https://spdx.org/licenses/ and increment the minor version by 1 for a new license.
+* `isOsiApproved` (for licenses, not exceptions): either "true" or "false" based on whether this license has been approved by the [Open Source Initiative](https://opensource.org/licenses/alphabetical)
+* `listVersionAdded`: in which release version of the SPDX License List was the license first added, e.g. "3.19".
+  * Typically you'll check the currently-released version at https://spdx.org/licenses/ and increment the minor version by 1 for a new license.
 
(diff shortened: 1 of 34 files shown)


DOCS/xml-fields.md
```diff
@@ -44,17 +44,19 @@ Within the .xml file, the first and last lines of the file should be the followi
 The main tag used to define the license or exception is, unsurprisingly, **`<license>`** or **`<exception>`**. All of the remaining content will be enclosed within a `<license></license>` or `<exception></exception>` pair of tags.
 
 There are two mandatory attributes for every `<license>` and `<exception>` tag:
-* `licenseId`: the unique SPDX Identifier for the license; should be identical to the filename
```
swinslow, line 47: Throughout this PR, it looks like in addition to removing the `isDeprecated` documentation, this PR is also changing all of the bullet types from `*` to `-`.

Are these changes needed, since [both bullet types are valid for Markdown lists on GitHub](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#lists)? 

If not needed, could you please revise the PR so that the only change is removing the `isDeprecated` documentation? Would prefer to keep the edits limited so that it's clear what's being changed here.

DOCS/xml-fields.md
```diff
@@ -44,17 +44,19 @@ Within the .xml file, the first and last lines of the file should be the followi
...
 There are two additional attributes which are optional, but highly encouraged:
-* `isOsiApproved` (for licenses, not exceptions): either "true" or "false" based on whether this license has been approved by the [Open Source Initiative](https://opensource.org/licenses/alphabetical)
-* `listVersionAdded`: in which release version of the SPDX License List was the license first added, e.g. "3.19".
-  * Typically you'll check the currently-released version at https://spdx.org/licenses/ and increment the minor version by 1 for a new license.
+
+- `isOsiApproved` (for licenses, not exceptions): either "true" or "false" based on whether this license has been approved by the [Open Source Initiative](https://opensource.org/licenses/alphabetical)
+- `listVersionAdded`: in which release version of the SPDX License List was the license first added, e.g. "3.19".
+  - Typically you'll check the currently-released version at https://spdx.org/licenses/ and increment the minor version by 1 for a new license.
 
 Finally, if the license ID has been deprecated, two additional attributes should be included:
```
swinslow, line 57: Note that this line should also be changed to reference "one additional attribute" instead of "two additional attributes"

swinslow (changes requested):
Thanks @BassCoder2808! I have a couple of comments, please take a look.

@goneall Feel free to weigh in if you would like. Otherwise, this looks good to me once @BassCoder2808 makes the requested changes.

goneall (approved):
The schema changes look good to me.

Thanks @BassCoder2808

swinslow (approved):
LGTM, thanks again @BassCoder2808!

PR #1867, opened 2023-03-09
Clarify language in issue template

swinslow:


swinslow:
@jlovejoy -- I _think_ this should improve the language for the issue template, in line with our discussion on the Legal Team call just now. Take a look and see what you think, and if it looks good then we can merge it and see how it renders in the actual template itself.

jlovejoy:
@swinslow  - looks like a good start in any case. I'm not entirely clear where the different sets of text shows up, so let's merge and iterate, if needed!

PR #1872, opened 2023-03-15
OpenPBS-2.3: fix p tags

jlovejoy:
move <p> tags to before <bullet> tag to improve formatting so that numbering isn't on line by itself, see https://spdx.org/licenses/OpenPBS-2.3.html. This may have formatted this way due to the online tool XML generation function, but before I make an issue there, I need to make sure I changed it in the right way?

### DIFF
--- a/src/OpenPBS-2.3.xml
+++ b/src/OpenPBS-2.3.xml
@@ -11,8 +11,8 @@
             OpenPBS (Portable Batch System) v2.3 Software License
          </titleText>
          <copyrightText>
-            Copyright (c) 1999-2000 Veridian Information
-            Solutions, Inc. All rights reserved.
+            <p>Copyright (c) 1999-2000 Veridian Information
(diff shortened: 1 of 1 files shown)


jlovejoy:
@goneall - did I fix this the right way given the way it's currently displaying? Or should the <p> tag just be removed?

(once you answer, I'll test in XML editor in online tools and add issue, if need be)

jlovejoy:
https://spdx.org/licenses/TPL-1.0.html is doing the same thing, so will fix that as well

goneall:
@jlovejoy Short answer is - I think your fix should work.  Longer answer - the `copyright`, `title`, and `bullet` tags should not not do any formatting, so you add whatever paragraph and line break formatting tags around the text as if there were not there.

swinslow:
good catch -- LGTM, thanks @jlovejoy! Merging now.

PR #1873, opened 2023-03-15
HPND-sell-variant-MIT-disclaimer - add line breaks

jlovejoy:
displaying weird where line breaks are needed, see https://spdx.org/licenses/HPND-sell-variant-MIT-disclaimer.html

still need extra line btw copyright notice and start of license? would that mean a <p> tag around copyright notice??

@swinslow  thoughts on this one?


swinslow (approved):
LGTM, thanks!

PR #1899, opened 2023-03-27
[TEST] TEST1-Apache

HirumalPriyashan:


jlovejoy:
@HirumalPriyashan - did you follow the documentation here https://github.com/spdx/license-list-XML/blob/main/DOCS/request-new-license.md#accepted-license-process or submit directly via the online tool with no manual editing?

HirumalPriyashan:
@jlovejoy I have directly submitted through online tool with no manual editing.

jlovejoy:
note, as per the documentation linked to above, some manual editing is required and the creation of the .txt file - that is part of understanding how the tools works :)

HirumalPriyashan:
@jlovejoy I have manually edited the XML file, there was a lot of formatting needed because validations are failing because of `<p>` tags were not inside the `item` tags in list. Also had to manually create the list in section 4 because online tool has been failed to capture that as a list

jlovejoy:
> @jlovejoy I have manually edited the XML file, there was a lot of formatting needed because validations are failing because of `<p>` tags were not inside the `item` tags in list.

Yes... now you are feeling some of the pains that the SPDX-legal team experiences! This is good, as it will help you think of ways the tools might be improved!

Interesting note on the <p> tag needing to be inside the item tag... is there a way to improve the XML generation on this?
@swinslow - looking at https://github.com/spdx/license-list-XML/blob/main/DOCS/xml-fields.md, maybe something to clarify there?

> Also had to manually create the list in section 4 because online tool has been failed to capture that as a list
yes, that was sort of what I was trying to "recreate" as I've run into this before and it's a lot of work. I think the way the text was copied in this submission v. in the submission for #1905 may hold the key. This may not be fixed by the XML editor/generation, but may be a matter of providing more specific instructions for submitters on how or where they copy text from?

jlovejoy:
looks like the tests passes, so well done. I'll leave this open for a bit as a reference

PR #1901, opened 2023-03-27
Non printing characters

jamin-aws-ospo:
Removing non breaking spaces and line feeds from text versions of licenses.


zvr:
:-1: 

First of all, this PR only changes the test texts, not the license texts.

But I'm not in favor of this blind change to all texts. When you want to match some license text that includes these non-breaking spaces (as it should, in some cases), you would need to implement the matching anyway.

MarkAtwood:
I disagree about "When you want to match some license text that includes these non-breaking spaces".  If a license text can conceivably be rendered into ASCII, it should never contain a non-printing-character.  We have been avoiding this problem until now because of accidental mis-implementations of regex  in Ruby as vs Python.

jamin-aws-ospo:
> First of all, this PR only changes the test texts, not the license texts.

It was the test texts I was instructed to change.  If you'd like other texts changed instead, I'm happy to do so.

goneall:
> First of all, this PR only changes the test texts, not the license texts.

The LicenseListPublisher actually copies the test text to the data repository text directory, so those would be the text to change if you want to alter the text directory.  Note that the template is generated from the License XML file, so this change won't impact those files.

zvr:
@MarkAtwood 5 out of the 6 of the license texts that contain non-breaking spaces can *not* be rendered in (7-bit) ASCII, as they are in French and have accented letters, guillemets, etc.

When one will encounter this license text elsewhere, in all likelihood it will still contain all these characters, including the non-breaking spaces. Having the SPDX version a little different (having replaced only the nbspace characters with ASCII spaces) is not very useful, in my view.

jamin-aws-ospo:
@zvr there are more than 6 licenses that have non-breaking spaces.

I'm happy to revert the changes for the files that _should_ have the non-breaking space, such as the French texts.

MarkAtwood:
@jamin-aws-ospo  please revert the changes to the FR language license texts, keep the changes in the EN texts.  And then we can nudge this to merge.

jlovejoy:
@zvr - what are you thoughts on this as updated?

PR #1909, opened 2023-04-03
Accept any copyright holders in BSD-4.3TAHOE

ppisar:
indent software has this license with more copyright holders <https://git.savannah.gnu.org/cgit/indent.git/tree/doc/indent.texi?id=a74c6b4ee49397cf330b333da1042bffa60ed14f#n1788> that originally listed. Augmenting the license markup was proposed in Fedora license review <https://gitlab.com/fedora/legal/fedora-license-data/-/issues/174>.


bsdimp:
I like this, but can't say if the xml change is perfect. It looks reasonable though

jlovejoy:
Thanks @ppisar - I think the markup for replaceable text like this is `match+".+"` as per https://github.com/spdx/license-list-XML/blob/main/DOCS/xml-fields.md#replaceable-text

But will defer to @swinslow or @goneall to confirm...

PR #1920, opened 2023-04-09
fix: add missing period

reggiedotre:


swinslow (approved):
Good catch, this is missing from the `standardLicenseHeader` field in AGPL-3.0-only and should be added.

Present in AGPL-3.0-or-later but also missing from AGPL-3.0 (the deprecated listing); I'll submit a follow-up PR to add it there.

PR #1923, opened 2023-04-10
AGPL-1.0: fix typo in Notes field

### LINKED ISSUE #1922: Deprecation notice for AGPL-1.0 doesn't make any sense
The line https://github.com/spdx/license-list-XML/blob/main/src/AGPL-1.0.xml#L13
states:
`      <notes><p>DEPRECATED: Use the license identifier AGPL-1.0-only instead of APL-1.0, and use AGPL-1.0-or-later instead of AGPL-1.0.</p></notes>`

The `APL-1.0` is presumably a typo (APL-1.0 is the unrelated Adaptive Public License). But if it's a typo for `AGPL-1.0`, the corrected statement
`DEPRECATED: Use the license identifier AGPL-1.0-only instead of AGPL-1.0, and use AGPL-1.0-or-later instead of AGPL-1.0.` still seems to not make any sense, because it is not specifying whether to use AGPL-1.0-only or AGPL-1.0-or-later instead of AGPL-1.0. 

I am not sure what was actually intended here. But I notice that there is no counterpart deprecation note for GPL-2.0 and GPL-3.0. Why should AGPL-1.0 need such a deprecation note if there isn't one for the closest textual counterpart (GPL-2.0)?

Furthermore, though perhaps this is a separate or larger issue, to the extent this was an outgrowth of some earlier effort to treat AGPL-1.0 similarly to the legacy GPL-family identifiers, AGPL-1.0 was not an FSF-stewarded license. The [commit message](https://github.com/spdx/license-list-XML/commit/96f8f6489f4bded6df4a8ca05a641e11dddeb118) that introduced this `DEPRECATED` note says:

`On Thu, Dec 28, 2017 at 09:56:54AM -0800, Jilayne Lovejoy wrote [1]:
> I am conferring with the FSF as to AGPL-1.0 - I will make the
> appropriate changes as needed when I get feedback there. No further
> discussion needed at this point.

On Thu, Jan 11, 2018 at 09:56:18PM +0000, Jilayne Lovejoy wrote [2]:
> ... I have confirmed that we should treat AGPL-1.0 that same) as we
> suspected.
`

Why would the opinion of the FSF have mattered at all here, given that AGPL-1.0 was not an FSF license?

jlovejoy:
fixes #1922 

### DIFF
--- a/src/AGPL-1.0.xml
+++ b/src/AGPL-1.0.xml
@@ -10,7 +10,7 @@
       <crossRefs>
          <crossRef>http://www.affero.org/oagpl.html</crossRef>
       </crossRefs>
-      <notes><p>DEPRECATED: Use the license identifier AGPL-1.0-only instead of APL-1.0, and use AGPL-1.0-or-later instead of AGPL-1.0.</p></notes>
+      <notes><p>DEPRECATED: Use the license identifier AGPL-1.0-only instead of AGPL-1.0, and use AGPL-1.0-or-later instead of AGPL-1.0.</p></notes>
     <text>
       <titleText>
          <p>AFFERO GENERAL PUBLIC LICENSE
--- a/src/AGPL-1.0.xml
+++ b/src/AGPL-1.0.xml
@@ -10,7 +10,7 @@
       <crossRefs>
          <crossRef>http://www.affero.org/oagpl.html</crossRef>
       </crossRefs>
-      <notes><p>DEPRECATED: Use the license identifier AGPL-1.0-only instead of AGPL-1.0, and use AGPL-1.0-or-later instead of AGPL-1.0.</p></notes>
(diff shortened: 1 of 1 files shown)


richardfontana:
@jlovejoy the fix still doesn't make any sense - maybe you mean to say `<notes><p>DEPRECATED: Use the license identifier AGPL-1.0-only instead of AGPL-1.0, and use AGPL-1.0-or-later instead of AGPL-1.0+</p></notes>`
but if so, why isn't there a corresponding note for GPL-1.0, GPL-2.0, etc.?

jlovejoy:
re adding notes for GPL-1.0, etc. I answered that here: https://github.com/spdx/license-list-XML/issues/1922#issuecomment-1502053335

I was going to add a PR to add those here as well, but didn't get to it and honestly, it's not really the most pressing issue right now ;)

jlovejoy:
@swinslow - can you review and merge?

swinslow (approved):

PR #1929, opened 2023-04-13
Python-2.0.1 is FSF approved but not OSI approved

akostadinov:
License was added with #1549 as a follow up on #1200

renaming license to match FSF identifier [1][2]

<del>In #1200 it is stated that the license is FSF approved but no OSI
approved or at least not with the right name. With a link to the FSF
website [1].</del>

<del>So I'm adding this attribute to the license for it to be properly tagged
as FSF approved.</del>

[1] https://www.gnu.org/licenses/license-list.en.html#Python
[2] https://wking.github.io/fsf-api/licenses-full.json


akostadinov:
Actually in FSF JSON [1] it's identifier is "Python". So probably just rename identifier to Python.

[1] https://wking.github.io/fsf-api/licenses-full.json

Issue #1880, opened 2023-03-19
New license request: Open Logistics Foundation License v1.3


swinslow:
_edited 2023-03-20 by @swinslow to update URLs_

### Introduction

This license was approved by OSI at its 2023-03-19 Board meeting, as posted at http://lists.opensource.org/pipermail/license-review_lists.opensource.org/2023-March/005366.html

(Note that, although that particular thread indicated that the name was "Open Logistics License version 1.3", the preceding email thread included a [PDF summary](https://lists.opensource.org/pipermail/license-review_lists.opensource.org/attachments/20230312/71254ca6/attachment-0001.pdf) which makes it clear that "Open Logistics Foundation License v1.3" was the name of the approved final license.)

As an OSI-approved license, it is by default approved for inclusion on the SPDX License List as described in the [license inclusion principles](https://github.com/spdx/license-list-XML/blob/main/DOCS/license-inclusion-principles.md). However, we'll want to make sure there's agreement on the ID to use for the license.

### License Name

Open Logistics Foundation License v1.3

### Suggested short identifier

OLFL-1.3

### URL to license text

Corrected encoding: https://gist.github.com/marbre/86635347bf69315cfa84f32788bd80f0

_OSI approval version with incorrect encoding: http://lists.opensource.org/pipermail/license-review_lists.opensource.org/2023-January/005332.html_

### OSI Status

Approved

### License author or steward

Open Logistics Foundation

### URL to project(s) that use license

Not known, but irrelevant due to OSI approval

swinslow:
A couple additional notes:

* The OSI license-review mailing list thread at http://lists.opensource.org/pipermail/license-review_lists.opensource.org/2023-January/005332.html appears to include the v1.3 / approved version of the license. The Open Logistics Foundation still has the old (v1.0) version of their license on their website at https://openlogisticsfoundation.org/licenses/, but that version should _not_ be used as it was not the OSI-approved version.
* In that mailing list thread, the "bullets" for the main section numbering appear to use an encoding that got mangled by the mailing list software (e.g., `Â§1 Definitions`). We'll probably want to take a look at how the license is rendered elsewhere to see how this should be represented.

swinslow:
@jlovejoy @bsdimp @copernicat and others, grateful for your thoughts on the license ID I proposed above. Once there's agreement on that, we can mark this as approved given the OSI approval. Thanks!

bsdimp:
[...]

PR #1930, opened 2023-04-13
Add OLFL-1.3

marbre:
The OLFL-1.3 was accepted in #1880.

### DIFF
--- a/src/OLFL-1.3.xml
+++ b/src/OLFL-1.3.xml
@@ -1,0 +1,337 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<SPDXLicenseCollection xmlns="http://www.spdx.org/license">
+   <license isOsiApproved="true" licenseId="OLFL-1.3" name="Open Logistics Foundation License Version 1.3">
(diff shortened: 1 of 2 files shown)


jlovejoy:
thanks @marbre - I think there might be some missing <p> tags, but running the tests and will see what it catches.

Also, the Notes say the license was releases in 2013, shouldn't that be 2023?

marbre:
> thanks @marbre - I think there might be some missing
> 
> tags, but running the tests and will see what it catches.

I used https://tools.spdx.org/app/validate/ to check the XML file to be syntactical correctness, so I hope it should be fine.
Anyway, and if I had missed any template fields, I am happy to fix this.

> Also, the Notes say the license was releases in 2013, shouldn't that be 2023?

Good catch, thanks a lot! I have just fixed that.

swinslow:
Looks good, thank you @marbre! I made just one change, to add in the License List release version for which this will be added (3.21). After the tests run again I'll merge it. Thanks for submitting this!

Issue #1888, opened 2023-03-22
New license request: Latex2e with translated notice permission


ppisar:
### Introduction

This is an augmented variant of [Latex2e](https://spdx.org/licenses/Latex2e.html) license. It has been approved by Fedora project  <https://gitlab.com/fedora/legal/fedora-license-data/-/issues/176>.

This is a difference against Latex2e license:

~~~~
--- Latex2e 2023-03-21 16:55:49.575042074 +0100
+++ new 2023-03-21 16:54:18.782689780 +0100
@@ -8,4 +8,6 @@
 notice identical to this one.
 
 Permission is granted to copy and distribute translations of this manual
-into another language, under the above conditions for modified versions.
+into another language, under the above conditions for modified versions,
+except that this permission notice may be stated in a translation approved
+by the Foundation.
~~~~

### License Name

Latex2e with a translated notice permission

### Suggested short identifier

Latex2e-translated-notice

### URL to license text

https://git.savannah.gnu.org/cgit/indent.git/tree/doc/indent.texi?id=a74c6b4ee49397cf330b333da1042bffa60ed14f#n74

### OSI Status

I don't know

### License author or steward

Unknown, maybe Free Software Foundation 

### URL to project(s) that use license

<https://www.gnu.org/software/indent/>

richardfontana:
I think it would be unfortunate to use "Latex2e" in the name of the identifier for this license, since I believe it likely predates Latex2e. Though that is probably also true of the Latex2e license.

ppisar:
Do you have any suggestion for a better name? What about "FSF Verbatim Copying"?

jlovejoy:
@richardfontana - ironically, I believe SPDX choose "Latex2e" as the name for https://spdx.org/licenses/Latex2e.html because it was what Fedora used and added to SPDX back in the 2014 timeframe when we added a bunch of licenses from Fedora.
see https://fedoraproject.org/wiki/Licensing/Latex2e

Given we are not going to go back and change that now, I think the name @ppisar suggested or something similar is appropriate, as it maintains the closeness of the two licenses

jlovejoy:
oh, and I'm +1 to add for the obvious reason that this is similar to something we already have on the list and is used in a project that is used widely

swinslow:
[...]

PR #1932, opened 2023-04-14
New license: Latex2e-translated-notice

ppisar:
A new license similar, yet distinct from Latex2e.

https://github.com/spdx/license-list-XML/issues/1888

### DIFF
--- a/src/Latex2e-translated-notice.xml
+++ b/src/Latex2e-translated-notice.xml
@@ -1,0 +1,47 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<SPDXLicenseCollection xmlns="http://www.spdx.org/license">
+   <license isOsiApproved="false" licenseId="Latex2e-translated-notice"
+   name="Latex2e with translated notice permission" listVersionAdded="3.21">
+      <crossRefs>
+         <crossRef>https://git.savannah.gnu.org/cgit/indent.git/tree/doc/indent.texi?id=a74c6b4ee49397cf330b333da1042bffa60ed14f#n74</crossRef>
+      </crossRefs>
+      <notes>The last section differs from Latex2e license.</notes>
+      <text>
+         <copyrightText>
+            <p>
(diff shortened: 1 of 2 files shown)


jlovejoy:
@goneall - the text file has "Copyright @copyright{}" but the XML file has " Copyright ©" - it's passing tests here, but I'm wondering if that might cause problems in the license-list-data or publisher? 
Would it be better if the XML mirrored the actual text?

goneall:
>  it's passing tests here, but I'm wondering if that might cause problems in the license-list-data or publisher?

We use the same code in the tests as the license list publisher, so if it passes the test we'll be fine in the publisher.  Since the data is used by a lot of other tools, this difference may cause issues, however.

> Would it be better if the XML mirrored the actual text?

I think it would be better to have them match.

ppisar:
I will extend XML to accept both forms. Though, I thing it's pointless because copyright block is completely optional and ignored. My original idea was present the meaning in the XML, while use a raw code in the textual test.

ppisar:
At the and I removed the alternations because I was unable to persuade licenseListPublisher-2.2.8.jar to deal with it. Thus now the XML file uses "@copyright{}" as is written in the text file.

jlovejoy:
@goneall - I'm going to merge this, but not sure if @ppisar comment above about licenseListPublisher-2.2.8.jar not dealing with it is something you want to know about?

goneall:
> I'm going to merge this, but not sure if @ppisar comment above about licenseListPublisher-2.2.8.jar not dealing with it is something you want to know about?

I have not investigated this.  @ppisar - If you could [create an issue in the LicenseListPublisher](https://github.com/spdx/LicenseListPublisher/issues) and attach the test and XML files, I can take a look.  Since I'm helping with the SPDX 3.0 release, I may not get to it before Open Source Summit since it looks like you have a workaround.

ppisar:
I filed <https://github.com/spdx/LicenseListPublisher/issues/162>.

Issue #1550, opened 2022-08-03
New license request: ASWF Digital Assets License 1.0


zvr:
- Full name: ASWF Digital Assets License version 1.0
- Short identifier: ASWF-Digital-Assets-1.0
- Other web pages for license/exception: https://github.com/AcademySoftwareFoundation/foundation/tree/main/digital_assets
- Notes:
- OSI Approved:
- FSF Free/Libre:
- Text:

ASWF Digital Assets License v1.0

License for <Asset Name> (the “Asset Name”).

<Asset Name> Copyright <Year> <Asset Owner>. All rights reserved.

Redistribution and use of these digital assets, with or without modification, solely for education, training, research, software and hardware development, performance benchmarking (including publication of benchmark results and permitting reproducibility of the benchmark results by third parties), or software and hardware product demonstrations, are permitted provided that the following conditions are met:

1. Redistributions of these digital assets or any part of them must include the above copyright notice, this list of conditions and the disclaimer below.

2. Publications showing images derived from these digital assets must include the above copyright notice.

3. The names of copyright holder or the names of its contributors may NOT be used to promote or to imply endorsement, sponsorship, or affiliation with products developed or tested utilizing these digital assets or benchmarking results obtained from these digital assets, without prior written permission from copyright holder.

4. The assets and their output may only be referred to as the Asset Name listed above, and your use of the Asset Name shall be solely to identify the digital assets. Other than as expressly permitted by this License, you may NOT use any trade names, trademarks, service marks, or product names of the copyright holder for any purpose.

[...]

PR #1956, opened 2023-04-26
New license request: ASWF Digital Assets License 1.0

BassCoder2808:
- Solves #1550 
- Added the relevant xml and txt file for the ASWF Digital Assets License 1.0


BassCoder2808:
Hi @jlovejoy, I have created the PR, can you please let me know if anything else needs to be changed from my end. Also a job was failing for this, but I was not able to figure out, exactly whats the error, can you also please help me in that.

jlovejoy:
A couple things I see wrong (which may or may not be causing the tests to fail!):

- missing License List version tag
- no need to repeat title outside of titleText tag 
- xml uses "" in first couple lines, whereas text file and original file uses <> e.g., `License for <Asset Name> ` 

this must be the same. I just added some guidance on that last item here: https://github.com/spdx/license-list-XML/blob/main/DOCS/xml-fields.md#other-things-to-watch-out-for

BassCoder2808:
Thanks @jlovejoy for responding back, I will make the changes which you have mentioned above

src/ASWF-Digital-Assets-1.0.xml
```diff
@@ -0,0 +1,76 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<SPDXLicenseCollection xmlns="http://www.spdx.org/license">
+   <license isOsiApproved="false" licenseId="ASWF-Digital-Assets-1.0"
+   name="ASWF Digital Assets License version 1.0" listVersionAdded="1.0">
```
swinslow, line 4: `listVersionAdded` is intended to be the version of the SPDX License List release in which this license is first added. The upcoming release is v3.21, so this should be "3.21".

swinslow (changes requested):
@BassCoder2808 thanks for this! Just one minor change needed, see inline -- then this should be good to go, as far as I can tell from a quick look.

src/ASWF-Digital-Assets-1.0.xml
BassCoder2808: Thanks @swinslow, I will make that change right away

jlovejoy:
woo-hoo, everything is passing now!

jlovejoy:
@swinslow @BassCoder2808 - I'm going to merge now :)

PR #1964, opened 2023-05-11
Document what is not allowed in alt and optional tags

goneall:
Related to https://github.com/spdx/LicenseListPublisher/issues/162


jlovejoy:
Thanks @goneall

PR #1970, opened 2023-05-16
New exception: GPL-3.0-interface-exception

### LINKED ISSUE #1921: New license request: GPL-3.0-interface-exception
### Introduction

GPLv3+ variant with an exception not listed by SPDX. Discussed and approved for Fedora at https://gitlab.com/fedora/legal/fedora-license-data/-/issues/191. 

### License Name

GPL-3.0 Interface Exception

GPL-3.0-interface-exception

### URL to license text

https://www.gnu.org/licenses/gpl-faq.en.html#LinkingOverControlledInterface

### OSI Status

I don't know

### License author or steward

Unknown

https://gitlab.mpi-magdeburg.mpg.de/software/flexiblas-release

Enchufa2:
Closes #1921


Enchufa2:
My logic was that the same "name of the library" should have the same identifier, but maybe different locations should have different identifiers? Please @swinslow let me know if I need to change that.

jlovejoy:
@goneall - does it matter if the name in the alt tag is the same across various instances?  It's passing the tests, so I'm assuming this is fine, but for some reason all the examples of this kind of thing that I've seen have them numbered (even if it's ostensibly the same value)

goneall:
> does it matter if the name in the alt tag is the same across various instances?

@jlovejoy technically, this should fail since the spec states that for the name field "This field is unique within each license template."

However, we don't check when verifying the license.  We could consider it a bug in the licenseListPublisher, but I'm afraid what will happen if we try to fix it - it may cause a lot of failures since we've never checked.

Enchufa2:
Changed to incremental numbers.

Issue #1973, opened 2023-05-16
Classpath-exception-2.0: var or optional text to handle variation in reference to GPL-2.0 license?


pmonks:
In [this license text that includes Classpath-exception-2.0](https://github.com/javaee/javamail/blob/master/LICENSE.txt#L746), the reference to the GPL-2.0 license includes the version number ("GNU General Public License version 2"), but the SPDX exception template only includes the license name ("GNU General Public License") which means that matching of this text via the template fails.  This was originally discovered in [this Spdx-Java-Library issue](https://github.com/spdx/Spdx-Java-Library/issues/166).

In that issue, @goneall suggested that this variation could be handled in the Classpath-exception-2.0 template via the addition of a `<var>` or `<begin/end optional>` section to the template.

jlovejoy:
agreed, we could simply put "version 2" in an optional tag

@pmonks - do you want to prepare a PR for this?

PR #1976, opened 2023-05-18
Address issue 1973

pmonks:
This PR should address issue #1973.

That said, I'm unfamiliar with what (if any) tooling exists to validate that my changes are correct (i.e. that they're both syntactically valid and also correctly address the matching issue raised in issue #1973). Suggestions/edits of this PR to ensure both of those things are encouraged!

### DIFF
--- a/src/GPL-2.0-with-classpath-exception.xml
+++ b/src/GPL-2.0-with-classpath-exception.xml
@@ -13,23 +13,8 @@
     <text>
 	     <p>insert GPL v2 license text here</p>
 	     <p>Class Path Exception</p>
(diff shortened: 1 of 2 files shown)


jlovejoy:
@pmonks - I'm not sure it's worth updating src/GPL-2.0-with-classpath-exception.xml at that is deprecated??

pmonks:
@jlovejoy I erred on the side of caution in case there are users who are still matching using deprecated licenses. Happy to remove that file from the PR if you're confident the inconsistency won't cause issues for consumers.

pmonks:
@jlovejoy done!

PR #1977, opened 2023-05-19
fix recent exceptions

jlovejoy:
needed exception tag and to be in exceptions folder - oops!


jlovejoy:
so, I screwed these up by using the <license> tag and putting them in the license folder, instead of <exception> tag and exceptions folder, but the tests passed this so it was merged, then I realized the mistake!

@rtgdk @BassCoder2808 - FYI - not sure if this is something that could be tested for? like if the id uses "exception" is throws some kind of warning that maybe this is an exception, not a license?

BassCoder2808:
Yes, @jlovejoy I think this can be done by adding some more tests in the `validate.yaml` in `github_workflows`. I will look into this and will let you know if I find any solutions for this

swinslow (approved):

Issue #1947, opened 2023-04-20
New license request: Linux-man-pages-copyleft-2-para (was: Latex2e-no-translation [SPDX-Online-Tools])


adobes1:
**1.** License Name: Latex2e without translation permission
**2.** Short identifier: Latex2e-no-translation
**3.** License Author or steward: Unknown
**4.** Comments: Latex2e variant without the last paragraph about translation distribution.
**5.** License Request Url: http://tools.spdx.org/app/license_requests/229
**6.** URL(s): https://github.com/mkerrisk/man-pages/blob/4f060b54679bcdfa3c1709ff6a984d8d85924fc2/man2/move_pages.2#L5-L12, https://github.com/mkerrisk/man-pages/blob/4f060b54679bcdfa3c1709ff6a984d8d85924fc2/man2/migrate_pages.2#L8-L15
**7.** OSI Status: Unknown
**8.** Example Projects: https://github.com/mkerrisk/man-pages

adobes1:
Link to related Fedora issue: https://gitlab.com/fedora/legal/fedora-license-data/-/issues/201

jlovejoy:
+1 to add as another variation that is (annoyingly just different enough from https://spdx.org/licenses/Latex2e.html and #1888 

can I get some other eyes on this? @swinslow @copernicat @richardfontana

Pizza-Ria:
It would be nice to be able to just make that clause optional in Latex2(e) as I'm generally opposed to minting all of these variants but I guess there really isn't a path to do that.  So, count me as an annoyed +1.

swinslow:
+1 on adding and +1 on wishing it wasn't another variation  :)

I'm not sure I would go with `Latex2e-no-translation` as the identifier, though. That could be misread as saying that translations are not permitted -- and probably wouldn't make sense to someone seeing it on its own, separate from `Latex2e`.

Maybe something like `Latex2e-2-Paragraph`, similar to `BSD-2-Clause`? That also kind of reflects the `VERBATIM_TWO_PARA` tag in the sample file.

adobes1:
The URLs I included in this request as the URLs of the license text point to outdated mirror site which I unfortunately didn't realize. 
The correct ones should be: https://git.kernel.org/pub/scm/docs/man-pages/man-pages.git/tree/man2/move_pages.2#n5 and https://git.kernel.org/pub/scm/docs/man-pages/man-pages.git/tree/man2/migrate_pages.2#n8. 
[...]

PR #1978, opened 2023-05-19
Update Linux-man-pages-1-para.xml

### LINKED ISSUE #1955: New license request: Linux-man-pages-1-para [SPDX-Online-Tools]
**1.** License Name: Linux man-pages One Paragraph
**2.** Short identifier: Linux-man-pages-one-para
**4.** Comments: This license appears in several Linux man-pages under the name of VERBATIM_ONE_PARA
**5.** License Request Url: http://tools.spdx.org/app/license_requests/231
**6.** URL(s): https://github.com/mkerrisk/man-pages/blob/master/man2/getcpu.2#L4-L7
**7.** OSI Status: Unknown
**8.** Example Projects: https://github.com/mkerrisk/man-pages

jlovejoy:
fixes #1955 

updates to name and id as per request by maintainer of man-pages, @alejandro-colomar 

note @adobes1


alejandro-colomar:
I change my original suggestion, and now suggest `Linux-man-pages-copyleft-1-para` (with a full name of: `Linux man-pages Copyleft (1-paragraph variation)`), for consistency with <https://github.com/spdx/license-list-XML/issues/1947>.  Sorry for the inconvenience :)

alejandro-colomar:
BTW, you'll also need to `mv src/Linux-man-pages-one-para.xml src/Linux-man-pages-copyleft-1-para.xml`.

jlovejoy:
ok @alejandro-colomar - last change of mind, please!

I did leave off "variation" in the full name as I don't think that's necessary

Please give a thumbs up and then we'll merge this (again) :)

alejandro-colomar:
Actually, as @g-branden-robinson suggested in the other thread, maybe it's better to be specific as to how does it differ from Linux-man-pages-copyleft, instead of how many paragraphs it has, as some future license may pick a different single paragraph from it.

How about this? `Linux-man-pages-copyleft-no-mod` - `Linux man-pages Copyleft (no modifications)`  The reason is that this variant only allows verbatim copies, while the longer variants allow modifications under the same license.

jlovejoy:
@alejandro-colomar - "no modifications" sounds like modifications are not allowed, which is not the case so I think that could be misleading

jlovejoy:
I think either `Linux-man-pages-one-para` or `Linux-man-pages-1-para` is the way to go. I don't see how "copyleft" really fits here as this license is much less specific to that end than https://spdx.org/licenses/Linux-man-pages-copyleft.html

Can we please decide on one of those and get this closed?  :)

g-branden-robinson:
My name was spoken, thus I am summoned in a puff of brimstone.  I agree with @jlovejoy that "no-mod" and "no modifications" are misleading (show-stopperingly so in my opinion, but SPDX hasn't been my show for years).

However I disagree that the fact of copyleft is not material here.  Reviewing the text:

`Permission is granted to distribute possibly modified copies of this page provided the header is included verbatim, and in case of nontrivial modification author and date of the modification is added to the header.`

If the license terms are part of the "header", as they do appear to be, then this is in fact a copyleft.  "Permission is granted" to all who read the terms (implicitly: as long as they uphold said terms).  No intermediate licensee B between licensees A and C can delete any granted permission, and that is the essence of copyleft  (IANAL; TINLA).

As with issue #1947 I think it is a bad idea to put the length of the license text in paragraphs into the license description, short or long.  What happens when some clever person re-uses some license text with a paragraph break deleted?

It is becoming clear to me that the "Linux man-pages copyleft" is not one license but several.  There appears to be a problem of proliferation.  Because @alejandro-colomar is an energetic person, I propose suspending the addition of further Linux man-pages copyleft licenses to SPDX until he has had a chance to attack this problem.  I think it would be a better world for everyone if all the man pages that project distributes were under terms already in the SPDX license list.  As a sometime contributor to that project's mailing list, I can perhaps be of some assistance.

alejandro-colomar:
> @alejandro-colomar - "no modifications" sounds like modifications are not allowed, which is not the case so I think that could be misleading

Sorry, I misread.  I thought I had read the first paragraph of Linux-man-pages-copyleft, which is different.

alejandro-colomar:
> My name was spoken, thus I am summoned in a puff of brimstone. I agree with @jlovejoy that "no-mod" and "no modifications" are misleading (show-stopperingly so in my opinion, but I haven't been an SPDX cast or crew member for years).
> 
> However I disagree that the fact of copyleft is not material here. Reviewing the text:
> 
> `Permission is granted to distribute possibly modified copies of this page provided the header is included verbatim, and in case of nontrivial modification author and date of the modification is added to the header.`
> 
> If the license terms are part of the "header", as they do appear to be, then this is in fact a copyleft. "Permission is granted" to all who read the terms (implicitly: as long as they uphold said terms). No intermediate licensee B between licensees A and C can delete any granted permission, and that is the essence of copyleft (IANAL; TINLA).
> 
> As with issue #1947 I think it is a bad idea to put the length of the license text in paragraphs into the license description, short or long. What happens when some clever person re-uses some license text with a paragraph break deleted, or modifies this one to turn it into two paragraphs, perhaps at "verbatim. ¶In case of nontrivial"?
> 
> It is becoming clear to me that the "Linux man-pages copyleft" is not one license but several. There appears to be a problem of proliferation. Because @alejandro-colomar is an energetic person, I propose suspending the addition of further Linux man-pages copyleft licenses to SPDX until he has had a chance to attack this problem. I think it would be a better world for everyone if all the man pages that project distributes were under terms already in the SPDX license list. As a sometime contributor to that project's mailing list, I can perhaps be of some assistance.

@g-branden-robinson Do you propose that we start a mission in linux-man@ of asking Copyleft holders to relicense into the "standard" version of the license?

g-branden-robinson:
> Do you propose that we start a mission in linux-man@ of asking Copyleft holders to relicense into the "standard" version of the license?

@alejandro-colomar, pretty much, and I see you already have this underway.  :)

jlovejoy:
see #1955 for final discussion/rationale on name :)

swinslow (approved):
It was a long journey but looks like we're good here  :)

PR #1997, opened 2023-06-08
Fix GPL URLs to use https

### LINKED ISSUE #1995: GPL-3.0-only.txt out-of-sync with XML
My project uses FSFE REUSE, which automatically downloads licenses in TXT format from https://github.com/spdx/license-list-data.

The GPL-3.0-only.txt file contains `http` URLs, which our security scanner (DevSkim) complains about.

The original XML was altered to use `https` in this commit: https://github.com/spdx/license-list-XML/commit/c92c09fc3dd1740dc29fe726482940ebf35fb5d0. That was in May 2021. However, GPL-3.0-only.txt (and probably others) was never regenerated and predates that change.

dorssel:
Scope:

- all files GPL*.txt (not the deprecated ones)
- URLs to fsf.org and gnu.org

Fixes #1995

### DIFF
--- a/test/simpleTestForGenerator/GPL-3.0-only.txt
+++ b/test/simpleTestForGenerator/GPL-3.0-only.txt
@@ -1,7 +1,7 @@
 GNU GENERAL PUBLIC LICENSE
 Version 3, 29 June 2007
 
-Copyright © 2007 Free Software Foundation, Inc. <http://fsf.org/>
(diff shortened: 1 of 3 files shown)


jlovejoy:
Thanks @dorssel - I put you back as "assigned to" as that is how we track participation of PRs :)

PR #2001, opened 2023-06-09
New Exception: LLGPL

### LINKED ISSUE #1988: New license request: LLGPL [SPDX-Online-Tools]
**1.** License Name: Lisp exception
**2.** Short identifier: Lisp-exception
**4.** Comments: Specific exception to LGPL used in Fedora, Slime package

Although it's referred to as a "preamble", it reads more like an exception as it states that "Accordingly, the license for the open-source Lisp applications consists of this document plus the LGPL. Wherever there is a conflict between this document and the LGPL, this document takes precedence over the LGPL."
and also states the section 6 is not applicable.
**5.** License Request Url: http://tools.spdx.org/app/license_requests/237
**6.** URL(s): http://opensource.franz.com/preamble.html
**7.** OSI Status: Unknown
**8.** Example Projects: https://github.com/slime/slime/blob/master/swank/ccl.lisp

jlovejoy:
fixes #1988


jlovejoy:
@goneall @swinslow  - is there a way to use alternative text between some text and nothing? (but not use optional)?

src/exceptions/LLGPL.xml
```diff
@@ -0,0 +1,97 @@
...
+            The concept of the GNU Lesser General Public License version 2.1
+            ("LGPL") has been adopted to govern the use and distribution of
+            <alt match="above-mentioned application|AllegroServe" name="app1">above-mentioned application</alt>. 
+            However, the LGPL uses terminology
+            that is more appropriate for a program written in C than one
+            written in Lisp. Nevertheless, the LGPL can still be applied to
+            a Lisp program if certain clarifications are made. This document
+            details those clarifications. Accordingly, the license for
+            <alt match="the open-source Lisp applications|AllegroServe" name="app">the
+            open-source Lisp applications</a> consists of this document plus
```
swinslow, line 27: @jlovejoy I think this is failing because the opening tag is `<alt>` but the closing tag is `</a>`. I'm going to fix it and see if that resolves this one.

swinslow:
@jlovejoy So that resolved the prior failure. The current one is failing at the first word: "Preamble" in the XML file (reflecting http://opensource.franz.com/preamble.html dated 2000) vs. "Prequel" in the test text file (reflecting https://raw.githubusercontent.com/franzinc/aserve/master/LICENSE dated 2016).

It looks like you were trying to incorporate some of the AllegroServe language into the alt markups to pick up the 2016 text. However, note that [the 2016 text](https://raw.githubusercontent.com/franzinc/aserve/master/LICENSE), in addition to using "Prequel" rather than "Preamble", has an entirely different intro paragraph: "Franz Inc. has adopted the concept of..."

How would you like to handle these? Do we want markup to be able to match to both of the versions linked above, or do we want to take out the AllegroServe 2016 markup and just match to the 2000 version?

jlovejoy:
@swinslow - changed to "preamble" in text file, since I don't think we need to accommodate differences in a title anyway.

Yeah, I wasn't sure how to deal with the differences in the first sentence - it's not an easy one for an alt tag in my experience... Can we have two sentences for options in an alt tag? As in:
`match="Franz Inc. has adopted the concept of the GNU Lesser General Public License version 2.1 ("LGPL") to govern the use and distribution of
AllegroServe.|The concept of the GNU Lesser General Public License version 2.1 has been adopted to govern the use and distribution of above-mentioned application."`

or maybe:
`match="Franz Inc. has adopted the concept of the GNU Lesser General Public License version 2.1 ("LGPL")|The concept of the GNU Lesser General Public License version 2.1 has been adopted" `
and then
`match="AllegroServe|above-mentioned application"`


I also had the thought of only capturing the "generic" iteration on the website, but... since we know of a variation, I thought I should accommodate it...

swinslow:
@jlovejoy This might work, but would need escaping characters in the `match` regular expression string to accommodate characters like `.` -- let me try it out and see what I can do.

swinslow:
Ugh, I think I fixed the license text but screwed up the merge -- now pulled in a bunch of interim commits. Rather than untangle this, I'm going to submit a new PR.

swinslow:
Replaced with #2014

PR #2014, opened 2023-06-18
LLGPL: Add fixed markup and test text

### LINKED ISSUE #1988: New license request: LLGPL [SPDX-Online-Tools]

swinslow:
Fixes #1988 

Replaces #2001 because I broke the commit tree there

cc @jlovejoy 

### DIFF
--- a/src/exceptions/LLGPL.xml
+++ b/src/exceptions/LLGPL.xml
@@ -1,0 +1,96 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<SPDXLicenseCollection xmlns="http://www.spdx.org/license">
+   <exception licenseId="LLGPL"
+   name="LLGPL Preamble" listVersionAdded="3.21">
+      <crossRefs>
+         <crossRef>http://opensource.franz.com/preamble.html</crossRef>
+      </crossRefs>
+      <notes>
+         Typically used with LGPL-2.1
+      </notes>
+      <text>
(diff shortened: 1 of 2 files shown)


swinslow:
@jlovejoy To fix the failing errors in #2001 I inserted both of the intro lines and marked each as optional. This isn't ideal, since both will show (marked as optional) in the text on the website.

But it is at least valid markup that will match to both of the existing license text variants; and I don't have time to untangle the regular expressions if 3.21 will get pushed today. So I'm going to go ahead and merge once the tests pass, and we can improve it in 3.22 if people want.

I also added a number of `<optional>` tags around the word `the` before `LGPL` in several places throughout the text, because the two variants differed on that in several places.

PR #2021, opened 2023-07-02
Replace wrong placeholder in documentation

nicoroeser:
Probably a no-brainer. See commit for details.


swinslow:
Yup, good catch -- thanks @nicoroeser!

PR #2054, opened 2023-07-29
New License: Soundex

xsuchy:
Fixes: 2018


jlovejoy:
I removed the standard header and notes tags as if they are blank, we don't need them

Issue #2028, opened 2023-07-10
New license request: Knuth-MMIXware [SPDX-Online-Tools]


xsuchy:
**1.** License Name: Knuth-MMIXware
**2.** Short identifier: Knuth-MMIXware
**3.** License Author or steward: Unknown
**4.** Comments: Found in Fedora during License review in https://gitlab.com/fedora/legal/fedora-license-data/-/issues/226
Used in project mmixware
**5.** License Request Url: http://tools.spdx.org/app/license_requests/248
**6.** URL(s): https://gitlab.lrz.de/mmix/mmixware/-/blob/master/boilerplate.w
**7.** OSI Status: Unknown
**8.** Example Projects: https://gitlab.lrz.de/mmix/mmixware/

jlovejoy:
agreed, +1 to add as well

ppisar:
What is a correct  capitalization of this identifier? "MMIXWare" or "MMIXware" ("W" letter)?

I ask because there is a dissonance:

* Upstream speels it as "MMIXware" <https://gitlab.lrz.de/mmix/mmixware/-/blob/master/README>
* Winslow defined it as "MMIXWare"
* Git repository contains a file called "src/MMIXWare.xml"
* The file defined licenseId as "MMIXware"

I would like to unify the file name and licenseId. I only don't know which spelling to use. I would lean to upstream's spelling with lower case "w".

xsuchy:
Good catch. I would prefer MMIXware to match upstream. If I get some thumbs up, I will prepare PR.

jlovejoy:
indeed, good catch @ppisar !

yes, let's fix this and matching upstream sounds good. Thanks @xsuchy !

PR #2060, opened 2023-08-04
New license: MMIXWare

xsuchy:
Fixes: #2028

Submitted with the name changed to what @swinslow suggested.

### DIFF
--- a/src/MMIXWare.xml
+++ b/src/MMIXWare.xml
@@ -7,7 +7,7 @@
       </crossRefs>
       <text>
         <copyrightText>
-          <p>copyright YEAR AUTHOR</p>
+          <p>copyright 1999 Donald E. Knuth</p>
         </copyrightText>
          <p>
             This file may be freely copied and
--- a/src/MMIXWare.xml
+++ b/src/MMIXWare.xml
@@ -1,0 +1,31 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<SPDXLicenseCollection xmlns="http://www.spdx.org/license">
+   <license isOsiApproved="false" licenseId="MMIXware"
+   name="MMIXware License" listVersionAdded="3.22">
+      <crossRefs>
+         <crossRef>https://gitlab.lrz.de/mmix/mmixware/-/blob/master/boilerplate.w</crossRef>
+      </crossRefs>
+      <text>
+        <copyrightText>
+          <p>copyright YEAR AUTHOR</p>
+        </copyrightText>
+         <p>
+            This file may be freely copied and
+            distributed, provided that no changes whatsoever are
+            made. All users are asked to help keep the MMIXware files
+            consistent and ``uncorrupted,'' identical everywhere in the
+            world. Changes are permissible only if the modified file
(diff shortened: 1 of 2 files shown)


src/MMIXWare.xml
```diff
@@ -0,0 +1,32 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<SPDXLicenseCollection xmlns="http://www.spdx.org/license">
+   <license isOsiApproved="false" licenseId="MMIXware"
+   name="MMIXware License" listVersionAdded="3.22">
+      <crossRefs>
+         <crossRef>https://gitlab.lrz.de/mmix/mmixware/-/blob/master/boilerplate.w</crossRef>
+      </crossRefs>
+      <text>
+         <list>
+            <item>
```
jlovejoy, line 10: not sure why the list and item tags got in here, but not needed

src/MMIXWare.xml
```diff
@@ -0,0 +1,32 @@
...
+               <p>
+                  copyright 1999 Donald E. Knuth This file may be freely copied and
```
jlovejoy, line 12: copyright notice needs to be enclosed in copyrightText tag

src/MMIXWare.xml
xsuchy: SPDX online tools generated that. I will remove it.

src/MMIXWare.xml
xsuchy: I checked how other licenses in src/ do that and I do that similarly. Tell me if needs to be done differently.

xsuchy:
Updated.

jlovejoy:
@swinslow @goneall - wondering if either of you have opinions on the "genercizing" of the copyright notice. I know some people have asked for that on specific licenses and we have sometimes obliged, but do we want to do that preemptively?

in this case, .txt file has name of copyright holder from original source and .xml file has YEAR, NAME - it's passing tests, but I sort am incline this should match b/w the files in any case?  I know the copyright notice isn't "matched", but for visualization of license, it might be good if we had a definitive policy (and documented it in the xml guidance doc)??

goneall:
> wondering if either of you have opinions on the "genercizing" of the copyright notice. I know some people have asked for that on specific licenses and we have sometimes obliged, but do we want to do that preemptively?

No opinion on the policy itself, but +1 on having a policy and documenting it.  As long as the copyright text is wrapped in the `<copyrightText>` element tag, the license list publisher will create an "alt" text so that any text will match the copyright text itself.

jlovejoy:
thanks @goneall - I'm going to boot the policy decision to a later time as I'd like to get this merged.

In this case, I think we should just match the text file but using "1999 Donald E. Knuth" in the copyright notice (which is contained in the copyrightText tag only) in the .xml file as well. It's a project-specific license as far as we know. If we later find other uses with a different project (and copyright holder) we can then add markup and genericize the copyright notice at that time

jlovejoy:
@xsuchy - I just went ahead and fixed that one thing myself. not sure if that's proper git etiquette, but trying to just get stuff done... :)

PR #2062, opened 2023-08-04
Update MIT-testregex.xml

jlovejoy:
fixed incorrect XML list tags due to how license was entered into SPDX submission form.

FYI @xsuchy  (I merged this too quickly and caught it afterwards)

@swinslow - the txt file for this license includes the code * at the start of each line. I usually remove these, but made me wonder if that is necessary or we have any guideline around that? https://github.com/spdx/license-list-XML/blob/main/test/simpleTestForGenerator/MIT-testregex.txt

### DIFF
--- a/src/MIT-testregex.xml
+++ b/src/MIT-testregex.xml
@@ -3,12 +3,10 @@
    <license isOsiApproved="false" licenseId="MIT-testregex"
    name="MIT testregex Variant" listVersionAdded="3.22">
       <crossRefs>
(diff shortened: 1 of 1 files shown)


xsuchy:
I did not included it, because
```
/* comment */
```
is just syntax for comments and people often do:
/* one 
 * two 
 * /
```
instead of
```
/* one
    two
*/
```
The former has benefits that you clearly see that you are insight of long multiline comment.  It does not have any semantical meaning.

Issue #2026, opened 2023-07-10
New license request: TTYP0 [SPDX-Online-Tools]


xsuchy:
**1.** License Name: TTYP0
**2.** Short identifier: TTYP0
**3.** License Author or steward: Unknown
**4.** Comments: MIT-like license. Found during Fedora license review in https://gitlab.com/fedora/legal/fedora-license-data/-/issues/213
Used in package easyrpg-player
**5.** License Request Url: http://tools.spdx.org/app/license_requests/246
**6.** URL(s): https://raw.githubusercontent.com/EasyRPG/Player/0.8/resources/ttyp0/LICENSE
**7.** OSI Status: Unknown
**8.** Example Projects: https://github.com/EasyRPG/Player/

swinslow:
License text for reference:

```
THE TTYP0 LICENSE

Permission is hereby granted, free of charge, to any person obtaining
a copy of this font software and associated files (the "Software"),
to deal in the Software without restriction, including without
limitation the rights to use, copy, modify, merge, publish,
distribute, embed, sublicense, and/or sell copies of the Software,
and to permit persons to whom the Software is furnished to do so,
subject to the following conditions:

(1) The above copyright notice, this permission notice, and the
    disclaimer below shall be included in all copies or substantial
    portions of the Software.

(2) If the design of any glyphs in the fonts that are contained in the
    Software or generated during the installation process is modified
    or if additional glyphs are added to the fonts, the fonts
    must be renamed. The new names may not contain the word "UW",
    irrespective of capitalisation; the new names may contain the word
    "ttyp0", irrespective of capitalisation, only if preceded by a
    foundry name different from "UW".

THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
```

[...]

PR #2064, opened 2023-08-05
New license: TTYP0

xsuchy:
Fixes: #2026

I updated the licenseName and moved the title to titleText (compared to what SPDX online tool generated).


n8willis:
(As [mentioned in](https://github.com/spdx/license-list-XML/issues/2026#issuecomment-1666497407) the issue thread): The Fedora "easyrpg" package is not the upstream for the font or for the license. The upstream is https://people.mpi-inf.mpg.de/~uwe/misc/uw-ttyp0/

License author/steward and URL should probably reflect that. I don't think it's ambiguous that it's the original license author; if there's some process needed to verify that, I'm not aware of what's needed, or I'd offer to help.

xsuchy:
Updated crossref. Let me know if it need any other change.

Issue #2039, opened 2023-07-20
New license request: stunnel-exception [SPDX-Online-Tools]


xsuchy:
**1.** License Name: stunnel exception
**2.** Short identifier: stunnel-exception
**3.** License Author or steward: Unknown
**4.** Comments: This exception is used in stunnel project and has been found during Fedora License Review https://gitlab.com/fedora/legal/fedora-license-data/-/issues/247
**5.** License Request Url: http://tools.spdx.org/app/license_requests/253
**6.** URL(s): https://github.com/mtrojnar/stunnel/blob/master/COPYING.md
**7.** OSI Status: Unknown
**8.** Example Projects: stunnel: upstream at https://www.stunnel.org/, browsable source code mirror at https://github.com/mtrojnar/stunnel/

copernicat:
Although the project's website says it is an "OpenSSL exception", the text is quite different than the existing OpenSSL exceptions, and the SPDX diff tool shows no close matches. I agree with adding it. 
Do we want to include GPL in the license identifier as we have for other exceptions?

richardfontana:
Note that there is likely a typo in the exception text (the "special exception" paragraph mentions LGPL in the first sentence and GPL in the second sentence).

PR #2074, opened 2023-08-11
New exception: stunnel-exception

xsuchy:
Fixes: #2039


neverpanic (approved):

jlovejoy:
FYI @xsuchy  - I added <p> tags so it formats/looks same as text file. the XML editor formatted it as one big paragraph b/c it was pasted in to the tool/submission that way.

added to your commit, as it didn't seem worth bugging you to add them :)

Issue #2048, opened 2023-07-27
New license request: URT-RLE-Legacy [SPDX-Online-Tools]


xsuchy:
**1.** License Name: URT-RLE Legacy License
**2.** Short identifier: URT-RLE-Legacy
**3.** License Author or steward: Unknown
**4.** Comments: This license is presented in netpbm package and was found during the license review in Fedora. https://gitlab.com/fedora/legal/fedora-license-data/-/issues/256
**5.** License Request Url: http://tools.spdx.org/app/license_requests/257
**6.** URL(s): https://sourceforge.net/p/netpbm/code/HEAD/tree/super_stable/converter/other/pnmtorle.c, https://sourceforge.net/p/netpbm/code/HEAD/tree/super_stable/converter/other/rletopnm.c
**7.** OSI Status: Unknown
**8.** Example Projects: https://sourceforge.net/projects/netpbm/

jlovejoy:
license text for convenience:
```
 * This software is copyrighted as noted below.  It may be freely copied,
 * modified, and redistributed, provided that the copyright notice is
 * preserved on all copies.
 *
 * There is no warranty or other guarantee of fitness for this software,
 * it is provided solely "as is".  Bug reports or fixes may be sent
 * to the author, who may or may not act on them as he desires.
 *
 * You may not include this software in a program or other software product
 * without supplying the source, or without informing the end-user that the
 * source is available for no extra charge.
 *
 * If you modify this software, you should include a notice giving the
 * name of the person performing the modification, the date of modification,
 * and the reason for such modification.
 ```

jlovejoy:
I think this meets the inclusion guidelines as it's a lightweight copyleft-ish license. It also is used in a major distro, so meets the substantial use criteria.

I don't have a strong opinion as to ID and name

richardfontana:
For the ID, I would suggest something along the lines of 'URT-RLE' reflecting what I believe are the origins of this license (Utah Raster Toolkit - Run Length Encoded format - see e.g. http://paulbourke.net/dataformats/urt/). The 'legacy' was just something I put in the description in the Fedora License Data gitlab issue because I like to call obscure ancient licenses like these "legacy" licenses.

swinslow:
+1 to add, as a historical and copyleft (maybe?) license used in Fedora.

[...]

PR #2089, opened 2023-08-21
New License: URT-RLE

xsuchy:
Fixes: #2048


jlovejoy:
fixed odd bullet that online tool pulled in and added p tags to match formatting in actual source file

Issue #2050, opened 2023-07-27
New license request: SANE-exception [SPDX-Online-Tools]


xsuchy:
**1.** License Name: SANE exception
**2.** Short identifier: SANE-exception
**3.** License Author or steward: Unknown
**4.** Comments: This exception is presented in GPL-2.0-or-later license of SANE package. It was found during Fedora license review. https://gitlab.com/fedora/legal/fedora-license-data/-/issues/264
**5.** License Request Url: http://tools.spdx.org/app/license_requests/259
**6.** URL(s): https://github.com/alexpevzner/sane-airscan/blob/master/LICENSE, https://gitlab.com/sane-project/backends/-/blob/master/sanei/sanei_pp.c?ref_type=heads, https://gitlab.com/sane-project/frontends/-/blob/master/sanei/sanei_codec_ascii.c?ref_type=heads
**7.** OSI Status: Unknown
**8.** Example Projects: https://gitlab.com/sane-project/

xsuchy:
This exception is similar to gnu-javamail-exception but @jlovejoy thinks the differences are justifiable for separate id.
Also note that some use usage says "if you link a sane-airscan library" while others "if you link a SANE library".

swinslow:
Exception text for reference:

> As a special exception, the authors of sane-airscan give permission for
> additional uses of the libraries contained in this release of sane-airscan.
> 
> The exception is that, if you link a sane-airscan library with other files
> to produce an executable, this does not by itself cause the
> resulting executable to be covered by the GNU General Public
> License.  Your use of that executable is in no way restricted on
> account of linking the sane-airscan library code into it.
> 
> This exception does not, however, invalidate any other reasons why
> the executable file might be covered by the GNU General Public
> License.
> 
> If you submit changes to sane-airscan to the maintainers to be included in
> a subsequent release, you agree by submitting the changes that
> those changes may be distributed with this exception intact.
> 
> If you write modifications of your own for sane-airscan, it is your choice
> whether to permit this exception to apply to your modifications.
> If you do not wish that, delete this exception notice.

swinslow:
+1 to add this as a separate license exception. 

A number of exceptions (the javamail one, also GNAT-exception and Fawkes-Runtime-exception) contain portions of language that are similar, but I don't see any that would be a substantive match for this one. In particular the requirement that submitting changes to the maintainers means that those changes may be distributed with this exception intact is something I don't recall seeing in other exceptions.

[...]

PR #2090, opened 2023-08-21
New License: SANE-exception

xsuchy:
Fixes: #2050


jlovejoy:
merge is pending final decision on license inclusion

jlovejoy:
note - this is passing tests with same name for all variable fields in alt tag (@swinslow - just noting this, as I had this question/noticed this once before, but as tests passed, I think we figured it's ok)

Issue #2092, opened 2023-08-22
markup for x11vnc OpenSSL Exception


jlovejoy:
Fedora has found a variation on [x11vnc OpenSSL Exception](https://spdx.org/licenses/x11vnc-openssl-exception.html) that replaces the name of the person and project with more generic wording - see https://gitlab.com/fedora/legal/fedora-license-data/-/issues/302#note_1523442080

This seems like an obvious case for adding markup - agree?

jlovejoy:
https://git.einval.com/cgi-bin/gitweb.cgi?p=jigdo.git;a=blob_plain;f=README;hb=HEAD uses 
* `the author` instead of a name 
* `jigdo code` instead of "code of its release of x11vnc"
* `any jigdo source file` instead of "this file"

`In addition, as a special exception, the author gives permission to link the jigdo code with the OpenSSL project's "OpenSSL" library (or with modified versions of it that use the same license as the "OpenSSL" library), and to distribute the linked executables. You must obey the GNU General Public License in all respects for all of the code used other than "OpenSSL". If you modify any jigdo source file, you may extend this exception to your version of the file, but you are not obliged to do so. If you do not wish to do so, delete this exception statement from your version.`

swinslow:
Discussed on 2023-08-24 legal team call, general consensus that these changes are not substantively different, OK to accommodate with markup.

PR #2096, opened 2023-08-25
extend x11vnc OpenSSL Exception markup to match similar exceptions

xsuchy:
Fixes: #2092


jlovejoy:
@swinslow - can you check and merge this one? You are better at the markup options than I am!

swinslow:
In 53e2d39 I made a couple of further edits:
* adding the optional word "to" before "distribute"
* changing the name of one of the alt-text tag names so that it didn't duplicate the earlier one

After the tests pass, I'll merge. Thank you @xsuchy!

Issue #2102, opened 2023-08-30
Enhance Libtool-exception to not be specific to GPL only


xsuchy:
### How license meets inclusion principles

Presented in Fedora (and likely other Linuxes) Found during Fedora license review https://gitlab.com/fedora/legal/fedora-license-data/-/issues/312

### License Name

Libtool-exception

### Suggested short identifier

Libtool-exception

### URL to license text

https://git.savannah.gnu.org/cgit/libtool.git/tree/libltdl/lt__alloc.c#n15

### OSI Status

I don't know

### License author or steward

Unknown

### URL to project(s) that use license

https://git.savannah.gnu.org/cgit/libtool.git/tree

xsuchy:
The `exceptions/Libtool-exception.xml` has this leading line:

`As a special exception to the GNU General Public License, if you ...`

while the license in https://git.savannah.gnu.org/cgit/libtool.git/tree/libltdl/lt__alloc.c#n15 there is

`As a special exception to the GNU Lesser General Public License, if you ...`

I suggest changing the Libtool-exception to allow any text instead of `GNU General Public License` so it can be an exception to any license.

jlovejoy:
this is fine for adding markup - but I'd add "Lesser" as optional.  I'll comment on your PR

PR #2103, opened 2023-08-30
allow Libtool-exception to be exception to any license

xsuchy:
Fixes: #2102


jlovejoy:
@xsuchy - I'd just add `<optional>Lesser</optional>` as that is what we've done in other instances like this (and I'm not sure this exception is really appropriate for any license beyond the L/GPL family)

xsuchy:
Updated.

jlovejoy:
hmmm... looks like this got updated with the MPEG license files??

PR #2110, opened 2023-09-06
Fix license URL of Ruby

hsbt:
Hello, I'm administrator of ruby-lang.org. I fixed distribution URL of Ruby's license.

### DIFF
--- a/src/Ruby.xml
+++ b/src/Ruby.xml
@@ -2,7 +2,7 @@
 <SPDXLicenseCollection xmlns="http://www.spdx.org/license">
    <license isOsiApproved="false" licenseId="Ruby" name="Ruby License">
       <crossRefs>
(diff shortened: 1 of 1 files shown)


jlovejoy:
Thanks @hsbt !  We have a policy of keeping old (even if stale) URLs and simply adding new ones. Some people use the old ones as a way to track the history or something like that. See https://github.com/spdx/license-list-XML/blob/main/DOCS/license-fields.md part C

Would you mind updating this PR as such?

hsbt:
@jlovejoy I don't know how deprecate current URL and add new one. Can you please show me its format?

jlovejoy:
sure, simply - you can just leave the old link as it was and add a new ` <crossRef>` with the new URL. So, for your PR, just add back the old link in the first line

PR #2111, opened 2023-09-06
Update new-license-request.yaml

### LINKED ISSUE #1855: Adjust new license template to add field for license text
Just used the "new license request" template for #1854, I think it works well! But it would help to have a field (at least optional) to copy and paste the license text.

It's helpful to have the license text present in the issue itself, to save jumping between tabs and to make sure we're aligned on which text is actually being requested. And at least having it as an optional field would save the step of needing to submit the issue, then follow up with a comment with the text.

jlovejoy:
fixes #1855 

### DIFF
--- a/.github/ISSUE_TEMPLATE/new-license-request.yaml
+++ b/.github/ISSUE_TEMPLATE/new-license-request.yaml
@@ -8,7 +8,7 @@ body:
       value: |
         ### PLEASE NOTE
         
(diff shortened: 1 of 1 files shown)


swinslow (approved):
LGTM! We'll have to try it out and see whether the new template populates the way we expect, but this looks good to me from a first look.

swinslow (approved):

PR #2118, opened 2023-09-07
Create license-match.md

jlovejoy:
first commit to add guidance on determining match or not

### DIFF
--- a/DOCS/license-match.md
+++ b/DOCS/license-match.md
@@ -3,18 +3,23 @@
 As the SPDX License List grows, we need to rely on tooling more and more to check if a license found matches to a license already on the SPDX License List. This document provides information on various helpful tools and other guidance to help with this task.
 
 ## SPDX Check License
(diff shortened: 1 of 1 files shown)


swinslow:
Probably worth discussing further, but here are a few initial thoughts:
* I'd suggest we include links to any tools that we're recommending people to use as well as the matching guidelines themselves
* Is it worth also suggesting that people run a direct comparison themselves for greater certainty, using something like LibreOffice / Word?
* Similarly, might suggest using web searches for `"text to match" site:spdx.org` or something similar

jlovejoy:
@swinslow - yup, I need to add stuff. I took some notes in a Google doc while I was reviewing the various HPND variants, etc. Need to go back and add generalized stuff from that here. Will tag you when I do for review!

swinslow:
Thanks @jlovejoy! To avoid merging accidentally I switched this to "draft," we can change it back to a regular PR once it's ready for review.

jlovejoy:
yup, sounds good. I want to add to this while it's fresh in my mind, but not critical for the next couple days!

PR #2131, opened 2023-09-10
Small corrections in NPL-1.0 text

nicoroeser:
Since its addition to this repository (in commit d0d11d311f2e783615d162b0538f0213e9493dff), the text of `src/NPL-1.0.xml` has had some minor mistakes when compared to the URI in the cross reference (see the [upstream license text](https://web.archive.org/web/20160304040122/http://www.mozilla.org/MPL/NPL/1.0/) at the time when the license was imported).

This pull request fixes these mistakes. See the individual commits for details.


jlovejoy:
Thanks!

Issue #2081, opened 2023-08-17
New license request: BSD-3-Sun [SPDX-Online-Tools]


mkoncek:
**1.** License Name: BSD-3-Oracle
**2.** Short identifier: BSD-3-Oracle
**3.** License Author or steward: Oracle
**4.** Comments: This license test almost matches  BSD-3-Clause-No-Nuclear-Warranty but still differs slightly and does not contain the section about nuclear facilities, it has been found in the msv project (https://github.com/xmlark/msv) which is to be packaged in Fedora.
**5.** License Request Url: http://tools.spdx.org/app/license_requests/273
**6.** URL(s): 
**7.** OSI Status: Unknown
**8.** Example Projects: 
**8.** License Text Diff: https://github.com/spdx/licenseRequestImages/blob/master/b4089cbb-13c3-4b20-8f10-28cedc8ca28e.png

**Note:**
The license closely matched with the following license ID(s): BSD-3-Clause-No-Nuclear-Warranty

jlovejoy:
As mentioned, this is a very close match to https://spdx.org/licenses/BSD-3-Clause-No-Nuclear-License.html except that it omits the nuclear warranty at the end and has the following variations in the disclaimer section:
- adds "or liabilities" in the third sentence: "shall not be liable for damages _or liabilities_" 
- and has "AS A RESULT OF OR  RELATING TO USE, MODIFICATION OR DISTRIBUTION OF THE SOFTWARE OR ITS DERIVATIVES" instead of "AS A RESULT OF USING, MODIFYING OR DISTRIBUTING THIS SOFTWARE OR ITS DERIVATIVES"

jlovejoy:
I suspect we probably need to add this as its own license. I'd opt for using "Sun" rather than "Oracle" in the name and id since that is used in the license and where it originally came from

richardfontana:
+1 for using 'Sun' in the name.

bsdimp:
+1 for sun name, and needing a new ID for this. I think it meets the inclusion criteria.

mkoncek:
I do not have a preference for the name, I can change the proposal to ...-Sun.
Also, see slight variations of the license in files:
* https://github.com/xmlark/msv/blob/b9316e2f2270bc1606952ea4939ec87fbba157f3/xsdlib/src/main/java/com/sun/msv/datatype/regexp/InternalImpl.java
[...]

PR #2133, opened 2023-09-11
Add BSD-3-Clause-Sun license

mkoncek:
Related: #2081
Notes:
* **TODO** at `crossRefs`, I didn't know what to fill in
* `copyrightText / list / item number 3`, I am unsure if it qualifies as an item with an empty bullet or if the text belongs to the following `<p>`
* I used name **BSD 3-Clause Sun Microsystems** instead of **BSD 3-Clause Sun** because I think the word Sun itself is too vague.

### DIFF
--- a/src/BSD-3-Clause-Sun.xml
+++ b/src/BSD-3-Clause-Sun.xml
@@ -1,0 +1,55 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<SPDXLicenseCollection xmlns="http://www.spdx.org/license">
+  <license isOsiApproved="false" licenseId="BSD-3-Clause-Sun" name="BSD 3-Clause Sun Microsystems">
+    <crossRefs>
+    <crossRef>https://github.com/xmlark/msv/blob/b9316e2f2270bc1606952ea4939ec87fbba157f3/xsdlib/src/main/java/com/sun/msv/datatype/regexp/InternalImpl.java</crossRef>
+    </crossRefs>
+    <notes>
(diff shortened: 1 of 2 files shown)


jlovejoy:
thanks @mkoncek !

For crossref - use https://github.com/xmlark/msv/blob/b9316e2f2270bc1606952ea4939ec87fbba157f3/xsdlib/src/main/java/com/sun/msv/datatype/regexp/InternalImpl.java

As for the third clause, I'd just end the <list> after the second clause, since that will reflect how it is in the file above

name update for full name sounds good

mkoncek:
> As for the third clause, I'd just end the after the second clause, since that will reflect how it is in the file above

Well, in that case, isn't it actually a 2-Clause BSD-style license?
I mean is the 2- vs. 3-Clause name related to how many clauses the license text has or how similar the license is to the pure 2- and 3-Clause licenses?
The license has 2 clauses but the text is more similar to the 3-Clause version.

mkoncek:
Rebased:
* set `isOsiApproved="false"`
* added `crossRef`
* moved the 3rd list item into a separate paragraph

jlovejoy:
> > As for the third clause, I'd just end the after the second clause, since that will reflect how it is in the file above
> 
> Well, in that case, isn't it actually a 2-Clause BSD-style license? I mean is the 2- vs. 3-Clause name related to how many clauses the license text has or how similar the license is to the pure 2- and 3-Clause licenses? The license has 2 clauses but the text is more similar to the 3-Clause version.

well, it's not really just formatting that determines number of clauses. :)

Issue #2115, opened 2023-09-06
New license request: Cronyx (was X11-Fonts) [SPDX-Online-Tools]


whot:
**1.** License Name: X11-Fonts
**2.** Short identifier: X11-Fonts
**3.** License Author or steward: Unknown
**4.** Comments: Packages using this license are used in Fedora and this particular license was found during license review. See the fedora-license-data issue [1] for details. 

https://gitlab.com/fedora/legal/fedora-license-data/-/issues/321#note_1544774036
**5.** License Request Url: http://tools.spdx.org/app/license_requests/285
**6.** URL(s): https://gitlab.freedesktop.org/xorg/font/alias/-/blob/master/COPYING, https://gitlab.freedesktop.org/xorg/font/cronyx-cyrillic/-/blob/master/COPYING, https://gitlab.freedesktop.org/xorg/font/misc-cyrillic/-/blob/master/COPYING, https://gitlab.freedesktop.org/xorg/font/screen-cyrillic/-/blob/master/COPYING
**7.** OSI Status: Not Submitted
**8.** Example Projects: https://gitlab.freedesktop.org/xorg/font/alias/, https://gitlab.freedesktop.org/xorg/font/cronyx-cyrillic/, https://gitlab.freedesktop.org/xorg/font/misc-cyrillic/, https://gitlab.freedesktop.org/xorg/font/screen-cyrillic/

whot:
I should note - coming up with a better name would be much appreciated.

awilfox:
I'm thinking this one could be named Cronyx Font License, since the origin of the license seems to be Cronyx and I'm only aware of these fonts using it.  Short identifier could be Cronyx-Font.

swinslow:
Among the four examples given, I see a couple of minor differences: 
* alias and cronyx-cyrillic versions are identical
* misc-cyrillic does not include "above" before "copyright notice"
* screen-cyrillic does not have a comma after "distributed" in the initial clause

I think all of the above are non-substantive changes and can be handled via markup in the XML.

I'm +1 to add this as a simple, historical and broadly permissive license, used in a major distro (Fedora) and in line with the license inclusion principles.

swinslow:
I don't know the background of these licenses, but given @awilfox's comment that Cronyx is the originator of this one, I'd be okay using `Cronyx-Font` as the ID and "Cronyx Font License" as the name.

[...]

PR #2148, opened 2023-09-18
new license: Cronyx

xsuchy:
Fixes: #2115


jlovejoy:
@xsuchy - I would make the copyright notice the same in both files (even though it's not matchable, but for visual matching)

@swinslow - I noticed that two of the links for this have line b/w copyright and license text with a version number - do you think we should add that as optional for the sake of scanning tools?
see:
https://gitlab.freedesktop.org/xorg/font/screen-cyrillic/-/blob/master/COPYING 
https://gitlab.freedesktop.org/xorg/font/cronyx-cyrillic/-/blob/master/COPYING

but not in https://gitlab.freedesktop.org/xorg/font/alias/-/blob/master/COPYING

PR #2172, opened 2023-09-28
Add markup to GCC-exception-2.0.xml

### LINKED ISSUE #2055: New license request: glibc startup code license (LGPL with old GCC exception)
### How license meets inclusion principles

I believe this is actually a variant of *LGPL-2.1-or-later WITH GCC-exception-2.0*, but some folks believe that the exception text in SPDX has missing alternatives before it can match.

```
   Copyright (C) 1997-2023 Free Software Foundation, Inc.
   This file is part of the GNU C Library.

   The GNU C Library is free software; you can redistribute it and/or
   modify it under the terms of the GNU Lesser General Public
   License as published by the Free Software Foundation; either
   version 2.1 of the License, or (at your option) any later version.

   In addition to the permissions in the GNU Lesser General Public
   License, the Free Software Foundation gives you unlimited
   permission to link the compiled version of this file with other
   programs, and to distribute those programs without any restriction
   coming from the use of this file.  (The Lesser General Public
   License restrictions do apply in other respects; for example, they
   cover modification of the file, and distribution when not linked
   into another program.)

   The GNU C Library is distributed in the hope that it will be useful,
   but WITHOUT ANY WARRANTY; without even the implied warranty of
   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
   Lesser General Public License for more details.

   You should have received a copy of the GNU Lesser General Public
   License along with the GNU C Library; if not, see
   <https://www.gnu.org/licenses/>.
```

There is a variant of this license which includes an additional paragraph:

```
   Copyright (C) 2001-2023 Free Software Foundation, Inc.
   This file is part of the GNU C Library.

   The GNU C Library is free software; you can redistribute it and/or
   modify it under the terms of the GNU Lesser General Public
   License as published by the Free Software Foundation; either
   version 2.1 of the License, or (at your option) any later version.

   In addition to the permissions in the GNU Lesser General Public
   License, the Free Software Foundation gives you unlimited
   permission to link the compiled version of this file with other
   programs, and to distribute those programs without any restriction
   coming from the use of this file. (The GNU Lesser General Public
   License restrictions do apply in other respects; for example, they
   cover modification of the file, and distribution when not linked
   into another program.)

   Note that people who make modified versions of this file are not
   obligated to grant this special exception for their modified
   versions; it is their choice whether to do so. The GNU Lesser
   General Public License gives permission to release a modified
   version without this exception; this exception also makes it
   possible to release a modified version which carries forward this
   exception.

   The GNU C Library is distributed in the hope that it will be useful,
   but WITHOUT ANY WARRANTY; without even the implied warranty of
   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
   Lesser General Public License for more details.

   You should have received a copy of the GNU Lesser General Public
   License along with the GNU C Library; if not, see
   <https://www.gnu.org/licenses/>.
```

This raises further concerns whether this is still *LGPL-2.1-or-later WITH GCC-exception-2.0* (although if that's still a match, this is still fine with me of course).

Downstream discussion: https://lists.fedoraproject.org/archives/list/legal@lists.fedoraproject.org/message/VKN5IU3HOB2LZCLWHAU7CNTOLLYRGLES/

### License Name

LGPL-2.1-or-later WITH GCC-exception-2.0

LGPL-2.1-or-later WITH GCC-exception-2.0

### URL to license text

https://sourceware.org/git/?p=glibc.git;a=blob;f=csu/abi-note.c;h=c2ec208e94fbe91f63d3c375bd254b884695d190;hb=HEAD

### OSI Status

Approved

### License author or steward

FSF

https://sourceware.org/git/gitweb.cgi?p=glibc.git

jlovejoy:
fixes #2055 

needs a bit more work for first set of markup:  `<alt match="Library|Lesser"</alt>` really needs to be Library or Lesser or nothing. @swinslow is there a way to do that?

which is also why it's not matching to the .txt file

### DIFF
--- a/src/exceptions/GCC-exception-2.0.xml
+++ b/src/exceptions/GCC-exception-2.0.xml
@@ -7,11 +7,11 @@
       <notes>Typically used with GPL-2.0-or-later.  Sometimes also referred to as "linking exception."</notes>
     <text>
       <p>In addition to the permissions in the GNU
-        <optional>Library</optional> General Public License, the <alt match=".+" name="authors">authors</alt>
+       <alt match="Library|Lesser</alt> General Public License, the <alt match=".+" name="authors">authors</alt>
         <alt match="give(s)?" name="gives">give</alt> you unlimited permission to link the compiled version of this
         <alt name="file" match="file|library">file</alt><optional> into combinations</optional> with other programs,
         and to distribute those <alt name="programs" match="combinations|programs">programs</alt> without any restriction
-        coming from the use of this file.  (The <alt name="general" match="General|Library">General</alt> Public License
+        coming from the use of this file.  (The <alt name="general" match="General|Library|GNU Lesser">General</alt> Public License
         restrictions do apply in other respects; for example, they cover modification of the file, and distribution
         when not linked into
         <alt name="anotherProgram" match="a combined? executable|another program">another program</alt>.)</p>
--- a/src/exceptions/GCC-exception-2.0.xml
+++ b/src/exceptions/GCC-exception-2.0.xml
@@ -8,7 +8,7 @@
       <notes>Typically used with GPL-2.0-or-later.  Sometimes also referred to as "linking exception."</notes>
     <text>
       <p>In addition to the permissions in the GNU
-       <alt match="Library|Lesser"</alt> General Public License, the <alt match=".+" name="authors">authors</alt>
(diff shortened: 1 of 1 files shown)


swinslow:
@jlovejoy Sure thing, fixed in 726ac7c (also a couple XML format fixes)

This appears to work by including another match option for `()` -- e.g. `match="Lesser|Library|()"`

PR #2218, opened 2023-10-21
BSD-3-Clause: Match IETF variant

l1k:
Per section 4.c. of the IETF Trust Legal Provisions, "Code Components" in IETF Documents are licensed on the terms of the BSD-3-Clause license:

https://trustee.ietf.org/documents/trust-legal-provisions/tlp-5/

However @richardfontana points out that the IETF's license text "does not actually match SPDX `BSD-3-Clause` because of one additional word in clause 3 (`specific` before `contributors`)":

https://lore.kernel.org/all/CAC1cPGx-cb+YZ9KgEFvSjtf+fp9Dhcn4sm9qHmFFDRDxb=7fHg@mail.gmail.com/

Per his request, modify SPDX's definition of `BSD-3-Clause` to match the IETF variant.


bsdimp:
Seems same.. both the idea and the implementation. It might be better to use the ? Operator to make "specific" optional for each of the alternatives.

l1k:
Like this?

Note that in addition to the word "specific", the IETF has also added a comma.

jlovejoy:
thanks @l1k - is there an issue that goes with this PR?

so, looks like the tests are passing the way you did it, but I'm wondering if it'd be cleaner to simply make this another entire phrase option, instead of adding variables within a variable for the phrase (not sure if I explained that well... )

@goneall - do you have any guidance on this?

l1k:
@jlovejoy 

> is there an issue that goes with this PR?

Yes, the one you created last week: https://github.com/spdx/license-list-XML/issues/2242

> I'm wondering if it'd be cleaner to simply make this another entire phrase option, instead of adding variables within a variable for the phrase

Yes, that was my initial proposal, pushed as commit https://github.com/spdx/license-list-XML/commit/d64a28ddc1bfe272db85ca1afbaf7f73531ca735, but @bsdimp suggested above to amend an existing phrase rather than add a new one because they're essentially the same. I force-pushed that approach as commit https://github.com/spdx/license-list-XML/pull/2218/commits/53a56dba6b13cf0cfdefb64c65442db2dbe1e4d6, but if you prefer the original approach (or another one entirely), I will be happy to amend the commit and force-push again. Just let me know what the preferred approach is, I don't have a strong opinion either way. Thanks!

jlovejoy:
so, I was going to merge this, but now it has conflicts and I'm not sure how to resolve that... @l1k - do you know how to solve that?

l1k:
@jlovejoy

> so, I was going to merge this, but now it has conflicts and I'm not sure how to resolve that... @l1k - do you know how to solve that?

Yes, the submitter of the pull request (me in this case) can rebase their local branch on the main branch (e.g. `git pull --rebase`), locally fix up the merge conflicts and then force-push to the branch from which the pull request is sent (e.g. `git push l1k +HEAD:refs/heads/BSD-3-Clause-IETF` in my case, "l1k" is the name of my repo previously added with `git remote`). GitHub then automatically updates the pull request.

So I just did that and the merge conflict is gone. :)

jlovejoy:
thanks @l1k !

@goneall - I don't think we've had this kind of markup within markup? ok?

goneall:
> don't think we've had this kind of markup within markup? ok?

The regex looks good to me.

PR #2221, opened 2023-10-22
Revise FSFUL.xml to accommodate config.status variant

### LINKED ISSUE #2219: Revise FSFUL.xml to accommodate "config.status script" variant
There are what I believe are generated Automake files that insert a license that is mostly identical to `FSFUL` except the name of the referenced file is different. In the example I am currently looking at, it's

```
Copyright (C) 2021 Free Software Foundation, Inc.
This config.status script is free software; the Free Software Foundation
gives unlimited permission to copy, distribute and modify it.
```

although I think there may be other examples using other file names. 

You can verify this by doing a search for "This config.status script is free software".

richardfontana:
Fixes #2219

### DIFF
--- a/src/FSFUL.xml
+++ b/src/FSFUL.xml
@@ -9,7 +9,7 @@
          <p>Copyright (C) 1992-1996, 1998-2012 Free Software Foundation, Inc.</p>
       </copyrightText>
 
-      <p>This <alt name="file" match="config.status|configure">configure</alt> script is free software; the Free Software Foundation gives unlimited permission to copy,
+      <p>This <alt name="file" match=".+">configure</alt> script is free software; the Free Software Foundation gives unlimited permission to copy,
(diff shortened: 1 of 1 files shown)


richardfontana:
Not sure if I did this right. :)

jlovejoy:
This markup means it can only be "config.status" or "configure" - we could also denote this by replaceable text without the specific options.  But I think this is fine as it's the only two variations we've come across and if we came across another, then we can determine how to adjust at that time

jlovejoy:
FYI @swinslow  for awareness

richardfontana:
I think it should be replaceable text - I just found another variant, using the file name `config.lt`.

jlovejoy:
@richardfontana - can you change this to add the other variant you found or just change it to `<alt match=".+"`

PR #2227, opened 2023-10-23
Update CONTRIBUTING.md

jlovejoy:
fix a couple links that were outdated, no corresponding issue

@swinslow can you check and merge?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -4,9 +4,9 @@ Please familiarize yourself with the SPDX License List and its supporting docume
 
 * [Explanation of the fields contained on the list](DOCS/license-fields.md)
 
-* [SPDX License List Matching Guidelines]([https://spdx.org/spdx-license-list/matching-guidelines](https://spdx.github.io/spdx-spec/v2.3/license-matching-guidelines-and-templates/)) provides guidelines to be used for the purposes of matching licenses and license exceptions against those included on the SPDX License List.
+* [SPDX License List Matching Guidelines](https://spdx.github.io/spdx-spec/v2.3/license-matching-guidelines-and-templates/) provides guidelines to be used for the purposes of matching licenses and license exceptions against those included on the SPDX License List.
 
-* [SPDX Specification]([https://spdx.org/specifications](https://spdx.dev/use/specifications/)): It is helpful to be familiar with certain sections of the SPDX Specification that use or deal with the SPDX License List. In particular: sub-sections related to license information in Section 7, 8, and 10; Appendices B, D, and E.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -4,15 +4,15 @@ Please familiarize yourself with the SPDX License List and its supporting docume
...
 
-* [SPDX License List Matching Guidelines](https://spdx.org/spdx-license-list/matching-guidelines) provides guidelines to be used for the purposes of matching licenses and license exceptions against those included on the SPDX License List.
+* [SPDX License List Matching Guidelines]([https://spdx.org/spdx-license-list/matching-guidelines](https://spdx.github.io/spdx-spec/v2.3/license-matching-guidelines-and-templates/)) provides guidelines to be used for the purposes of matching licenses and license exceptions against those included on the SPDX License List.
 
-* [SPDX Specification](https://spdx.org/specifications): It is helpful to be familiar with certain sections of the SPDX Specification that use or deal with the SPDX License List. In particular: sub-sections related to license information in Section 7, 8, and 10; Appendices B, D, and E.
+* [SPDX Specification]([https://spdx.org/specifications](https://spdx.dev/use/specifications/)): It is helpful to be familiar with certain sections of the SPDX Specification that use or deal with the SPDX License List. In particular: sub-sections related to license information in Section 7, 8, and 10; Appendices B, D, and E.
```
xsuchy, line 9: This does not make a sense. You added link to a link. This part should be reverted.

CONTRIBUTING.md
```diff
@@ -4,15 +4,15 @@ Please familiarize yourself with the SPDX License List and its supporting docume
 
 * [Explanation of the fields contained on the list](DOCS/license-fields.md)
 
-* [SPDX License List Matching Guidelines](https://spdx.org/spdx-license-list/matching-guidelines) provides guidelines to be used for the purposes of matching licenses and license exceptions against those included on the SPDX License List.
+* [SPDX License List Matching Guidelines]([https://spdx.org/spdx-license-list/matching-guidelines](https://spdx.github.io/spdx-spec/v2.3/license-matching-guidelines-and-templates/)) provides guidelines to be used for the purposes of matching licenses and license exceptions against those included on the SPDX License List.
```
xsuchy, line 7: You added link to link. That does not have sense. The old link is dead so it should be updated - that is correct. But if we keep the version we will have to update it every second year. You can use https://spdx.github.io/spdx-spec/latest/license-matching-guidelines-and-templates/  which redirect to latest tag. Right now to v2.3. That should work in future without need for a change.

CONTRIBUTING.md
jlovejoy: fixed, thanks!

CONTRIBUTING.md
jlovejoy: fixed now

swinslow (commented):
Reviewing after the fact, but LGTM

Issue #2169, opened 2023-09-26
New license request: BSD-Systemics-W3Works


xsuchy:
**1.** License Name: W3Works License
**2.** Short identifier: W3Works
**3.** License Author or steward: Unknow
**4.** Comments: This license is included in perl-Crypt-Blowfish. It was discovered during Fedora License review https://gitlab.com/fedora/legal/fedora-license-data/-/issues/349
**5.** License Request Url: http://tools.spdx.org/app/license_requests/300
**6.** URL(s): https://metacpan.org/release/DPARIS/Crypt-Blowfish-2.14/source/COPYRIGHT#L7
**7.** OSI Status: Unknown
**8.** Example Projects: https://metacpan.org/release/DPARIS/Crypt-Blowfish-2.14/view/Blowfish.pm

xsuchy:
<details>
<summary>Text of the license</summary>
<pre>
Modifications and Current Implimentation Copyright (C) 2000 W3Works, LLC.
 
All rights reserved.
 
Current implimentation contains modifications made by W3Works, LLC.  The 
modifications remain copyright of W3Works, LLC and attribution for these 
modification should be made to W3Works, LLC.  These modifications and 
this copyright must remain with this package.
 
Additions to the Restrictions set out below are:
1. All advertising materials mentioning features or use of this software
   must display the following acknowledgement:
   This product inculdes software developed by W3Works, LLC (http://www.w3works.com)
 
   NO ADDITIONAL WARRANTIES, EXPRESS OR IMPLIED, INCLUDING, BUT NOT LIMITED TO,
   THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
   ARE EXTENDED BY THIS DISTRIBUTION.
 
   Any subsequent derrivations of this package must retainl this copyright.
</pre>
</details>

Note that there is several typos in the original license. "derrivations", "retail", "implementation",  "acknowledgement", and  "includes".

Pizza-Ria:
This is more of a rider than a license unto itself - I wonder if we could create another category similar to the exceptions that we have been approving recently.  Otherwise, I think we would have to include both the original (as depicted in https://metacpan.org/release/DPARIS/Crypt-Blowfish-2.14/source/COPYRIGHT#L7) and the rider to have a complete license.

Pizza-Ria:
I'd also vote for fixing the spelling errors ... unless that would cause an issue with the matching tool.

[...]

PR #2229, opened 2023-10-26
new license: BSD-Systemics-W3Works

xsuchy:
Fixes: #2169

### DIFF
--- a/src/BSD-Systemics-W3Works.xml
+++ b/src/BSD-Systemics-W3Works.xml
@@ -21,21 +21,17 @@
         </p>
                     </copyrightText>
         <p>
-                Current <alt name="implimentation" match="implimentation|implementation">implimentation</alt> contains modifications made by
-                <alt name="author" match=".*">W3Works, LLC</alt>.  The
-                modifications remain copyright of <alt name="author" match=".*">W3Works, LLC</alt>
-                and attribution for these
-                modification should be made to
-                <alt name="author" match=".*">W3Works, LLC</alt>.  These modifications and
+                Current <alt name="implimentation" match="implimentation|implementation">implimentation</alt> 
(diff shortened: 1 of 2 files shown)


xsuchy:
I believe that the failed check is not because of the content of the PR, but because failed CI script.

swinslow:
Thanks @xsuchy -- very strange. I agree this looks like a script issue. It's saying that [BSD-Systemics](https://spdx.org/licenses/BSD-Systemics.html) is matching this one, though that seems very odd given that this clearly has additional text (as non-optional required text) so it shouldn't be matching.

I'm going to play around with it a bit more to see what's going on. I'm wondering if perhaps it's something to do with the `match=".*"` alt-tags matching more than they're supposed to. I'll take a closer look and will follow up.

jlovejoy:
was there another instance of this license found with a different name? I noticed that @xsuchy put markup around all the names, but we usually don't do that if it's a one-off license (or until we find another example). Not sure if that would mess up the tests?

xsuchy:
It is only one instance. I added there the alts because I got here one thumbs up. https://github.com/spdx/license-list-XML/issues/2169#issuecomment-1770122139

jlovejoy:
> It is only one instance. I added there the alts because I got here one thumbs up. [#2169 (comment)](https://github.com/spdx/license-list-XML/issues/2169#issuecomment-1770122139)

I think that was markup for spelling errors, not names. Just wondering if that in any way was contributing to failing tests.

xsuchy:
> spelling errors, not names

??? what names? I thought that I added <alt> for spelling errors only. Do I miss something?

jlovejoy:
well, removing the alt markup around the names worked so I'm merging!

PR #2245, opened 2023-11-06
Update link for MIT license on OSI website

lornajane:
I noticed that the page for MIT lists a broken link - the information does seem to exist on OSI's site though, at a new URL: https://opensource.org/license/mit/

Updated the MIT file to reflect this.

### DIFF
--- a/src/MIT.xml
+++ b/src/MIT.xml
@@ -2,7 +2,7 @@
 <SPDXLicenseCollection xmlns="http://www.spdx.org/license">
    <license isOsiApproved="true" licenseId="MIT" name="MIT License">
       <crossRefs>
-         <crossRef>https://opensource.org/licenses/MIT</crossRef>
+         <crossRef>https://opensource.org/license/mit/</crossRef>
       </crossRefs>
     <text>
       <titleText>


xsuchy (approved):
It is not broken per se as there is redirect in place, but yeah, the URL changed. +1 to this change.

richardfontana:
+1. The SPDX-legal folks may not like the fact that OSI is not using the corresponding SPDX identifier in the URL in this case, but I think that is a silly thing to want since SPDX license identifiers are not particularly designed to be used in URLs and the use in a URL undoubtedly does little to nothing to promote the adoption of SPDX identifiers. There is also the recurring problem of deviation between OSI published license texts and supposedly corresponding SPDX identifiers, and the fact that the things the OSI approves are conceptually different from the things SPDX assigns identifiers for.

zvr:
Yikes! It seems opensource.org changed the links to their license pages to lowercase...

Besides MIT, this affects 120 `crossref` entries in our LIcenseList!

goneall:
> There is also the recurring problem of deviation between OSI published license texts and supposedly corresponding SPDX identifiers, and the fact that the things the OSI approves are conceptually different from the things SPDX assigns identifiers for.

@richardfontana I've used the SPDX tools to compare the text and have been filing issues at https://github.com/OpenSourceOrg/licenses/issues to track any inconsistencies.  If and when the OpenSourceOrg license data is consistent, I plan merge https://github.com/spdx/LicenseListPublisher/pull/118 which will automatically set the OSI approved flag based on the OSI metadata when we publish the license list.

jlovejoy:
@goneall - I think you and @richardfontana are onto a bigger issue re: license text, but as to this PR, can we merge? Or is it a non-issue due to redirects or other such automation?

goneall:
> @goneall - I think you and @richardfontana are onto a bigger issue re: license text, but as to this PR, can we merge? Or is it a non-issue due to redirects or other such automation?

I think this is fine to merge.

swinslow:
FWIW, lots of existing code still uses the old URL, e.g.: https://github.com/search?q=https%3A%2F%2Fopensource.org%2Flicenses%2FMIT&type=code

Because of that, we've generally taken the position that we don't _replace_ old URLs (even if they go dead), but we will _add_ new URLs if they get changed. That way, the old URLs appearing in code can still be used as a pointer towards matching to this license.

Happily, it looks like OSI has fixed their site so that the old URL now redirects to the new one; https://opensource.org/licenses/MIT no longer returns a 404.

Given that, @jlovejoy, are you good if I modify this PR so that it adds the new URL but also retains the old one as well?

jlovejoy:
fine by me @swinslow - I was sort of surprised that the case made a difference in terms of matching to the old URL ;)

Issue #2260, opened 2023-11-28
New license request: MIT-veillard-variant [SPDX-Online-Tools]


xsuchy:
**1.** License Name: MIT Veillard Variant
**2.** Short identifier: MIT-veillard-variant
**3.** License Author or steward: Unknown
**4.** Comments: This license is used in libxml2 package and was found during the Fedora License review https://gitlab.com/fedora/legal/fedora-license-data/-/issues/411
**5.** License Request Url: http://tools.spdx.org/app/license_requests/321
**6.** URL(s): https://raw.githubusercontent.com/GNOME/libxml2/4c2e7c651f6c2f0d1a74f350cbda95f7df3e7017/hash.c, https://github.com/GNOME/libxml2/blob/master/dict.c
**7.** OSI Status: Unknown
**8.** Example Projects: https://github.com/GNOME/libxml2/

xsuchy:
Text of the license:
Copyright (C) 2003-2012 Daniel Veillard.
 
  Permission to use, copy, modify, and distribute this software for any
  purpose with or without fee is hereby granted, provided that the above
  copyright notice and this permission notice appear in all copies.
 
  THIS SOFTWARE IS PROVIDED ``AS IS'' AND WITHOUT ANY EXPRESS OR IMPLIED
  WARRANTIES, INCLUDING, WITHOUT LIMITATION, THE IMPLIED WARRANTIES OF
  MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE. THE AUTHORS AND
  CONTRIBUTORS ACCEPT NO RESPONSIBILITY IN ANY CONCEIVABLE MANNER.

xsuchy:
It is known to scancode https://scancode-licensedb.aboutcode.org/mit-veillard-variant Their note:
>  this is similar to the mit-taylor-variant but uses a different warranty disclaimer. This is found in the original versions of https://en.wikipedia.org/wiki/Libxml2

And the text is close to ISC.

jlovejoy:
this is the same grant as https://spdx.org/licenses/ISC.html but a different disclaimer paragraph.

+1 to add but I'm not sure we should use MIT in the name or id b/c it's closer to ISC than to MIT??

richardfontana:
I think `ISC-Veillard-variant` would be better than `MIT-Veillard-variant`, all else being equal.

jlovejoy:
[...]

PR #2281, opened 2023-12-21
new license: ISC-Veillard

xsuchy:
Fixes: #2260


jlovejoy:
a-ha! there are two "Ls" in Veillard, which was reflected in the id tag, but not the filenames... should pass now!

PR #2297, opened 2024-01-02
Add note to SWI-exception

jlovejoy:
Add similar note to this exception as we have already for GNU-compiler-exception.


swinslow (commented):
Reviewing after the fact, but generally LGTM. There's one typo ("softare") and I'd word it slightly differently, so I'm proposing a couple of tweaks in #2312

PR #2298, opened 2024-01-02
Add markup to GNU-compiler-exception.xml

### LINKED ISSUE #2152: Recognise a variant of the GNU-compiler-exception
The current exception text is:

            As a special exception, if you link this library with files
            compiled with a GNU compiler to produce an executable, this
            does not cause the resulting executable to be covered by
            the GNU General Public License. This exception does not
            however invalidate any other reasons why the executable
            file might be covered by the GNU General Public License.

I recently ran into a variant of this with minor changes and an additional sentence:

        As a special exception, if you link the code in this file with
        files compiled with a GNU compiler to produce an executable,
        that does not cause the resulting executable to be covered by
        the GNU Lesser General Public License.  This exception does not
        however invalidate any other reasons why the executable file
        might be covered by the GNU Lesser General Public License.
        This exception applies to code released by its copyright holders
        in files containing the exception.

This wording is used in several libio source files in glibc; e.g.:
https://sourceware.org/git?p=glibc.git;a=blob;f=libio/fputc.c;h=9ef6166dd6b5cf566c7bd591832c2b73d34e08db;hb=HEAD#l18

Can the GNU-compiler-exception XML be enhanced to match this version as well?

jlovejoy:
Fixes #2152 

### DIFF
--- a/src/exceptions/GNU-compiler-exception.xml
+++ b/src/exceptions/GNU-compiler-exception.xml
@@ -10,12 +10,17 @@
       </notes>
       <text>
          <p>
-            As a special exception, if you link this library with files
-            compiled with a GNU compiler to produce an executable, this
+            As a special exception, if you link 
+            <alt name="library" match="this library|the code in this file">this library</alt> 
+            with files compiled with a GNU compiler to produce an executable, 
+            <alt name="this" match="this|that">this</alt>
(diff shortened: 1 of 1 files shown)


swinslow (approved):
LGTM

PR #2301, opened 2024-01-03
Update Apache-1.1.xml to match original

### LINKED ISSUE #2270: Apache-1.1 XML template does not match official license texts
The Apache-1.1 XML template does not match the following texts:

- https://www.apache.org/licenses/LICENSE-1.1
- https://opensource.org/license/apache-1-1/

The mismatch is:
XML template: `please see http://www.apache.org/.`
Other texts: `please see <http://www.apache.org/>.`

jlovejoy:
fixes #2270 


jlovejoy:
this PR fixes some holdovers from the prior, pre-XML way we stored the license texts and accommodated variations.

In sum, it should match: https://www.apache.org/licenses/LICENSE-1.1

As such, I have:
- update the title
-  removed the copyrightText tag, b/c this is one of those few examples where the copyright notice is (allegedly) to cover the license itself
- removed the holdover variable names text in []
- add the <> in the second to last paragraph around the URL

swinslow:
Thanks @jlovejoy! My only comments on this one is that, if we're really trying to match the upstream version, there appears to be one line near the end (just before "AS IS") where single-quotes were being used instead of backticks from the original.

I've added a couple of commits to fix this, and I'll merge if/when the tests pass afterwards. Thanks!

PR #2306, opened 2024-01-03
Update template-decision.md

jlovejoy:
making this simpler as to avoid less need for deleting stuff and faster to fill in

@swinslow  as you and I are most likely to use this, I don't think we need the explanatory text?


swinslow (approved):
+1, thanks @jlovejoy!

PR #2312, opened 2024-01-04
SWI-exception: Revise notes

swinslow:
As a follow-on from #2297, a couple of things here:

* typo fix: "softare" => "software"
* wording is slightly different between the two (beyond just the "free software" reference), so I opted for "closely similar" rather than "the same as"


swinslow:
@jlovejoy If you're good with the tweaks here, feel free to merge!

Issue #2315, opened 2024-01-11
New license request: OGL-Thailand-1.0 [SPDX-Online-Tools]


bact:
**1.** License Name: Thailand DGA Open Government License 1.0
**2.** Short identifier: OGL-Thailand-1.0
**3.** License Author or steward: Digital Government Development Agency
**4.** Comments: The license applied to open government data that Thai government made publicly available at https://data.go.th.
It doesn't match another license already on the SPDX License List.
Although the license doesn't target primarily source code, but as SPDX will include AI and Data Profiles soon, this data license is relevant.
The license has identifiable and stable text, in Thai and in English. It has an explicit version number.
**5.** License Request Url: http://tools.spdx.org/app/license_requests/333
**6.** URL(s): https://data.go.th/th/pages/dga-open-government-license, https://data.go.th/uploads/page_images/2021-06-11-053145.086775Terms-and-Conditions-for-DGA-Open-Government-LicenseTH04062021.pdf, https://data.go.th/uploads/page_images/2021-06-11-053225.908526Terms-and-Conditions-for-DGA-Open-Government-LicenseEN04062021.pdf
**7.** OSI Status: Not Submitted
**8.** Example Projects: https://github.com/covid19db/fetchers-python, https://www.tomtom.com/en_ie/thirdpartyproductterms/eula-202203/

bact:
I fixed one typo ("Terms and Conditions of the Term" -> "Terms and Conditions of the Terms") and fixed the list/item structure in the XML file here #2316

swinslow:
For reference, here is the text of the license as taken from the English language PDF at https://data.go.th/th/pages/dga-open-government-license:

= = = = =

```
Terms and Conditions for DGA Open Government License

Open Government Data Platform provided by Digital Government Development Agency
(Public Organization) (DGA), is broadenly facilitated the public and government organizations
by digital system. DGA supports government organizations to share or connect their data to
the public, openly published on the website at https://data.go.th, for the purposes of
transparently providing data search and access to the public as well as promoting their
involvement in and inspection on governmental activities and encouraging them to further
use the data to develop innovations and services, which would be benefitial to the country
in certain aspects.

The user will be able to use the data, which are contained on Open Government Data Platform
or thereunder, under the agreement to be bound by and comply with the following terms
and conditions for DGA Open Government License.

Definitions

“Terms” means the terms and conditions for DGA Open Government License

“Data” means information are presented as facts or others as the same meaning.
[...]

PR #2316, opened 2024-01-11
[OGL-Thailand-1.0] Fix list/item wrong parent/child

bact:
Fix list/item wrong parent/child


jlovejoy:
This license is still under review in issue #2315

Issue #2233, opened 2023-10-28
New license request: BSD-Source-Code-Justin-Gibbs


richardfontana:
### How license meets inclusion principles

This is similar to `BSD-Source-Code` but adds some problematic language ("immediately at the beginning of the file") to clause 1 reminiscent of `libutil-David-Nugent`. It seems to have originated in FreeBSD but has propagated to a number of other projects. For example, it occurs in the Linux kernel (though sadly it may have been replaced with an incorrect SPDX-License-Identifier by now) and in the Linux userspace package acl. 

### License Name

BSD Source Code Attribution Justin Gibbs Variant

### Suggested short identifier

BSD-Source-Code-Justin-Gibbs

### License or Exception?

license

### URL to license text

https://github.com/lattera/freebsd/blob/master/sys/cam/cam.c

### OSI Status

Not submitted

### License author or steward

Unknown

### URL to project(s) that use license

https://kernel.org
https://savannah.nongnu.org/projects/acl/

### paste text of license here

``` 
* Copyright (c) 1997 Justin T. Gibbs.
 * All rights reserved.
 *
 * Redistribution and use in source and binary forms, with or without
 * modification, are permitted provided that the following conditions
 * are met:
 * 1. Redistributions of source code must retain the above copyright
 *    notice, this list of conditions, and the following disclaimer,
 *    without modification, immediately at the beginning of the file.
 * 2. The name of the author may not be used to endorse or promote products
 *    derived from this software without specific prior written permission.
 *
 * THIS SOFTWARE IS PROVIDED BY THE AUTHOR AND CONTRIBUTORS ``AS IS'' AND
 * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
 * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
 * ARE DISCLAIMED. IN NO EVENT SHALL THE AUTHOR OR CONTRIBUTORS BE LIABLE FOR
 * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
 * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS
 * OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
[...]

PR #2329, opened 2024-01-14
add BSD-Source-beginning-file

xsuchy:
I kept there the <alt> variations from original BSD-Source-Code and slightly altered it to the actual example in the submission. Fixes: #2233

### DIFF
--- a/src/BSD-Source-beginning-file.xml
+++ b/src/BSD-Source-beginning-file.xml
@@ -1,0 +1,62 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<SPDXLicenseCollection xmlns="http://www.spdx.org/license">
+   <license isOsiApproved="false" licenseId="BSD-Source-beginning-file"
+            name="BSD Source Code Attribution - beginning of file variant"
+            listVersionAdded="3.23">
+      <crossRefs>
+         <crossRef>https://github.com/lattera/freebsd/blob/master/sys/cam/cam.c#L4</crossRef>
+      </crossRefs>
+      <notes>
+         This license is mostly the same as BSD-Source-Code but adds includes
+         a specific requirement requiring retention of the copyright notice
+         "immediately at the beginning of the file".
+      </notes>
+    <text>
+      <copyrightText>
+         <p>Copyright (c) 1997 Justin T. Gibbs.
+            <br/>All rights reserved.
+      </p>
+      </copyrightText>
+
+      <p>Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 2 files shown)


src/BSD-Source-beginning-file.xml
```diff
@@ -0,0 +1,62 @@
...
+      <notes>
+         This license is mostly the same as BSD-Source-Code but adds includes
+         a specific requirement requiring retention of the copyright notice
+         "immediately at the beginning of the file".
+      </notes>
+    <text>
+      <copyrightText>
+         <p>Copyright (c) 1997 Justin T. Gibbs.
+            <br/>All rights reserved.
+      </p>
```
bsdimp, line 18: This is just an example of the copyright text, right? Done we have generic year and holder text we try to use?

xsuchy: Right. This is for the sake of completeness in examples. But for matching, it is ignored. See https://spdx.github.io/spdx-spec/v2.3/license-matching-guidelines-and-templates/#b11-copyright-notice

jlovejoy: @bsdimp - we have made the copyright notice generic in some of the more "popular" license or when someone has asked to do so, but we have not made a hard rule on this. I think many of the licenses being found in Fedora at this point are quite old, so not sure it matters (and in this case, we probably would hope others don't re-use this license!)

bsdimp: > @bsdimp - we have made the copyright notice generic in some of the more "popular" license or when someone has asked to do so, but we have not made a hard rule on this. I think many of the licenses being found in Fedora at this point are quite old, so not sure it matters (and in this case, we probably would hope others don't re-use this license!)

OK. Works for me then. I'm otherwise happy with this. Totally agree with 'others not using it'. I've been talking with Justin and a few other people who have this license in FreeBSD to persuade them to switch to something more standard. They are open to the idea, but the devil is in the details.

bsdimp (approved):

swinslow (approved):
Thanks @xsuchy and others commenting in this thread.

I was debating whether to ask about the alt match `organizationClause2` at the end of the second clause. I didn't see that there was an equivalent sample for this license match that had additional language there.

However, given that it's 0-or-more -- and that it looks like this is meant to pick up the "written permission of..." bit from the related [BSD-Source-Code](https://github.com/spdx/license-list-XML/blob/main/src/BSD-Source-Code.xml) license -- I'm comfortable with us being a bit over-inclusive here, and I'll go ahead and merge this as-is.

Issue #2223, opened 2023-10-23
New license request: LPD-document [SPDX-Online-Tools]


richardfontana:
**1.** License Name: ScanCode calls it "Peter Deutsch Document License"
**2.** Short identifier: LPD-document
**3.** License Author or steward: L. Peter Deutsch, possibly
**4.** Comments: This is a license that seems to have originally appeared in RFCs in the 1990s. It should be understood as granting permissions for text documents comparable in scope to FOSS licenses. A minor variant of this license (adding one sentence that does not narrow its permissions) has been found covering files in one Fedora Linux package. Since this is likely the original version, I want to submit this one first.
**5.** License Request Url: http://tools.spdx.org/app/license_requests/314
**6.** URL(s): https://scancode-licensedb.aboutcode.org/peter-deutsch-document.html, https://www.ietf.org/rfc/rfc1952.txt
**7.** OSI Status: Not Submitted
**8.** Example Projects: https://www.ietf.org/rfc/rfc1952.txt (not a project, I know)

jlovejoy:
@richardfontana - to make sure I understand, this license has NOT been found in Fedora but sounds like you are submitting due to legacy in terms of having ben used prior?

swinslow:
Discussed on legal team call 2024-01-11, consensus to treat this as equivalent to https://github.com/spdx/license-list-XML/issues/2224 with the final sentence from that one being optional.

ID: LPD-document

Name: LPD Documentation License

PR #2330, opened 2024-01-14
add LPD-document

xsuchy:
Fixes: #2223


jlovejoy:
@swinslow - the tests are passing, but isn't (or shouldn't) the text within the optional tag supposed to be in the .txt file as well?

xsuchy:
I intentionally used the other license from crossref as test. I am willing to change if anyone has different preference.

swinslow (approved):
@jlovejoy I don't think there's a requirement (within the license validation test tooling) that optional text be included in the test .txt file. At least, it doesn't look like it's being required here...

The optional text will appear on the website based on being included and tagged in the XML file, not in the test text file. So I'm +1 to go ahead and merge this as-is.

PR #2337, opened 2024-01-19
Add fmt-exception

### LINKED ISSUE #2282: New exception request: No Attribution for Binary distributions
### How license meets inclusion principles

A simple and hopefully clear exception to a standard permissive license (used with MIT) that specifies that there is no need for reproducing copyright and license texts if the code is compiled and included in a binary artifact.

### License Name

No Attribution for Binary distributions

NoAttribBinary

### License or Exception?

exception

### URL to license text

https://github.com/fmtlib/fmt/blob/master/LICENSE

### OSI Status

Not Submitted

### License author or steward

Victor Zverovich

https://github.com/fmtlib/fmt

### paste text of license here

As an exception, if, as a result of your compiling your source code, portions of this Software are embedded into a machine-executable object form of such source code, you may redistribute such embedded portions in such object form without including the above copyright and permission notices.

zvr:
This adds a new exception, found in the license of `fmtlib`, as per #2282.

Notes:

- I assume the `listVersionAdded` attribute will be (automatically?) added in the next release of the License List, so I didn't add it.
- I made the heading line `<optional>`, since the exception text might appear without it. If there is disagreement, I can make it mandatory.

Closes #2282

### DIFF
--- a/src/exceptions/fmt-exception.xml
+++ b/src/exceptions/fmt-exception.xml
@@ -1,10 +1,13 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <SPDXLicenseCollection xmlns="http://www.spdx.org/license">
-   <exception licenseId="fmt-exception" name="fmt exception">
+   <exception licenseId="fmt-exception" name="fmt exception" listVersionAdded="3.23">
    <crossRefs>
       <crossRef>https://github.com/fmtlib/fmt/blob/master/LICENSE</crossRef>
       <crossRef>https://github.com/fmtlib/fmt/blob/2eb363297b24cd71a68ccfb20ff755430f17e60f/LICENSE#L22C1-L27C62</crossRef>
    </crossRefs>
+   <notes>
+      Typically used with MIT.
+  </notes>
    <text>
       <optional><p>--- Optional exception to the license ---</p></optional>
       <p>
--- a/src/exceptions/fmt-exception.xml
+++ b/src/exceptions/fmt-exception.xml
@@ -1,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<SPDXLicenseCollection xmlns="http://www.spdx.org/license">
+   <exception licenseId="fmt-exception" name="fmt exception">
+   <crossRefs>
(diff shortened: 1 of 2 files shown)


src/exceptions/fmt-exception.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<SPDXLicenseCollection xmlns="http://www.spdx.org/license">
+   <exception licenseId="fmt-exception" name="fmt exception">
+   <crossRefs>
+      <crossRef>https://github.com/fmtlib/fmt/blob/master/LICENSE</crossRef>
+      <crossRef>https://github.com/fmtlib/fmt/blob/2eb363297b24cd71a68ccfb20ff755430f17e60f/LICENSE#L22C1-L27C62</crossRef>
+   </crossRefs>
```
xsuchy, line 7: Here should be:
```
<notes>
  Typically used with MIT.
</notes>

File: src/exceptions/fmt-exception.xml [hunk 2]
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<SPDXLicenseCollection xmlns="http://www.spdx.org/license">
+   <exception licenseId="fmt-exception" name="fmt exception">
```
xsuchy, line 3: Please add this attribute: `listVersionAdded="3.23"`

src/exceptions/fmt-exception.xml
jlovejoy: that is correct, it was noted in the "license decision" in the issue

src/exceptions/fmt-exception.xml
jlovejoy: @zvr - the listVersionAdded tag is not added automatically when published (but interesting thought). The tag is included automatically if you use the online tool to generate the xml which serves as a nice reminder if one has used that tool to submit and make the xml files :)

jlovejoy:
thanks @zvr !! and thanks @xsuchy  for the review.
I'll just make the minor fixes and merge :)

PR #2391, opened 2024-02-15
BSD-3-Clause-No-Nuclear-License: Update the license URL

rilysh:
Remove the "AuthParam". It seems unnecessary to add as it will be generated each time a new user goes to http://download.oracle.com/otn-pub/java/licenses/bsd.txt. Also, with that AuthParam, I can't see the license (the site throws an "unauthorized request" error).

### DIFF
--- a/src/BSD-3-Clause-No-Nuclear-License.xml
+++ b/src/BSD-3-Clause-No-Nuclear-License.xml
@@ -3,7 +3,7 @@
    <license isOsiApproved="false" licenseId="BSD-3-Clause-No-Nuclear-License"
             name="BSD 3-Clause No Nuclear License">
       <crossRefs>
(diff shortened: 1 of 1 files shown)


jlovejoy:
generally we don't remove old URLs but add new ones. 

In this case, though, I'm wondering if the AuthParam was not supposed to be there to begin with. @goneall - do you have thoughts on this scenario? It's otherwise the same link

rilysh:
> generally we don't remove old URLs but add new ones.

This patch points to the same (old URL) but just without AuthParam.

> In this case, though, I'm wondering if the AuthParam was not supposed to be there to begin with. 

Yes, AuthParam was not supposed to be there, as it gets generated each time when a user visits the URL.

goneall:
I think this should be fine to merge - the auth section likely is not used for matching

swinslow (approved):
+1, this looks like it was included in error and I think it's fine to remove the AuthParam from the crossRef.

However, please note that this is one of the licenses that's picked up in #2398, so to avoid merge conflicts there I'm going to hold off on merging this one for the moment until after #2398 is merged.

Issue #2419, opened 2024-03-13
New license request: NCBI-PD [SPDX-Online-Tools]


Pandapip1:
**1.** License Name: NCBI Public Domain Notice
**2.** Short identifier: NCBI-PD
**3.** License Author or steward: Unknown
**4.** Comments: NOTE: This is a bit of an edge case, since it isn't actually a license.

All definitive factors are met.

Non-definitive factors:

1. Works released into the public domain under 17 U.S. Code § 105 arguably meet the FSF Free Software Definition.
2. This does not meet factor two, since this is, by definition, applicable only to US government works.
3. The United States government produces quite a lot of work, a lot of which is in the form of code. As such, it is likely to be encountered.
4. Not applicable (meets factor 1)
5. Not applicable (there is no license steward)
**5.** License Request Url: http://tools.spdx.org/app/license_requests/357
**6.** URL(s): https://www.govinfo.gov/content/pkg/USCODE-2010-title17/pdf/USCODE-2010-title17-chap1-sec105.pdf
**7.** OSI Status: Not Submitted
**8.** Example Projects: https://github.com/ncbi/sra-tools

karsten-klein:
**ScanCode**
reference id: [public-domain](https://github.com/nexB/scancode-toolkit/blob/develop/src/licensedcode/data/licenses/public-domain.LICENSE)
matched id: [ncbi](https://github.com/nexB/scancode-toolkit/blob/develop/src/licensedcode/data/licenses/ncbi.LICENSE)

**{metæffekt} Universe**
canonical name: [Public Domain](https://github.com/org-metaeffekt/metaeffekt-universe/blob/main/src/main/resources/ae-universe/[p]/README.md#user-content-Public-Domain:~:text=Public%20Domain)
short name: Public-Domain
Open CoDE similar license id: public-domain
markers: US Copyright Act Marker

**Comments**
- For me, this raises the question how we'd represent public domain content with an general SPDX license identifier. (Probably not desired)
- Since we have been modeling other public domain licenses in SPDX (e.g. NIST-PD) I would give it a +1 as NCBI-PD and not with focus on governmental work.

Pandapip1:
The trick is that in this context, "public domain" means something entirely different than what people expect. It means that US government works don't have copyright protection inside the jurisdiction of the United States. **However,** international copyright laws mean that the United States government is automatically granted a copyright to those works. Without an additional license, those works can't be legally used outside the United States.

[...]

PR #2433, opened 2024-03-28
Add NCBI-PD

Pandapip1:
Closes https://github.com/spdx/license-list-XML/issues/2419

### DIFF
--- a/src/NCBI-PD.xml
+++ b/src/NCBI-PD.xml
@@ -17,7 +17,7 @@
             the terms of the United States Copyright Act.  It was written as part of the 
             <alt match="authors'|author's" name="author1">author's</alt> official duties as <optional>a</optional>
             United States Government <alt match="employees|employee" name="employee">employees</alt> and 
-            <alt match="thus cannot be copyrighted|is not subject to copyright in the United States \(17 USC 105\)" name="nocopyright">thus cannot be copyrighted</alt>.  
+            thus cannot be copyrighted.  
             This software<optional>/database</optional> is freely available to the public for use. 
             The <alt match="NLM|National Library of Medicine|National Library of Medicine \(NLM\)" name="nlm">NLM</alt> 
             and the U.S. Government have not placed any restriction on its use or reproduction.</p>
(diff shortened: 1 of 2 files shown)


Pandapip1:
> First please wait till the issue is approved.

Moving PR to draft.

Pandapip1:
Requesting @swinslow for review and re-labeling

Pandapip1:
(I found a few more variations of the text, ergo the commits)

jlovejoy:
@swinslow - can you review? I think some of the markup we usually do a bit differently

Pandapip1:
I can't figure out why it's failing. Any tips?

xsuchy:
> Normal text of license does not match starting at line #1 column #5 "software" when comparing to template text "PUBLIC DOMAIN NOTICE

You are missing the tile text in the test file.

Pandapip1:
Added the title; it's still failing with what appears to be the same error. If you'd like to take over this PR, I'd totally understand. Otherwise, got any ideas as to why it's failing? Everything seems to be correct, but for some reason the regexes aren't matching correctly.

jlovejoy:
I'm having a go at updating this in small increments. Also adding some line breaks to make it easier to "see" what is going on.

jlovejoy:
ok, so there were a bunch of optional tags that shouldn't be there. Also, I don't see any variation in the [Github search](https://github.com/search?q=%22PUBLIC+DOMAIN+NOTICE%22+%22National+Center+for+Biotechnology+Information%22&type=code&p=1) that @swinslow did that has "created" instead of "written" - we only add markup for variations actually found (and sparingly). But if I missed something, let me know!

Also, given that all of these have the same name in them (National Center for Biotechnology Information / NLM) - I'd be inclined to not mark that up at this point.

We can always add markup if we find variations later

jlovejoy:
ok, last item to consider: the FAQ added at the end as optional - I only see this in one instance. I'd be inclined to just leave it out. It's not part of the license in any case and I think scanning tools can sort this one out

xsuchy (approved):
Technically speaking the alt matches can be simplified, but this way it is more readable. So +1 from me.

swinslow (approved):
LGTM; if needed, we can tweak the markup again to add additional variations for future releases. Thank you all!

Issue #2386, opened 2024-02-08
New license request: fftpack-2004 (NCL) [SPDX-Online-Tools]


nielsdg:
**1.** License Name: FFTPACK License 2004
**2.** Short identifier: fftpack-2004
**3.** License Author or steward: Unknown
**4.** Comments: This license was discovered during license review in Fedora Linux. https://gitlab.com/fedora/legal/fedora-license-data/-/issues/475. It is used in the package pipewire.

The license is known to scancode-toolkit: https://github.com/nexB/scancode-toolkit/blob/develop/src/licensedcode/data/licenses/fftpack-2004.LICENSE
**5.** License Request Url: http://tools.spdx.org/app/license_requests/349
**6.** URL(s): https://gitlab.freedesktop.org/pipewire/pipewire/-/blob/master/src/modules/module-filter-chain/pffft.c?ref_type=heads#L1-52
**7.** OSI Status: Unknown
**8.** Example Projects: https://pipewire.org/

jlovejoy:
this is very similar to BSD-3-Clause, except:
- it adds "use of the Software in source and binary forms" instead of just "use in source and binary forms" in the lead in
- the clauses are in a different order with what we normally consider the third clause appearing first
- the warranty of disclaimers has different wording: it adds a disclaimer of noninfringement, and then has less details as to exclusion of types of damages in the last part and also expands that to "use of other dealings" with the software

Given the above, I suspect we should add this as a new license

karsten-klein:
**ScanCode**
* [fftpack-2004](https://github.com/nexB/scancode-toolkit/blob/develop/src/licensedcode/data/licenses/fftpack-2004.LICENSE)

**{metæffekt} Universe**
* [Spherepack License](https://github.com/org-metaeffekt/metaeffekt-universe/blob/main/src/main/resources/ae-universe/[s]/README.md#:~:text=Spherepack%20License)

**Comments**
Seems we detected that licenses in another context (Spherepack); see also https://sources.debian.org/src/spherepack/3.3~a1-5/debian/copyright/

jlovejoy:
interesting, so the license in link for the submission seems to got to piperwire project, and then it calls it the "FFTPACK license" in the file. Has the same 2004 copyright notice as what @karsten-klein found in Debian seems to inicate the file is part of spherepack. 

Doing a Google search on a phrase, turns up a few other places, but notably: https://www.ncl.ucar.edu/Download/NCL_source_license.shtml

[...]

PR #2434, opened 2024-03-28
add NCL license

xsuchy:
Fixes: #2386

### DIFF
--- a/src/NCL.xml
+++ b/src/NCL.xml
@@ -1,0 +1,63 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<SPDXLicenseCollection xmlns="http://www.spdx.org/license">
+   <license isOsiApproved="false" licenseId="NCL"
+   name="NCL Source Code License" listVersionAdded="3.24">
+      <crossRefs>
+         <crossRef>https://gitlab.freedesktop.org/pipewire/pipewire/-/blob/master/src/modules/module-filter-chain/pffft.c?ref_type=heads#L1-52</crossRef>
+      </crossRefs>
+      <text>
+         <copyrightText>
+            <p>
+            Copyright (c) 2004 the University Corporation for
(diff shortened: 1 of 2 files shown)


src/NCL.xml
```diff
@@ -0,0 +1,63 @@
...
+         <copyrightText>
+            <p>
+            Copyright (c) 2004 the University Corporation for
+            Atmospheric Research ("UCAR"). All rights reserved.
+            Developed by NCAR's Computational and Information
+            Systems Laboratory, UCAR, www.cisl.ucar.edu.
+            </p>
+         </copyrightText>
+         <p>
+            Redistribution and use <alt name="software" match=".*">of the Software</alt> in source and
```
jlovejoy, line 18: the alt tag and match=.* here would mean "of the Software" could be replace with anything. I don't think we want that, right? The variation we have found has "of the Software" or that is simply omitted, so I think this should just be optional language? @swinslow - thoughts?

xsuchy: Technically .* match to 0 or more characters. So, an empty string is a match too.

jlovejoy: @goneall @swinslow - seems like this kind of think can be represented in two ways: alt tag, as @xsuchy has done, which is part of the regex standard; or with the optional tag, which is an SPDX xml special tag. I believe we have mostly done the latter for a case like this? It'd be good to be consistent. Thoughts? (and if then we can add something to the XML documentation)

xsuchy: I changed the `alt` tag to `optional` as I have no strong preference and to ease the review.

jlovejoy: actually, I realized this does matter b/c Optional means that this exact language can be there or be omitted. But if we use .* then any words can replace those in the alt tag. This is probably a good clarification to make in our XML documentation.

Issue #2348, opened 2024-01-25
New license request: BSD-2-Clause-Copyright [SPDX-Online-Tools]


jrisc:
**1.** License Name: BSD 2-Clause license with unmodified copyright notice
**2.** Short identifier: BSD-2-Clause-Copyright
**3.** License Author or steward: Unknown
**4.** Comments: This license is similar to the BSD 2-Clause "Simplified" License, but with an addition to the first clause. It also matches the first half of the FreeBSD Documentation License.

This license is used in Fedora, and was found during license review:
https://gitlab.com/fedora/legal/fedora-license-data/-/issues/461
**5.** License Request Url: http://tools.spdx.org/app/license_requests/341
**6.** URL(s): 
**7.** OSI Status: Not Submitted
**8.** Example Projects: 
**8.** License Text Diff: https://github.com/spdx/licenseRequestImages/blob/master/84b193b8-e27d-4b69-8099-b44aeb6d5705.png

**Note:**
The license closely matched with the following license ID(s): BSD-2-Clause

jlovejoy:
this is indeed a match to BSD-2-Clause, but adds the text in italics in the first clause

"Redistributions of source code must retain the above copyright notice, this list of conditions and the following disclaimer _as the first lines of this file unmodified_."

sigh... thoughts?

jlovejoy:
note: we have added other licenses with specifics as to where to put the notice as a new license

bsdimp:
I think this needs to be a new license. It's another instance of at start of file varaints of the BSD license.

jlovejoy:
agree @bsdimp 

Adding link to license text, as I realized they are not in this issue, but over in the Fedora Gitlab repo issue
https://github.com/krb5/krb5/blob/krb5-1.21.2-final/NOTICE#L664-L690

I would note that this link was also provided: https://web.mit.edu/kerberos/krb5-1.21/doc/mitK5license.html which seems to also have some extra stuff at the bottom and I'm not sure if that's supposed to be part of the license. But based on the "go with the license closest to the code" guidance, I'd go with the link above.

jlovejoy:
did a bit more research to see if this is used elsewhere and found:
* looks like a bunch of files in pthread use it, see https://opensource.apple.com/source/libpthread/libpthread-416.60.2/man/pthread_kill.2.auto.html
[...]

PR #2435, opened 2024-03-29
add BSD-2-clause-first-lines

xsuchy:
Fixes: #2348


xsuchy:
Ah, got it. I had incorrect id and name (copied form spdx tools template). Fixed.

swinslow:
See #2487 re: fixing the capitalization for "C" -- I know lowercase 'c' is what the notes in #2348 said, but I suspect that was in error.

PR #2436, opened 2024-03-30
chore: update to actions/setup-python@v5

xsuchy:
v3 is using node16
Cauing a warning:

Validate schema
Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: actions/setup-python@v3. For more information see: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/.

v5 seems to be current according to https://github.com/actions/setup-python


jlovejoy:
@goneall @swinslow - I think one of you is more qualified to review this one!

Issue #2430, opened 2024-03-26
LGPL-2.x(-*) templates are missing a line of text compared to the GNU-published texts, causing matching to fail


pmonks:
The [LGPL-2.0 license text published by GNU](https://www.gnu.org/licenses/lgpl-2.0.txt) has, at line 102, this text:
```text
GNU LIBRARY GENERAL PUBLIC LICENSE
```
Similarly, the [LGPL-2.1 license text published by GNU](https://www.gnu.org/licenses/lgpl-2.1.txt) also has the same line of text, at line 115.

The associated SPDX templates are all missing this text, so matching against the GNU-published texts fails:
* [LGPL-2.0](https://github.com/spdx/license-list-XML/blob/main/src/LGPL-2.0.xml#L154)
* [LGPL-2.0-only](https://github.com/spdx/license-list-XML/blob/main/src/LGPL-2.0-only.xml#L160)
* [LGPL-2.0-or-later](https://github.com/spdx/license-list-XML/blob/main/src/LGPL-2.0-or-later.xml#L146)
* [LGPL-2.1](https://github.com/spdx/license-list-XML/blob/main/src/LGPL-2.1.xml#L170)
* [LGPL-2.1-only](https://github.com/spdx/license-list-XML/blob/main/src/LGPL-2.1-only.xml#L172)
* [LGPL-2.1-or-later](https://github.com/spdx/license-list-XML/blob/main/src/LGPL-2.1-or-later.xml#L154)

karsten-klein:
... with respect to the LGPL-2.0 license text...

**{metæffekt} Universe**
canonical name: [GNU Library General Public License 2.0](https://github.com/org-metaeffekt/metaeffekt-universe/blob/main/src/main/resources/ae-universe/[g]/README.md#user-content-GNU-Library-General-Public-License-2.0:~:text=GNU%20Library%20General%20Public%20License%202.0)
short name: LGPL-2.0
markers: General Terms Matches Marker, Incompatible with Secondary License Marker, No Warranty Marker, Patent Information Marker
category: LGPL
SPDX reference id: [LGPL-2.0-only](https://spdx.org/licenses/LGPL-2.0-only.html)
ScanCode reference id: [lgpl-2.0](https://github.com/nexB/scancode-toolkit/blob/develop/src/licensedcode/data/licenses/lgpl-2.0.LICENSE)
OSI status: approved
Open CoDE status: approved
[...]

PR #2440, opened 2024-04-09
Addresses issue #2430

pmonks:
Comments / suggestions / edits welcome.

### DIFF
--- a/src/LGPL-2.0-only.xml
+++ b/src/LGPL-2.0-only.xml
@@ -157,6 +157,7 @@
       ordinary General Public License rather than by this special one.
     </p>
     <p>
+      GNU LIBRARY GENERAL PUBLIC LICENSE
       TERMS AND CONDITIONS FOR COPYING, DISTRIBUTION AND MODIFICATION
     </p>
     <list>
--- a/src/LGPL-2.0-only.xml
+++ b/src/LGPL-2.0-only.xml
@@ -158,6 +158,8 @@
     </p>
     <p>
       GNU LIBRARY GENERAL PUBLIC LICENSE
+    </p>
+    <p>
       TERMS AND CONDITIONS FOR COPYING, DISTRIBUTION AND MODIFICATION
     </p>
     <list>
--- a/src/LGPL-2.0+.xml
+++ b/src/LGPL-2.0+.xml
@@ -78,6 +78,7 @@
          with the library.</p>
       <p>Note that it is possible for a library to be covered by the ordinary General Public License rather than
          by this special one.</p>
+      <p>GNU LIBRARY GENERAL PUBLIC LICENSE</p>
       <p>TERMS AND CONDITIONS FOR COPYING, DISTRIBUTION AND MODIFICATION</p>
       <list>
         <item>
--- a/src/LGPL-2.0-or-later.xml
+++ b/src/LGPL-2.0-or-later.xml
@@ -144,6 +144,8 @@
     </p>
     <p>
       GNU LIBRARY GENERAL PUBLIC LICENSE
+    </p>
+    <p>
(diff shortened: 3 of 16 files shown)


src/LGPL-2.0-only.xml
```diff
@@ -157,6 +157,7 @@
       ordinary General Public License rather than by this special one.
     </p>
     <p>
+      GNU LIBRARY GENERAL PUBLIC LICENSE
```
xsuchy, line 160: Can you wrap it to <p></p>?  The same in 5 other files.

xsuchy:
Additionally, it would be great if you could describe the commit text more descriptively. E.g.

```
Add a missing title to LGPL licenses

Fixes: #2430 
```

src/LGPL-2.0-only.xml
xsuchy: Not sure what you mean.

Before your change, there was:
```
<p>
   TERMS AND CONDITIONS FOR COPYING, DISTRIBUTION AND MODIFICATION
 </p>
```
That is one paragraph.

After your change, there is:
```
<p>
   GNU LIBRARY GENERAL PUBLIC LICENSE
   TERMS AND CONDITIONS FOR COPYING, DISTRIBUTION AND MODIFICATION
 </p>
```
That is still one paragraph. And https://spdx.org/licenses/preview/LGPL-2.0+ will render it as one paragraph:
```
GNU LIBRARY GENERAL PUBLIC LICENSE TERMS AND CONDITIONS FOR COPYING, DISTRIBUTION 
AND MODIFICATION
```

I am asking for change:
```
<p>
   GNU LIBRARY GENERAL PUBLIC LICENSE
</p>
</p>
   TERMS AND CONDITIONS FOR COPYING, DISTRIBUTION AND MODIFICATION
 </p>
```

or 

```
<p>
   GNU LIBRARY GENERAL PUBLIC LICENSE
   <br/>TERMS AND CONDITIONS FOR COPYING, DISTRIBUTION AND MODIFICATION
 </p>
```

jlovejoy:
Thanks @pmonks ! 

To add a bit of explanation to @xsuchy comment above about the commit text: if you put "Fixes #[issue number] in the first comment for the PR, then when that PR is merged, it will automatically close the associated issue, which is quite handy. I didn't know that before, so thought I'd pass that along in case you didn't either! (and if you did, then ignore me)

I think the tests are failing b/c the changes also need to be made to the test text files - the tests run to make sure the xml and txt files match. I'll push a commit to illustrate what I mean.

jlovejoy:
@pmonks - I just realized that you also updated the deprecated LGPLs - I don't think we need to?? (sorry, should have caught this in the issue)

pmonks:
@jlovejoy if they're not updated, they'll never match texts out in the wild. Is that the desired behaviour? If so, might it be better to remove those templates entirely instead?

It seems to me that having "broken" templates is worse than not having them; if they exist there's an expectation that they're correct.

jlovejoy:
@pmonks - ah, okay. wasn't sure if that was an issue or how that works. Yes, then we should update them too!

pmonks:
@jlovejoy yeah I'm not using the matching templates for deprecated licenses myself, but I have no way of knowing if others might be, so updated them out of an abundance of caution.

src/LGPL-2.0-only.xml
jlovejoy: FYI - appreciate the attention to detail @xsuchy and I think my commits may have perpetuated the lack of line break (personally, I prefer the <br/> for this case :) - but if the tests pass, I'm going to merge this. we can see how it looks and do any formatting updates later, if need be.

Issue #2431, opened 2024-03-26
CC-BY-3.0 template is missing a line of text compared to the Creative Commons-published texts, causing matching to fail


pmonks:
The [CC-BY-3.0 license text published by Creative Commons](https://creativecommons.org/licenses/by/3.0/legalcode.txt) commences with this text:
```text
Creative Commons Legal Code
```
But the [associated SPDX template](https://github.com/spdx/license-list-XML/blob/main/src/CC-BY-3.0.xml#L12-L18) is missing this text, so matching against the Creative Commons-published text fails.

karsten-klein:
**{metæffekt} Universe**
canonical name: [Creative Commons BY 3.0](https://github.com/org-metaeffekt/metaeffekt-universe/blob/main/src/main/resources/ae-universe/[c]/README.md#user-content-Creative-Commons-BY-3.0:~:text=Creative%20Commons%20BY%203.0)
markers: Incompatible with Secondary License Marker
category: CC-BY
SPDX reference id: [CC-BY-3.0](https://spdx.org/licenses/CC-BY-3.0.html)
ScanCode reference id: [cc-by-3.0](https://github.com/nexB/scancode-toolkit/blob/develop/src/licensedcode/data/licenses/cc-by-3.0.LICENSE)
OSI status: pending
Open CoDE status: approved
Open CoDE approved license id: CC-BY-3.0 (SPDX)

**ScanCode**
matched id: [cc-by-3.0](https://github.com/nexB/scancode-toolkit/blob/develop/src/licensedcode/data/licenses/cc-by-3.0.LICENSE)

pmonks:
@jlovejoy I did check one or two of the `CC-BY-*-4.0` license texts, and they don't have this line, so they're unaffected.  I didn't check either `CC-BY-*-1.0` or `CC-BY-*-2.0`.

PR #2441, opened 2024-04-09
Addresses issue #2431 (CC licenses issue)

pmonks:
Comments / suggestions / edits welcome.

I checked all CC-BY-*-3.0 licenses, and only modified those with text renditions published by Creative Commons (CC doesn't publish all of their license variations in plain text, sadly).


xsuchy:
Tests are failing. Can you edit the tests in test/simpleTestForGenerator/ too?
I guess that the tests should match even the optional text.

pmonks:
@xsuchy how do I run the tests locally, so I can confirm I've fixed them before updating the PR?

xsuchy:
```
./test-one-license CC-BY-SA-3.0
```

pmonks:
This is now fixed, however the `CC-BY-3.0` test is still failing due to an unrelated issue in that template.  I'll raise a separate issue for that.

xsuchy:
Note that I am not a huge fan of rewriting something just because - i.e. you reformated all the tests, that makes the diff very ugly.
And you are mixing two issues together. Yes, the bullet e. is different in CC legal text  https://creativecommons.org/licenses/by/3.0/legalcode.txt compared to what we have in XML and test, but that is for different discussion. So in this PR we should stick just to adding the missing header. Nothing else.

pmonks:
@xsuchy

> Note that I am not a huge fan of rewriting something just because - i.e. you reformated all the tests, that makes the diff very ugly.

I copied the license texts from the license author verbatim into the tests (for example [here's the official `CC-BY-3.0` license text](https://creativecommons.org/licenses/by/3.0/legalcode.txt) that I copied into the `CC-BY-3.0.txt` file in the tests in this PR).  I consider it **exceptionally** bad practice to edit those canonical texts *in any way*, but of course this PR is editable so you may do so if you prefer introducing inconsistencies in order to preserve diff aesthetics.

> And you are mixing two issues together. Yes, the bullet e. is different in CC legal text https://creativecommons.org/licenses/by/3.0/legalcode.txt compared to what we have in XML and test, but that is for different discussion. So in this PR we should stick just to adding the missing header. Nothing else.

I'm not mixing anything together.  As I [clearly wrote above](https://github.com/spdx/license-list-XML/pull/2441#issuecomment-2046064208), "I'll raise a separate issue for that."  That issue was indeed raised shortly after I wrote that, and can be [seen here](https://github.com/spdx/license-list-XML/issues/2445).

jlovejoy:
hey @pmonks and @xsuchy - I'm a bit lost on the last few comments about re-writing tests?? Or maybe there is a typo somewhere and that was really about "texts"?

In any case, my understanding of the original issue #2431 was to add the missing title text, which @pmonks did in the xml files. But the text files also needed to be updated. In doing so, it looks like @pmonks cut and pasted the whole license from the CC text files which has line breaks where we didn't before (which makes the diff look messy, but it's really not a substantive change) - is that right?

Also just read the issue with CC-BY-3.0 - @pmonks - can you just delete the xml file for that license from this PR? We are going to need to update that otherwise and then this will pass tests.

pmonks:
> In any case, my understanding of the original issue #2431 was to add the missing title text, which @pmonks did in the xml files. But the text files also needed to be updated. In doing so, it looks like @pmonks cut and pasted the whole license from the CC text files which has line breaks where we didn't before (which makes the diff look messy, but it's really not a substantive change) - is that right?

That's correct, yes. In doing so, I also just happened to discover issue #2445 (which is unrelated to this issue & PR, except by virtue of colliding via the updated `CC-BY-3.0` text file).

jlovejoy:
ugh, I think I messed this up. I was trying to delete the CC-BY-3.0 file from this PR, but I think I just simply deleted it. Given the CC-CY-3.0 has other issues and there is now a PR to address that separately, we can not include those files here. Let me see how I can un-do or otherwise clean-up my mess...

pmonks:
Or maybe just close this PR without merging it?  I _think_ (but am not 100% sure) that because I already merged it into PR #2445 the changes that are here will come through via that PR anyway.

jlovejoy:
ah, thanks @pmonks - I will close this one then. Sorry for screwing it up and making more work

PR #2455, opened 2024-04-21
Add cross-ref for ISC-Veillard variant

nwellnhof:
This license originates from the Trio printf library.


xsuchy:
+1 to adding crossref, but not sure about the note.

jlovejoy:
@nwellnhof - thanks for the URL addition. We use the Notes for pretty limited scope of information, see https://github.com/spdx/license-list-XML/blob/main/DOCS/license-fields.md#d-notes
Did you have a particular thought behind adding it?

nwellnhof:
> Did you have a particular thought behind adding it?

Well, the name `ISC-Veillard` is a bit misleading. It is used in libxml2, originally authored by Daniel Veillard, but this license was copied from the Trio library which I think is the original source.

swinslow:
Discussed on 2024-05-09 legal team call. We're good to include the note, but we'd lean towards saying "It appears to originate from Trio" rather than "It originates from Trio" since we usually don't opine definitively on license history.

I'll make this change and then merge. Thank you all!

PR #2464, opened 2024-05-02
Move matching guidelines

jlovejoy:
move matching guidelines to SPDX License List repo, DOCS (from Spec Annex) as of version 3.0 of SPDX spec


jlovejoy:
note, this move will mean that links to the matching guidelines will likely still go to v2.3 and no longer be the most up-to-date version, so we'll need to work on letting people know and updating links that we know of

swinslow (approved):
LGTM, thank you @jlovejoy! We can figure out longer-term whether to rework the formatting, but for now agreed that this works.

Issue #2243, opened 2023-11-05
New license request: any-osi [SPDX-Online-Tools]


xsuchy:
**1.** License Name: Any OSI License
**2.** Short identifier: any-osi
**3.** License Author or steward: Unknown
**4.** Comments: This license is used in perl module Exporter::Tidy and is used in Fedora
**5.** License Request Url: http://tools.spdx.org/app/license_requests/320
**6.** URL(s): https://metacpan.org/pod/Exporter::Tidy#LICENSE
**7.** OSI Status: Unknown
**8.** Example Projects: https://metacpan.org/pod/Exporter::Tidy

xsuchy:
Text of the license:

Pick your favourite OSI approved license :)

http://www.opensource.org/licenses/alphabetical

jlovejoy:
ugh, that is really annoying that Tidy has done that. From SPDX perspective, we can't have an ID that represents multiple licenses (and in this case, which are very different!). 

I think Fedora can just pick one of the OSI licenses - at least that seems to be the intent. 

@richardfontana did you see this?

richardfontana:
I saw this issue but not sure if @xsuchy has submitted an issue on the Fedora side. While Fedora could pick one of the OSI licenses, that is not what it has traditionally done; the current Callaway license tag for this package is: 
```
AAL or AFL or AGPLv3 or APSL 2.0 or ASL 2.0 or Artistic 2.0 or BSD or
             : Boost or CATOSL or CDDL-1.0 or CNRI or CPAL or CeCILL or ECL 2.0 or EFL
             : 2.0 or EPL-1.0 or EU Datagrid or EUPL 1.1 or Entessa or Fair or GPLv2
             : or GPLv3 or IBM or IPA or ISC or LGPLv2 or LGPLv3 or LPL or LPPL or MIT
             : or MPLv1.1 or MPLv2.0 or MS-PL or MS-RL or MirOS or Motosoto or NCSA or
             : NGPL or Naumen or Nokia or OFL or OSL 3.0 or PHP or PostgreSQL or
             : Python or QPL or RPSL or SPL or Sleepycat or VSL or W3C or ZPLv2.0 or
             : zlib
```
[...]

PR #2470, opened 2024-05-09
add Any OSI License

xsuchy:
Fixes: #2243


swinslow:
@xsuchy Thanks! Just updated to add the `listVersionAdded` value. After the tests pass again, I'll merge.

I do need to note the irony of `isOsiApproved="false"` here. I think it's correct, it just bothers me.  :)

swinslow (approved):

Issue #2349, opened 2024-01-25
New license request: HPND-export-US-ext-disclaimer [SPDX-Online-Tools]


jrisc:
**1.** License Name: HPND with US Government export control warning and extended disclaimer
**2.** Short identifier: HPND-export-US-ext-disclaimer
**3.** License Author or steward: Unknown
**4.** Comments: This license is similar to HPND-export-US, but it has a longer disclaimer.

This license is used in Fedora, and was found during license review:
https://gitlab.com/fedora/legal/fedora-license-data/-/issues/467
**5.** License Request Url: http://tools.spdx.org/app/license_requests/342
**6.** URL(s): https://github.com/krb5/krb5/blob/krb5-1.21.2-final/NOTICE#L831-L852, https://web.mit.edu/kerberos/krb5-1.21/doc/mitK5license.html
**7.** OSI Status: Not Submitted
**8.** Example Projects: https://web.mit.edu/kerberos/

jlovejoy:
Aside from the differences in the disclaimer section, this also has a different obligation relating to notice:
`WITHIN THAT CONSTRAINT, permission to copy, modify, and distribute
   this software and its documentation in source and binary forms is
   hereby granted, provided that any documentation or other materials
   related to such distribution or use acknowledge that the software
   was developed by the University of Southern California.`

and omits the part about not using it in advertising or publicity

I'd add as new license - maybe `HPND-export-US-acknowledgement`?
or could use a different name, since it's borderline on close matching to `HPND-export-US`

karsten-klein:
**{metæffekt} Universe**
canonical name: [Permission Terms (with export note; variant 002)](https://github.com/org-metaeffekt/metaeffekt-universe/blob/main/src/main/resources/ae-universe/[p]/README.md#user-content-Permission-Terms-(with-export-note;variant-002):~:text=Permission%20Terms%20(with%20export%20note;%20variant%20002))
short name: PT-with-export-note-Var-02
markers: Export Restriction Note (US)
category: Permission
OSI status: _none_

**ScanCode**
matched id: [mit-no-advert-export-control](https://github.com/nexB/scancode-toolkit/blob/develop/src/licensedcode/data/licenses/mit-no-advert-export-control.LICENSE)
matched id: [proprietary-license](https://github.com/nexB/scancode-toolkit/blob/develop/src/licensedcode/data/licenses/proprietary-license.LICENSE)

**Comment**
+1 to add following @jlovejoy's proposal.

[...]

PR #2474, opened 2024-05-14
add HPND-export-US-acknowledgement

xsuchy:
Fixes: #2349

I am not sure if the "rqmt" in the license name was intentional or a typo.


jlovejoy:
it was a typo - fixed now!

PR #2475, opened 2024-05-14
Create Asterisk-linking-protocols-exception.xml

### LINKED ISSUE #2390: Asterisk linking exceptions
In the [LICENSE file of asterisk](https://github.com/asterisk/asterisk/blob/master/LICENSE) there is a custom paragraph to allow some linking:
> Specific permission is also granted to link Asterisk with OpenSSL, OpenH323
UniMRCP, and/or the UW IMAP Toolkit and distribute the resulting binary files.

I haven't found a way to express that in a spdx license identifier.

`Asterisk-exception` is for something different if I understand it correctly.

jlovejoy:
fixes #2390


swinslow (approved):
+1, LGTM

PR #2480, opened 2024-05-17
Apache-2.0: Break out optional end sections

### LINKED ISSUE #2418: Apache-2.0 license text without appendix is a closer match for "Pixar" than "Apache-2.0" since v3.23
Many projects in Fedora Linux are using askalono to match license files against SPDX. Since we updated the SPDX data to version 3.23, we're getting somewhat unexpected results for files that are definitely Apache-2.0 license texts. If the "APPENDIX" is present, then matches are reported with a score of `1.000`.

But if the (optional? looking at https://spdx.org/licenses/Apache-2.0) appendix is not present in the file, it is now reported as a closer match to the newly added "Pixar" license (with score `0.983`) despite being a perfect match to the Apache-2.0 license text without appendix.

For example, the LICENSE-APACHE file from the Rust programming language (which is definitely Apache-2.0, not Pixar) is now classified as "Pixar" because it's apparently a closer match: https://github.com/rust-lang/rust/blob/master/LICENSE-APACHE

So it appears that there's some issue with the matching data, either for Apache-2.0 or the newly added Pixar license.

swinslow:
Fixes #2418 

### DIFF
--- a/src/Apache-2.0.xml
+++ b/src/Apache-2.0.xml
@@ -196,6 +196,8 @@
 
       <optional>
          <p>END OF TERMS AND CONDITIONS</p>
+      </optional>
+      <optional>
          <p>APPENDIX: How to apply the Apache License to your work.</p>
          <p>To apply the Apache License to your work, attach the following boilerplate notice, with the fields
          enclosed by brackets "[]" replaced with your own identifying information. (Don't


swinslow:
@jlovejoy Want to merge this if it looks good to you?

Issue #2017, opened 2023-06-26
New license request: MIT-Khronos [SPDX-Online-Tools]


nicholas-rh:
**1.** License Name: MIT Khronos Variant
**2.** Short identifier: MIT-Khronos
**3.** License Author or steward: Khronos Group
**4.** Comments: It appears to be a variant of the MIT license. The licensecheck tool reports it as a separate license called "Khronos license." Here is the Fedora license review discussion, in which it was approved for Fedora: https://gitlab.com/fedora/legal/fedora-license-data/-/issues/238
**5.** License Request Url: http://tools.spdx.org/app/license_requests/240
**6.** URL(s): https://github.com/KhronosGroup/SPIRV-Cross/blob/main/LICENSES/LicenseRef-KhronosFreeUse.txt
**7.** OSI Status: Not Submitted
**8.** Example Projects: https://github.com/KhronosGroup/glslang, https://github.com/KhronosGroup/SPIRV-Cross

laurathefuerte:
I can take this one with @Pizza-Ria.

Pizza-Ria:
@nicholas-rh - is the SPIRV-Cross project that mainly seems to use this license integrated into O3DE as well or just utilized as part of Fedora?  I noticed some O3DE stuff in your repo including https://github.com/nicholas-rh/o3de-fedora.  Thanks!

nicholas-rh:
> @nicholas-rh - is the SPIRV-Cross project that mainly seems to use this license integrated into O3DE as well or just utilized as part of Fedora? I noticed some O3DE stuff in your repo including https://github.com/nicholas-rh/o3de-fedora. Thanks!

It is used as part of O3DE, but there isn't a Fedora package for it. I was looking into packaging SPIRV-Cross when I came across the license and started looking for clarification on it.

nicholas-rh:
It's also found in files in the glslang project which is packaged for Fedora: https://github.com/KhronosGroup/glslang/blob/main/SPIRV/spirv.hpp

jlovejoy:
This matches MIT, except for:
- the use of "Materials" instead of "Software"
- the addition of the paragraph:
`MODIFICATIONS TO THIS FILE MAY MEAN IT NO LONGER ACCURATELY REFLECTS KHRONOS
STANDARDS. THE UNMODIFIED, NORMATIVE VERSIONS OF KHRONOS SPECIFICATIONS AND
[...]

PR #2496, opened 2024-06-10
markup change to MIT-Khronos-old

xsuchy:
In the review in #2017 it was said that this paragraph: This text may be omittable (B.3.5 Guideline: omittable text), as it is merely informative or explanatory, not any right grant or restriction.

but the optional tags were not added in the initial commit. Adding now.

### DIFF
--- a/src/MIT-Khronos-old.xml
+++ b/src/MIT-Khronos-old.xml
@@ -26,12 +26,14 @@
             The above copyright notice and this permission notice shall be
             included in all copies or substantial portions of the Materials.
          </p>
+         <optional>
          <p>
             MODIFICATIONS TO THIS FILE MAY MEAN IT NO LONGER ACCURATELY
             REFLECTS KHRONOS STANDARDS. THE UNMODIFIED, NORMATIVE
             VERSIONS OF KHRONOS SPECIFICATIONS AND HEADER INFORMATION
             ARE LOCATED AT https://www.khronos.org/registry/
          </p>
+         </optional>
          <p>
             THE MATERIALS ARE PROVIDED "AS IS", WITHOUT WARRANTY OF
             ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO


swinslow:
Hi @xsuchy, thanks for this.

I'm not sure I agree with that conclusion from the reviewer of the issue originally. I think they were proposing to make that optional in the course of deciding whether it could be considered a match for standard MIT (which we decided it wasn't, hence the new ID).

If we make that paragraph optional, then I think the only differences between this and the standard MIT license would be using "Materials" instead of "Software as the defined term; and an "and/or" instead of "and" in the first sentence.

Just to ask, though, are you seeing any instances of the MIT-Khronos-old text appearing without that central paragraph? Put another way, are there examples you're encountering where making this paragraph optional would cause it to match a license text you've seen?

xsuchy:
> Just to ask, though, are you seeing any instances of the MIT-Khronos-old text appearing without that central paragraph? 

Yes. 
https://sourceforge.net/p/mingw-w64/mingw-w64/ci/master/tree/mingw-w64-headers/include/GL/glext.h

swinslow:
Got it, thanks @xsuchy! 

@jlovejoy So this is a tricky one. It is _either_ (1) very close to standard MIT, except with "Materials" instead of "Software" and a couple of other very minor changes:

image

...or (2) very close to MIT-Khronos-old, but without the central paragraph:

image

Do you have a preference for how to handle this? I'd lean towards minor markup to MIT as I think that's substantively closer rather than making the central paragraph optional for the Khronos variant, but I could be persuaded either way.

xsuchy:
Gentle ping @jlovejoy

pombredanne:
FWIW, we have tracked this MIT variant without the extra paragraph as a separate license for about 10 years: https://github.com/aboutcode-org/scancode-toolkit/blob/develop/src/licensedcode/data/licenses/khronos.LICENSE

jlovejoy:
Sorry this slipped for so long @xsuchy - Steve and I finally looked more closely at this. We do not want to revise `MIT-Khronos-old` as that was denoted as old because of that center paragraph. We can add markup to MIT to accommodate "material" though, so I'm going to close this PR and create a new one.

PR #2498, opened 2024-06-14
Use HTTPS for URLs in MPL-1.1 and MPL-2.0

leoheitmannruiz:
As in the texts published at https://www.mozilla.org/en-US/MPL/1.1/ and https://www.mozilla.org/en-US/MPL/2.0/


xsuchy:
Per https://spdx.github.io/spdx-spec/v2.3/license-matching-guidelines-and-templates/#b14-http-protocol http and https are considered equal for the matching purpose. But +1 from me. The original license has it now. It will make no harm if we change it in our XML file.

xsuchy (approved):

Issue #2507, opened 2024-07-02
Markup change of Boehm-GC


xsuchy:
During Fedora License review https://gitlab.com/fedora/legal/fedora-license-data/-/issues/530 we discovered  in
mariadb10.11
https://github.com/MariaDB/server/blob/11.6/libmysqld/lib_sql.cc

a license that has the wording:

```
This material is provided "as is", with absolutely no warranty expressed
or implied. Any use is at your own risk.

Permission to use or copy this software for any purpose is hereby
granted without fee, provided the above notices are retained on all
copies.

Permission to modify the code and to distribute modified code is
granted, provided the above notices are retained, and a notice that
the code was modified is included with the above copyright notice.
```

This is very close to Boehm-GC.
I will file a PR with markup change proposal where it will be more visible what needs to be changed.

karsten-klein:
**{metæffekt} Universe**
canonical name: [STLport License 4.5](https://github.com/org-metaeffekt/metaeffekt-universe/blob/main/src/main/resources/ae-universe/[s]/README.md#user-content-STLport-License-4.5:~:text=STLport%20License%204.5)
short name: STLport-4.5
category: STLport License
ScanCode reference id: [stlport-4.5](https://github.com/nexB/scancode-toolkit/blob/develop/src/licensedcode/data/licenses/stlport-4.5.LICENSE)
OSI status: _none_
Open CoDE status: approved
Open CoDE approved license id: stlport-4.5 (ScanCode)

**ScanCode**
matched id: [stlport-4.5](https://github.com/nexB/scancode-toolkit/blob/develop/src/licensedcode/data/licenses/stlport-4.5.LICENSE)

**Comment**
I'm not fully aware of the history at scancode. +1 to add dur to harmonization considerations.

jlovejoy:
copying the license texts here b/c I'm having a hard time "seeing" in the PR!

`This material is provided "as is", with absolutely no warranty expressed
or implied. Any use is at your own risk.`
--> this is the same, other than not being in all caps and use of quotes around "as is"

`Permission to use or copy this software for any purpose is hereby granted without fee, `
--> versus "Permission is hereby granted to use or copy this program for any purpose" - so reverse order for the grant, but this version adds "without fee"

`provided the above notices are retained on all copies.`
--> same

`Permission to modify the code and to distribute modified code is granted, provided the above notices are retained, and a notice that the code was modified is included with the above copyright notice.`
--> also looks the same

[...]

PR #2508, opened 2024-07-02
markup change to Boehm-GC

xsuchy:
To accomodate the license found in mariadb
Fixes: #2507

### DIFF
--- a/src/Boehm-GC.xml
+++ b/src/Boehm-GC.xml
@@ -6,6 +6,7 @@
          <crossRef>https://fedoraproject.org/wiki/Licensing:MIT#Another_Minimal_variant_(found_in_libatomic_ops)</crossRef>
          <crossRef>https://github.com/uim/libgcroots/blob/master/COPYING</crossRef>
          <crossRef>https://github.com/ivmai/libatomic_ops/blob/master/LICENSE</crossRef>
+         <crossRef>https://github.com/MariaDB/server/blob/11.6/libmysqld/lib_sql.cc</crossRef>
       </crossRefs>
       <text>
          <copyrightText>Copyright (c) ...</copyrightText>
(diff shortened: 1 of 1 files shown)


src/Boehm-GC.xml
```diff
@@ -14,8 +15,8 @@
             WARRANTY EXPRESSED OR IMPLIED. ANY USE IS AT YOUR OWN RISK.
          </p>
          <p>
-            Permission is hereby granted to use or copy this program for any
-            purpose, provided the above notices are retained on all copies.
+            Permission <optional>is hereby granted to</optional> use or copy this <alt name="program" match="program|software">program</alt> for any
```
richardfontana, line 18: I think this has to be 
```
<optional>is hereby granted</optional> to use [etc.]

Comment (xsuchy) [reply src/Boehm-GC.xml]: True. Updated.

File: src/Boehm-GC.xml [hunk 2]
```diff
@@ -14,8 +15,8 @@
             WARRANTY EXPRESSED OR IMPLIED. ANY USE IS AT YOUR OWN RISK.
          </p>
          <p>
-            Permission is hereby granted to use or copy this program for any
-            purpose, provided the above notices are retained on all copies.
+            Permission <optional>is hereby granted to</optional> use or copy this <alt name="program" match="program|software">program</alt> for any
+            purpose, <optional>is hereby granted without fee,</optional> provided the above notices are retained on all copies.
             Permission to modify the code and to distribute modified code is
```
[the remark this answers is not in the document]
swinslow: @richardfontana I'd say probably not (i.e. it probably doesn't matter).

Agreed that this is in theory an issue for lots of SPDX XML files, and especially those with "anything goes" / `.*` regular expressions. However, I think in practice we accept that this is the practice at present. I'm not aware of any "real world" license texts that have tried to abuse it in a meaningful way.

src/Boehm-GC.xml
```diff
@@ -14,8 +15,8 @@
             WARRANTY EXPRESSED OR IMPLIED. ANY USE IS AT YOUR OWN RISK.
          </p>
          <p>
-            Permission is hereby granted to use or copy this program for any
-            purpose, provided the above notices are retained on all copies.
+            Permission <optional>is hereby granted</optional> to use or copy this <alt name="program" match="program|software">program</alt> for any
+            purpose, <optional>is hereby granted without fee,</optional> provided the above notices are retained on all copies.
```
swinslow, line 19: I'm not sure this actually matches https://github.com/MariaDB/server/blob/11.6/libmysqld/lib_sql.cc with the present edits. Note the comma after "purpose" in Boehm-GC which isn't present in the MariaDB link.

Might be able to make that optional, but ugh... to @richardfontana's point, this does start to look a bit ugly...

xsuchy:
Closing per discussion in the associated issue.

Issue #2495, opened 2024-06-04
New license request: erlang-otp-linking-exception [SPDX-Online-Tools]


laf0rge:
**1.** License Name: AGPLv3 Erlang/OTP linking exception
**2.** Short identifier: erlang-otp-linking-exception
**3.** License Author or steward: Harald Welte
**4.** Comments: * The exception is not a stand-alone license, but an additional permission to AGPLv3
* it grants an additional permission
* it does not impose additional restrictions
* it is used since 2012 in a variety of (at least) Osmocom projects, see https://erlang.org/pipermail/erlang-questions/2012-May/066355.html for the original proposal
**5.** License Request Url: http://tools.spdx.org/app/license_requests/371
**6.** URL(s): https://gitea.osmocom.org/erlang/osmo_ss7/src/commit/2286c1b8738d715950026650bf53f19a69d6ed0e/src/ss7_links.erl#L20
**7.** OSI Status: Approved
**8.** Example Projects: https://gitea.osmocom.org/erlang/osmo_ss7, https://gitea.osmocom.org/erlang/osmo-epdg

swinslow:
Text of requested license exception to add:

```
If you modify this Program, or any covered work, by linking or
combining it with runtime libraries of Erlang/OTP as released by
Ericsson on http://www.erlang.org (or a modified version of these
libraries), containing parts covered by the terms of the Erlang Public
License (http://www.erlang.org/EPLICENSE), the licensors of this
Program grant you additional permission to convey the resulting work
without the need to license the runtime libraries of Erlang/OTP under
the GNU Affero General Public License. Corresponding Source for a
non-source form of such a combination shall include the source code
for the parts of the runtime libraries of Erlang/OTP used as well as
that of the covered work.
```

swinslow:
This exception is close to [`GPL-3.0-linking-source-exception`](https://spdx.org/licenses/GPL-3.0-linking-source-exception.html), taking into account the replaceable text bits of that exception.

However, it does add an additional phrase that isn't present in `GPL-3.0-linking-source-exception`:

> . . . without the need to license the runtime libraries of Erlang/OTP under the GNU Affero General Public License.

I'm not inclined to add that phrase as markup to `GPL-3.0-linking-source-exception`, since that exception is [based on a template taken from the FSF's FAQ](https://www.gnu.org/licenses/gpl-faq.en.html#GPLIncompatibleLibs).

So I'm +1 to add this as a new exception on the license list.

[...]

PR #2527, opened 2024-07-29
Add new erlang-otp-linking-exception.xml

laf0rge:
I've tried to use an <alt> tag to make the exception work with other licenses.  The orignal wording was focused on AGPLv3, but it would just as well work with GPLv3, for example.

Fixes: #2495

### DIFF
--- a/src/exceptions/erlang-otp-linking-exception.xml
+++ b/src/exceptions/erlang-otp-linking-exception.xml
@@ -10,13 +10,11 @@
       This exception is based on the suggested template from the Free Software Foundation's FAQ about the GPL.
     </notes>
     <text>
-      <titleText>
 	<optional>
 	<p>
 	  Additional permission under GNU AGPL version 3 section 7
 	</p>
 	</optional>
-      </titleText>
       <p>
 	If you modify this Program, or any covered work, by linking or
 	combining it with runtime libraries of Erlang/OTP as released by
--- a/src/exceptions/erlang-otp-linking-exception.xml
+++ b/src/exceptions/erlang-otp-linking-exception.xml
@@ -1,0 +1,36 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<SPDXLicenseCollection xmlns="http://www.spdx.org/license">
+  <exception licenseId="erlang-otp-linking-exception" name="Erlang/OTP Linking Exception" listVersionAdded="3.25.0">
+    <crossRefs>
+      <crossRef>https://www.gnu.org/licenses/gpl-faq.en.html#GPLIncompatibleLibs</crossRef>
+      <crossRef>https://erlang.org/pipermail/erlang-questions/2012-May/066355.html</crossRef>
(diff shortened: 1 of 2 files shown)


src/exceptions/erlang-otp-linking-exception.xml
```diff
@@ -0,0 +1,36 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<SPDXLicenseCollection xmlns="http://www.spdx.org/license">
+  <exception licenseId="erlang-otp-linking-exception" name="Erlang/OTP Linking Exception" listVersionAdded="3.25">
```
xsuchy, line 3: The version should be 3.25.0. Since 3.24.0 we use 3 components.

xsuchy: This is a new thing, so I guess the DOC was not updated. I will open a PR to update it.

xsuchy: I did https://github.com/spdx/license-list-XML/pull/2528

swinslow: Sorry about that @laf0rge -- yes, still catching up on documentation updates. Thanks @xsuchy for the fix in the docs.

swinslow:
Hi @laf0rge, it looks like the tests are failing because a test text file hasn't been included in the PR for this new exception. 

For new licenses / exceptions, the tests check for a file with the same ID and a `.txt` extension in the `/test/simpleTestForGenerator` directory, with that test text file containing the plain text version of the license (linked [here] in the original issue). Feel free to add that to the PR if you can, or if not then I can add it later today/tomorrow. Thank you!

Thompson1985 (approved):
Thank you

Thompson1985 (approved):

jlovejoy:
> I've tried to use an tag to make the exception work with other licenses. The orignal wording was focused on AGPLv3, but it would just as well work with GPLv3, for example.

Hi @laf0rge - have you seen/are there examples of this exception being used with GPLv3 (or other versions) as well as AGPLv3? We generally add markup for actual variations that have been observed, not as a hypothetical or something that could be done.  

I pushed a commit to remove the titleText tag, as that isn't needed here (sorry if that was... pushy... :)

Still needs the text file, as @swinslow  noted, but I didn't want to do too much at once.

laf0rge:
> Hi @laf0rge - have you seen/are there examples of this exception being used with GPLv3 (or other versions) as well as AGPLv3? We generally add markup for actual variations that have been observed, not as a hypothetical or something that could be done.

I know of no existing practical examples; this was just the software developer mindset in me trying to solve a problem in a generic way, rather than a specific way (which would require duplicate effort later on).

jlovejoy:
@swinslow - do you have an opinion on the markup? We aren't usually proactive in this way, but not sure it hurts anything??

swinslow:
I'm fine with including the markup -- I don't see any harm in including it here, since it's matching the actual identified exception text itself.

This looks good to go -- merging now. Thank you @laf0rge and all!

PR #2535, opened 2024-08-07
License matching guidelines -> License list matching guidelines

bact:
Make the title of the Annex match the name "SPDX License List Matching Guidelines" that use everywhere else.


jlovejoy:
@bact - I don't think we need this separate PR b/c it's covered in #2537 ??

bact:
Agree. The two eventually converged. I will close this one.

PR #2537, opened 2024-08-07
Add missing anchors B.6.4, B.15.3; Swap HTTP cases

bact:
- Remove "Annex" in the title, re-run heading numbering 
  - See comments starting from [here](https://github.com/spdx/license-list-XML/pull/2537#pullrequestreview-2226200346)
- Add missing anchors B.6.4 and B.15.3 (now 5.4 and 14.3, after re-numbering)
- Put B.15.2 (14.2) anchor to the right heading
- Make link to SPDX License List
- Add internal links to guidelines
- Put some special characters example inside \`...\` to differentiate
- HTTP cases in B.14.1 and B.14.2 (13.1 and 13.2)
  - Protocol names (HTTP and HTTPS) should be in UPPERCASE (they are initialisms)
  - The schema names at the first part of URL ("http://" and "https://", which will be the actual texts to be compared) should be in lowercase -- see https://www.rfc-editor.org/rfc/rfc9110.html#name-http-related-uri-schemes

### DIFF
--- a/DOCS/license-matching-guidelines-and-templates.md
+++ b/DOCS/license-matching-guidelines-and-templates.md
@@ -2,7 +2,7 @@
 
 ## B.1 SPDX license list matching guidelines <a name="B.1"></a>
 
-The SPDX License List Matching Guidelines provide guidelines to be used for the purposes of matching licenses and license exceptions against those included on the [SPDX License List](https://spdx.org/licenses/). There is no intent here to make a judgment or interpretation, but merely to ensure that when one SPDX user identifies a license as "BSD-3-Clause," for example, it is indeed the same license as what someone else identifies as "BSD-3-Clause" and the same license as what is listed on the SPDX License List.  As noted here, some of the matching guidelines are implemented in the XML files of the SPDX License List repository.
+The SPDX License List Matching Guidelines provide guidelines to be used for the purposes of matching licenses and license exceptions against those included on the [SPDX License List](https://spdx.org/licenses/). There is no intent here to make a judgment or interpretation, but merely to ensure that when one SPDX user identifies a license as "BSD-3-Clause," for example, it is indeed the same license as what someone else identifies as "BSD-3-Clause" and the same license as what is listed on the SPDX License List. As noted here, some of the matching guidelines are implemented in the XML files of the SPDX License List repository.
 
 ## B.2 How these guidelines are applied <a name="B.2"></a>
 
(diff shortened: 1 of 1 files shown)


DOCS/license-matching-guidelines-and-templates.md
```diff
@@ -99,15 +99,15 @@ Punctuation should be matched, unless otherwise stated in these guidelines or un
 
 Any hyphen, dash, en dash, em dash, or other variation should be considered equivalent.
 
-### B.6.4 Guideline: Quotes
+### B.6.4 Guideline: Quotes <a name="B.6.4"></a>
```
xsuchy, line 102: Is it needed for something?

DOCS/license-matching-guidelines-and-templates.md
```diff
@@ -203,31 +203,31 @@ To implement this guideline, use the `<optional>` XML element tag as described i
 
 ### B.14.1 Purpose <a name="B.14.1"></a>
 
-To avoid a license mismatch due to a difference in a hyperlink protocol (e.g. http vs. https).
+To avoid a license mismatch due to a difference in a hyperlink protocol (e.g. HTTP vs. HTTPS).
 
 ### B.14.2 Guideline <a name="B.14.2"></a>
 
-HTTP:// and HTTPS:// should be considered equivalent. 
+"http://" and "https://" should be considered equivalent.
```
xsuchy, line 210: Why? It does not match the change above.

DOCS/license-matching-guidelines-and-templates.md
bact: It is represented in other headings (see B.6.3, for example)  and it is used to point to a specific heading (when rendered in HTML).

DOCS/license-matching-guidelines-and-templates.md
bact: Strings "http://" and "https://" are the first part of a URL.
"http" and "https" in this context are URI schemes of URL.

`http` is an URI scheme for Hypertext Transfer Protocol (HTTP).

`https` is an URI scheme for Hypertext Transfer Protocol Secure (HTTPS).

HTTP and HTTPS are the name of the protocols, they are written in uppercase (as an initialism).
But their corresponding URI schemes, http and https, are written in lowercase, [similar to other URI schema](https://www.iana.org/assignments/uri-schemes/uri-schemes.xhtml).

HTTP and HTTPS URI schemes are defined here:
https://www.rfc-editor.org/rfc/rfc9110.html#name-http-related-uri-schemes

If the guideline is about a comparison of URL strings (as it shows in "xxxx://"), we should use a convention for URL.

zvr (commented):
Since you are making changes, you might as well remove all the numbering; this is not an "Annex" here

bact:
> Since you are making changes, you might as well remove all the numbering; this is not an "Annex" here

Please explain more please.

Do you mean the heading numbering?

This the current text, with heading numbering: https://github.com/spdx/license-list-XML/blob/main/DOCS/license-matching-guidelines-and-templates.md

bact:
@zvr just to make sure before making changes, do you mean to
- Remove the "Annex" word from the title
- Convert existing heading numbering from "B.1, B.2, B.2.1, B.2.2" to "1, 2, 2.1, 2.2" ?

zvr:
Yes, the whole text can be simplified to something like:

```

# SPDX license list matching guidelines

The SPDX License List Matching Guidelines provide guidelines ...

## How these guidelines are applied

### Purpose

To ensure consistent results ...

### Guideline: official license headers

The matching guidelines apply to license and exception text, ...
```

DOCS/license-matching-guidelines-and-templates.md
jlovejoy: @bact - is there a reason you changed the capitalization in two places here?

jlovejoy:
Thanks for this @bact and @zvr !

(good catch on updating the numbering - now we are back to the numbering we had had for years prior :)

I just had one question on HTTP v. http caps changes

bact:
@jlovejoy on HTTP, please see this comment https://github.com/spdx/license-list-XML/pull/2537#discussion_r1707847386

swinslow (approved):
LGTM, thanks @bact and all!

PR #2541, opened 2024-08-11
deprecate Net-SNMP.xml

### LINKED ISSUE #2506: SPDX should deprecate `Net-SNMP`
The SPDX identifier `Net-SNMP` is a single identifier meant to cover the exact contents of the COPYING file of the Net-SNMP project as they existed at a particular point in time, frozen on this webpage for the project: http://www.net-snmp.org/about/license.html

It is not uncommon for relatively old non-copyleft projects to have (at some point in their history) attempted to collect a set of relevant disparate license texts associated with multiple copyright holders into one LICENSE or COPYING file. Some such projects may further attempt to conceptualize this collection of licenses as "the license" of the project. I think there are other examples of this found in the SPDX license list, generally cases where earlier in SPDX's history, SPDX contributors regarded website presentations of project licenses as definitive. In more recent times I think SPDX has tended to look more closely at source code distributions as "primary sources" when deciding what to add to the SPDX license list.  

Anyway, in this case, the frozen state of the Net-SNMP COPYING file consists of a set of nine short permissive licenses associated with different copyright holders, most of which match `BSD-3-Clause` exactly, two of which might match `BSD-3-Clause` or might have minor substantive variations, and one of which probably matches `HPND`.

The COPYING file for Net-SNMP has continued to evolve since the time that its contents were recorded on the Net-SNMP website and the Net-SNMP identifier was added to the SPDX license list. Five new licenses have been added.  So today the COPYING file in its entirety does not match SPDX `Net-SNMP`. 
https://github.com/net-snmp/net-snmp/blob/master/COPYING

Today, SPDX's general approach seems to be to atomize single-file collections of disparate license texts rather than attempt to create "collection licenses" unique to particular projects. It is likely that any individual licenses in `Net-SNMP` that are not already on the SPDX license list will be found in old packages and thus may get individually added to the SPDX license list in the future. The approach taken in `Net-SNMP` is inconsistent with current SPDX practice, therefore, and is also in conflict with SPDX's mission to promote standardization of license identifiers.

jlovejoy:
fixes #2506 

### DIFF
--- a/src/Net-SNMP.xml
+++ b/src/Net-SNMP.xml
@@ -1,13 +1,16 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <SPDXLicenseCollection xmlns="http://www.spdx.org/license">
    <license isOsiApproved="false" licenseId="Net-SNMP" listVersionAdded="2.6"
-            name="Net-SNMP License">
+            name="Net-SNMP License" deprecatedVersion="3.25">
       <crossRefs>
          <crossRef>http://net-snmp.sourceforge.net/about/license.html</crossRef>
       </crossRefs>
-      <notes>This is the overall Net-SNMP license, which is comprised of several
-         licenses which are referred to in totality by the notices in
-         the source files.</notes>
+      <notes>DEPRECATED: The Net-SNMP license id represented the license stack 
+         of 9 licenses found at https://net-snmp.sourceforge.io/about/license.html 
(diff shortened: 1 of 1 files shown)


19714 (approved):

swinslow:
@jlovejoy I added 8d30dd5 just to fix "3.25" => "3.25.0" for the deprecation version. After the tests re-pass, I'll approve and merge this. Thanks!

swinslow (approved):

PR #2567, opened 2024-09-22
libutil-David-Nugent: Fix regex and spacing

### LINKED ISSUE #2566: Line break within <alt> tag regex in libutil-David-Nugent
In [`libutil-David-Nugent.xml`](https://github.com/spdx/license-list-XML/blob/main/src/libutil-David-Nugent.xml), there's a line break in the middle of an alt tag's regular expression:

https://github.com/spdx/license-list-XML/blob/b6d6eba54d0335e8fd0ed70bb8401f41a72e18b3/src/libutil-David-Nugent.xml#L44-L48

I think there are some things we may want to discuss more generally about what we do and don't include in regular expressions. But that's for a separate issue. Either way I suspect the line break here is not helpful and should be removed.

swinslow:
Fixes #2566 

### DIFF
--- a/src/libutil-David-Nugent.xml
+++ b/src/libutil-David-Nugent.xml
@@ -43,9 +43,7 @@
             <item>
                <bullet>4.</bullet>
                Absolutely no warranty of function or purpose is made
-               by the<alt name="author" match="authors|author Peter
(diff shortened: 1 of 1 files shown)


xsuchy (approved):
+1

PR #2573, opened 2024-09-27
New Exception: Independent-modules-exception

### LINKED ISSUE #2530: Classpath-exception-2.0 variant in Free Pascal
During Fedora License review
  https://gitlab.com/fedora/legal/fedora-license-data/-/issues/548
we found in package `fpc` (Free Pascal) an exception that is close to `Classpath-exception-2.0` and used in combination with LGPL-2.1-or-later.

This exception is identical to `Classpath-exception-2.0` except that it omits the first paragraph/sentence:

```
Linking this library statically or dynamically with other modules is making a combined work based on this library. Thus, the terms and conditions of the GNU General Public License version 2 cover the whole combination.
```

Should this section be marked as optional in current exception? Or this should be new id?

Full text of the license:
https://gitlab.com/freepascal.org/fpc/source/-/blob/release_3_2_2/rtl/COPYING.FPC

```
This is the file COPYING.FPC, it applies to the Free Pascal Run-Time Library 
(RTL) and packages (packages) distributed by members of the Free Pascal 
Development Team.

The source code of the Free Pascal Runtime Libraries and packages are 
distributed under the Library GNU General Public License 
(see the file COPYING) with the following modification:

As a special exception, the copyright holders of this library give you
permission to link this library with independent modules to produce an
executable, regardless of the license terms of these independent modules,
and to copy and distribute the resulting executable under terms of your choice,
provided that you also meet, for each linked independent module, the terms
and conditions of the license of that module. An independent module is a module
which is not derived from or based on this library. If you modify this
library, you may extend this exception to your version of the library, but you are
not obligated to do so. If you do not wish to do so, delete this exception
statement from your version.
```

jlovejoy:
fixes #2530


jlovejoy:
@swinslow @karsten-klein @Pizza-Ria - we didn't really decide on name, so I went with something similar to what @karsten-klein had, but a bit shorter on the id

open to suggestions!

src/exceptions/Classpath-exception-2.0.xml
```diff
@@ -5,6 +5,9 @@
          <crossRef>http://www.gnu.org/software/classpath/license.html</crossRef>
          <crossRef>https://fedoraproject.org/wiki/Licensing/GPL_Classpath_Exception</crossRef>
       </crossRefs>
+      <note>This is an exception from the FSF for Classpath. Independent-modules-exception
+         is very similar, but has a separate id due to it being used with LGPL and
+      a different lead-in that Classpath-exception-2.0.</note>
```
xsuchy, line 10: The tag should be 'notes' instead of 'note'.

xsuchy (approved):
+1 from me.
I will let other to give feedback on the name. 
If there will be no feedback I will merge it in a week.

PR #2574, opened 2024-09-27
Update Motosoto.xml

### LINKED ISSUE #1966: Motosoto license text is incomplete
The `Motosoto` license text is missing several paragraphs. It has probably been imported from <https://opensource.org/license/motosoto-php/>, and that source is incomplete, too. The original pull request, #134, already has this problem.

I assume the issue originates from an encoding problem with one character in the license text. Compare the Jabber Open Source License, version 1.0 on which `Motosoto` is said to be based: the [copy at OSI](https://opensource.org/license/jabberpl-php/) looks OK (perhaps offending character removed?), and the [copy in the FSF Directory](https://directory.fsf.org/wiki/License:JOSL-1.0) has a “Õ” in the place where the `Motosoto` text ends.

I have reported OpenSourceOrg/licenses#86 for this in the OSI licenses project.

jlovejoy:
fixes #1966 

Note, the XML file will look like there are more updates than there is, as  I ended up pulling it into another editor to fix the existing nested tables and add the missing text.

The numbering was also a challenge b/c the updates on OSI that is the source of the original issue here (https://github.com/OpenSourceOrg/licenses/issues/86 ) has been completed, but the numbering on the website at https://opensource.org/license/motosoto is not sequential

### DIFF
--- a/src/Motosoto.xml
+++ b/src/Motosoto.xml
@@ -553,7 +553,7 @@
             </item>
             <item>
                <bullet>11.</bullet>
-               Responsibility for Claims
+               Responsibility for Claims.
                <p>
                   As between Licensor and Contributors, each party is
                   responsible for claims and damages arising, directly
--- a/src/Motosoto.xml
+++ b/src/Motosoto.xml
@@ -4,415 +4,714 @@
       <crossRefs>
          <crossRef>https://opensource.org/licenses/Motosoto</crossRef>
       </crossRefs>
-    <text>
-      <titleText>
-         <p>MOTOSOTO OPEN SOURCE LICENSE - Version 0.9.1</p>
-      </titleText>
-
-      <p>This Motosoto Open Source License (the "License") applies to "Community Portal
-         Server" and related software products as well as any updatesor maintenance releases of that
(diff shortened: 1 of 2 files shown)


jlovejoy:
ugh, I have spent a bunch of time on this and can't see why it's failing right now!

xsuchy:
Do this and it will pass tests:
```
diff --git src/Motosoto.xml src/Motosoto.xml
index c45ea195..249197e0 100644
--- src/Motosoto.xml
+++ src/Motosoto.xml
@@ -553,7 +553,7 @@
             </item>
             <item>
                <bullet>11.</bullet>
-               Responsibility for Claims
+               Responsibility for Claims.
                <p>
                   As between Licensor and Contributors, each party is
                   responsible for claims and damages arising, directly
```

swinslow (approved):
Given the size of the diff, I can't really gauge what exactly had changed here, but it looks like the main non-formatting change is to add sections 11-15 and Exhibit A.

Since the tests are passing, I'm going to go ahead and approve + merge this. Thank you @jlovejoy!

PR #2578, opened 2024-10-04
Update RELEASE-HISTORY for 3.25.0

swinslow:
Whoops -- should probably have included this in the 3.25.0 release...

Signed-off-by: Steve Winslow <steve@swinslow.net>

PR #2579, opened 2024-10-04
Update release process to include history

swinslow:
Parallel to #2578 -- adding reminder as part of the release process


jlovejoy:
+1

Issue #2587, opened 2024-10-15
Spelling Correction


Pizza-Ria:
Correct spelling of university in https://spdx.org/licenses/xpp.html.

xsuchy:
There are two options specified using alt tag. One of them is grammatically correct.
https://github.com/spdx/license-list-XML/blob/main/src/xpp.xml#L48

But I agree that the one inside of the tag (and that is rendered in the documentation) should be the correct one.

swinslow:
Thanks @Pizza-Ria! Merged in #2588.

PR #2588, opened 2024-10-15
change spelling of "University" in xpp license

xsuchy:
Fixes: #2587
Both spelling are possible using the alt tag. So it is safe to swap it, because this does not change any matching. It just change the text of example in our documentation.


swinslow:
Good eye, thanks @xsuchy   :)  Merging now.

PR #2589, opened 2024-10-21
Update xml-fields.md

### LINKED ISSUE #2584: XML explanation re: use of ? in markup
Add another example of using `?` in a string to the existing example ("express(ed)") that is more complex, e.g. BSD-3-Clause and "specific" in #2529 

question mark applies to thing immediately before it, which could be a comma also 

(compare to optional tag - note can't have optional tag within alt tag)

jlovejoy:
fixes #2584 

need some input on a few things - see comments below

### DIFF
--- a/DOCS/xml-fields.md
+++ b/DOCS/xml-fields.md
@@ -166,9 +166,14 @@ The **<alt>** tag includes two attributes:
 
 Learning regular expressions is outside the scope of this document. However, here are a few of the most common patterns for the `match` attribute:
 * `<alt match="are|is">`: Matches if the text is either the word "are" or the word "is"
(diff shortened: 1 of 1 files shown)


jlovejoy:
@goneall @swinslow - this mentions `<obsoletedBys>` - but are we not using that any more? If so, should we remove here?

jlovejoy:
@swinslow can you check my explanation of `?` - especially in regard to the "specific" example :)

swinslow (approved):
LGTM, thanks @jlovejoy!

swinslow:
And `obsoletedBy` is still in the spec and some of the licenses, so I would keep it in the docs for now.

PR #2604, opened 2024-11-04
Fixed a typo in Gmsh-exception.xml

coky-t:
The ">" at the end of "Gmsh exception>" appears to be a typo.


xsuchy (approved):
+1
Good catch.

PR #2605, opened 2024-11-04
Update Mackerras-3-Clause-acknowledgment.xml

coky-t:
The current email address is listed from ppp-2.5.1 onwards, but the email address in the license is different up to ppp-2.5.0.

https://github.com/ppp-project/ppp/blob/ppp-2.5.1/pppd/auth.c#L20: paulus@ozlabs.org

https://github.com/ppp-project/ppp/blob/ppp-2.5.0/pppd/auth.c#L20: paulus@samba.org

### DIFF
--- a/src/Mackerras-3-Clause-acknowledgment.xml
+++ b/src/Mackerras-3-Clause-acknowledgment.xml
@@ -34,7 +34,7 @@
                <bullet>3.</bullet>
                Redistributions of any form whatsoever must retain the
                following acknowledgment: "This product includes software
-               developed by Paul Mackerras &lt;paulus@ozlabs.org&gt;".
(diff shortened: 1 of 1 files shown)


xsuchy:
Can you make the regexp smaller?

```
&lt;<alt match="(paulus@ozlabs.org|paulus@samba.org)&gt;"
```
Or even provide alt just for the domain name after `@`.

xsuchy:
Test suite fails with
```
 	Test for license Mackerras-3-Clause-acknowledgment failed: Test for license ID Mackerras-3-Clause-acknowledgment failed due to difference found Normal text of license does not match starting at line #17 column #17 "." when comparing to template text " Redistributions of any form whatsoever must retain the following acknowledgment: "This product incl..."
```

xsuchy:
Thank you for trying. I see that putting lt/gt outside of regexp somehome breaks test. So I do not mind now having it inside. Despite that it results in longer formula.

xsuchy (approved):
+1

PR #2616, opened 2024-11-10
Fix URLs to OSI website

ziemek99:
Apparently OSI webite was restructured a bit some time ago. Old links to license info return HTTP 301 redirects. While it's not a problem for a browser to follow such link, some kind of parser logic at SPDX erroneously detects such URL as "no longer live" when you see "Other web pages for this license" on SPDX website.

### DIFF
--- a/DOCS/xml-fields.md
+++ b/DOCS/xml-fields.md
@@ -48,7 +48,7 @@ There are two mandatory attributes for every `<license>` and `<exception>` tag:
 There are two additional highly encouraged attributes which are technically optional:
 * `listVersionAdded`: in which release version of the SPDX License List was the license first added, e.g., "3.24.0".
   * Typically you'll check the currently-released version at https://github.com/spdx/license-list-XML/releases and increment the minor version by 1 for a new license. E.g., when last release is "3.24.0" you add there "3.25.0".
-* `isOsiApproved` (for licenses, not exceptions): either "true" or "false" based on whether this license has been approved by the [Open Source Initiative](https://opensource.org/licenses/alphabetical)
+* `isOsiApproved` (for licenses, not exceptions): either "true" or "false" based on whether this license has been approved by the [Open Source Initiative](https://opensource.org/licenses)
 
 Finally, if the license ID has been deprecated, one additional attribute should be included:
 * `deprecatedVersion`: in which release version of the SPDX License List was the license first marked as deprecated, e.g. "3.24.0"
--- a/DOCS/history-OSI-SPDX.md
+++ b/DOCS/history-OSI-SPDX.md
@@ -34,9 +34,9 @@ See SPDX-legal email list from April 2012 [here](https://lists.spdx.org/g/Spdx-l
 Suggestion for OSI: Consider listing all licenses ever approved on OSI website to retain a full historical record (even if noted as superceded).
 
 ## Artistic License variations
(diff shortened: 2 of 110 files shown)


goneall:
@ziemek99 - Thanks for pointing this out and proposing a solution

Rather than replacing the current OSI reference, I would suggest we add a new reference and keep the original so that tools the use the URL's to correlate license information won't break.

We'll end up with one live and one "not so live" URL.  Note, this may appear unnecessary to human users, but for many tools it is quite important to correlate licenses with URLs that are no longer used.

ziemek99:
> Rather than replacing the current OSI reference, I would suggest we add a new reference and keep the original so that tools the use the URL's to correlate license information won't break.

I suppose it only applies to files in `src` directory and not `DOCS`?

> We'll end up with one live and one "not so live" URL.

Is there any required order to keep? For aesthetic reasons I'd like to keep "not so live" URLs under the live ones. If that'd break compatibility, though, I can keep the order of the previous entries and add current URLs underneath.

Another (better) solution would be fixing the parser logic so it follows any HTTP redirects and doesn't mark these URLs as no longer live.

goneall:
> > Rather than replacing the current OSI reference, I would suggest we add a new reference and keep the original so that tools the use the URL's to correlate license information won't break.
> 
> I suppose it only applies to files in `src` directory and not `DOCS`?

Correct - only the `src`

> 
> > We'll end up with one live and one "not so live" URL.
> 
> Is there any required order to keep? For aesthetic reasons I'd like to keep "not so live" URLs under the live ones. If that'd break compatibility, though, I can keep the order of the previous entries and add current URLs underneath.

I don't think order matters.

> 
> Another (better) solution would be fixing the parser logic so it follows any HTTP redirects and doesn't mark these URLs as no longer live.

Possibly - Although it is nice having the correct (non-redirected) URLs added for 2 reasons:
- Improves the correlation of license identifications since it will identify licenses with the new location
- The redirect may go away someday and having the "early warning" with the redirects may be helpful

Flagging these on the website gives us a chance to add the new URL.  There may be a better way to flag them, but this did result in an very welcome update to the `src` URLs.  Again - thanks for the PR and noticing the change.

DOCS/xml-fields.md
```diff
@@ -48,7 +48,7 @@ There are two mandatory attributes for every `<license>` and `<exception>` tag:
 There are two additional highly encouraged attributes which are technically optional:
 * `listVersionAdded`: in which release version of the SPDX License List was the license first added, e.g., "3.24.0".
   * Typically you'll check the currently-released version at https://github.com/spdx/license-list-XML/releases and increment the minor version by 1 for a new license. E.g., when last release is "3.24.0" you add there "3.25.0".
-* `isOsiApproved` (for licenses, not exceptions): either "true" or "false" based on whether this license has been approved by the [Open Source Initiative](https://opensource.org/licenses/alphabetical)
+* `isOsiApproved` (for licenses, not exceptions): either "true" or "false" based on whether this license has been approved by the [Open Source Initiative](https://opensource.org/licenses)
```
xsuchy, line 51: This should be just `https://opensource.org/license` (without the trailing s.

xsuchy:
Otherwise LGTM.

xsuchy:
@ziemek99 Can you update the PR with requested change pls?

swinslow:
Holding for 3.27.0, I'd like to take a closer look at this given the number of files it's touching.

xsuchy:
@ziemek99 Can you please update the PR?

khanbilal732 (approved):

nickvidal:
Fixed by #2960

Issue #2560, opened 2024-09-10
New license request: wwl [SPDX-Online-Tools]


xsuchy:
**1.** License Name: WWL License
**2.** Short identifier: wwl
**3.** License Author or steward: Unknown
**4.** Comments: This license was discovered during Fedora Linux license review in package wwl. https://gitlab.com/fedora/legal/fedora-license-data/-/issues/565
**5.** License Request Url: http://tools.spdx.org/app/license_requests/391
**6.** URL(s): http://www.db.net/downloads/wwl+db-1.3.tgz
**7.** OSI Status: Unknown
**8.** Example Projects: http://www.db.net/downloads/, https://src.fedoraproject.org/rpms/wwl/blob/rawhide/f/wwl.spec#%5C_5

xsuchy:
Text of the license:
```
db@FreeBSD.ORG wrote this file.  As long as you retain this notice you
can do whatever you want with this code, except you may not
license it under any form of the GPL.
A postcard or QSL card showing me you appreciate
this code would be nice. Diane Bruce va3db
```

Pizza-Ria:
Not allowing GPL is a restriction on use.  That seems to take it out of OSD. It potentially restricts other software licensed under the GPL that, if combined, would copyleft a component with this license.  Not a fan but I did think the ask for a card was cute.  If it was allowed, it should be the Hallmark license.

swinslow:
I don't think I'd view "you may not license this under the GPL" as a restriction. At least, not one in the sense of "use restrictions" we usually talk about. Arguably any GPL-incompatible license (e.g. Apache-2.0 or CDDL-1.1, if you believe they aren't GPL-compatible) cannot be licensed under the GPL, but would still be "open source" licenses.

(Arguably _any_ software under most non-GPL licenses cannot be simply "license[d]... under the GPL", but that's a rabbit hole I don't really want to go down here.)  :)

I view this as being roughly comparable to [Beerware](https://spdx.org/licenses/Beerware.html) (which it feels like it might've been based on). I'd be +1 to add it to the list given this and its inclusion in Fedora.

From a quick search I don't see this showing up in anything other than this wwl package. I'd be fine with `wwl` as the ID (probably lowercase, as that's how it appears to be rendered in the project?) and "wwl License" as the full name.

jlovejoy:
also used in Debian, see https://sources.debian.org/src/wwl/1.3%2Bdb-3/wwl.c/
[...]

PR #2620, opened 2024-11-15
add wwl license

xsuchy:
Fixes: #2560


xsuchy:
I was thinking about making the email and name `alt` but given that we have no proof of any alternative in use, I left it as it is. If we find any other usage we can add markup later.

Issue #2559, opened 2024-09-09
New license request: CGAL-linking-exception [SPDX-Online-Tools]


xsuchy:
**1.** License Name: CGAL Linking Exception
**2.** Short identifier: CGAL-linking-exception
**3.** License Author or steward: Unknown
**4.** Comments: This exception was found during Legal review in Fedora Linux https://gitlab.com/fedora/legal/fedora-license-data/-/issues/562 and it is used on OpenSCAD package.
**5.** License Request Url: http://tools.spdx.org/app/license_requests/390
**6.** URL(s): https://github.com/openscad/openscad/blob/openscad-2021.01/COPYING#L3
**7.** OSI Status: Unknown
**8.** Example Projects: https://github.com/openscad/openscad

xsuchy:
Text of the exception:
```
  As a special exception, you have permission to link this program
  with the CGAL library and distribute executables, as long as you
  follow the requirements of the GNU GPL in regard to all of the
  software in the executable aside from CGAL.
```

hroncok:
I belive this is [originally from OpenCSG](https://github.com/floriankirsch/OpenCSG/blob/opencsg-1-4-2-release/license.txt#L3), but it was since [changed there](https://github.com/floriankirsch/OpenCSG/commit/573cea0ccf3431a7a3d6f32733aa9a6682b41758).

karsten-klein:
**{metæffekt} Universe**
canonical name: [Linking Exception (CGAL)](https://github.com/org-metaeffekt/metaeffekt-universe/blob/main/src/main/resources/ae-universe/[l]/README.md#user-content-Linking-Exception-(CGAL):~:text=Linking%20Exception%20(CGAL))
short name: Linking-Exception-CGAL
category: Linking Exception CGAL
OSI status: _none_

jlovejoy:
+1 to add. 

I notice that at https://github.com/floriankirsch/OpenCSG/blob/opencsg-1-4-2-release/license.txt#L3  is include a parenthetical with the URL, which does not appear at https://github.com/openscad/openscad/blob/openscad-2021.01/COPYING#L3
Perhaps  we include that parenthetical with optional markup

jlovejoy:
name and ID seem fine to me - but would like others to weigh in

Pizza-Ria:
Sure - ok to add.

swinslow:
+1 to add, fine with name and ID

PR #2622, opened 2024-11-15
add CGAL-linking-exception exception

xsuchy:
Fixes: #2559


xsuchy:
The second instance of this instance has `library` instead of `program`. I added it as a possible alt.

PR #2626, opened 2024-12-04
Update AML.xml

coky-t:
Recent AML license text may have "Disclaimer:" at the beginning.
It also have "Apple Inc." instead of "Apple Computer, Inc.".
https://github.com/apple/ml-cvnets/blob/main/LICENSE

### DIFF
--- a/src/AML.xml
+++ b/src/AML.xml
@@ -9,21 +9,22 @@
          <p>Copyright: Copyright (c) 2006 by Apple Computer, Inc., All Rights Reserved.</p>
       </copyrightText>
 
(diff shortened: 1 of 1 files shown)


jlovejoy:
thanks @coky-t - we don't want to change the existing, old license though. Rather, we can add mark-up to accommodate these differences. The old "version" may still be out there.

I'm going to try to change your PR and if that doesn't work easily, I'll just make a new one.

jlovejoy:
NEVER MIND my comment above! That is what you did, the diff was just showing it like the entire paragraphs were replaced and I didn't look closely enough. Merging now

PR #2623, opened 2024-11-15
Add alternative signee to GPL-2.0-only

szepeviktor:
I've never sent a PR to SPDX, please be gentle with me.

From https://github.com/spdx/license-list-XML/issues/2568#issuecomment-2431316819

PR #2632, opened 2024-12-15
Add alternative signee to GPL licenses

szepeviktor:
From https://github.com/spdx/license-list-XML/pull/2623#issuecomment-2480565281

Reference: https://www.gnu.org/licenses/old-licenses/gpl-1.0.html

---

Is this a good start?

### DIFF
--- a/src/GPL-1.0-only.xml
+++ b/src/GPL-1.0-only.xml
@@ -354,7 +354,7 @@
         to make passes at assemblers) written by James Hacker.
       </p>
       <p>
-        &lt;signature of Ty Coon&gt;, 1 April 1989 Ty Coon, President of Vice
+        &lt;signature of <alt match="Ty Coon|Moe Ghoul" name="sample-signee">Ty Coon</alt>&gt;, 1 April 1989 <alt match="Ty Coon|Moe Ghoul" name="sample-signee">Ty Coon</alt>, President of Vice
       </p>
       <p>
         That's all there is to it!


xsuchy:
This is good start. It fact, it seems to be correct. But for some reason the test fails. I tracked it to the first `alt`, but I could not find what is the problem there. It seems correct to me, but the test thinks otherwise. :(

swinslow:
Thank you @szepeviktor and @xsuchy! I'll take a look and see if I can figure anything out from the failing tests.

swinslow:
I think the tests here are failing because of some combination of issues in the license matcher involving:
1. the `spacing` attribute for `<alt>` tags (see the XML schema [line 262](https://github.com/spdx/license-list-XML/blob/cb00a23b3b52d7cb20f5f3d2d69ff40c13f9c7f0/schema/ListedLicense.xsd#L262C1)); and
2. the `lt&;` / `gt&;` entities being used for the `<` / `>` characters in the signature block.

I believe the reference to line 70 in the test error message is because the above failures to match are causing the entire "END OF TERMS AND CONDITIONS" + Appendix `<optional>` block to fail to match. (Line 70 is the "END OF TERMS AND CONDITIONS" line in the test text file.)

From my understanding of the `spacing` attribute, I would expect that putting `spacing="before"` within both of the `<alt>` tags in this line ought to make the match work... but it seems to still be failing.

Removing the `&lt;` / `&gt;` entities from the XML, and the `<` / `>` characters from the test text, makes it pass successfully. But that doesn't seem to be the right answer either.

@goneall I'm afraid we may need your help on this one. I'm not certain whether I'm misunderstanding something in how the `spacing` attribute works; or whether this might be a bug in the license matcher.

In any case, I'm going to move this PR to 3.27.0. Once we get this issue sorted out, we can replicate this to the other L/A/GPL licenses as well. Thank you @szepeviktor and @xsuchy for your help on this so far!

goneall:
@swinslow - I'll take a look this week.  It could also be a bug in the license compare software.

goneall:
I've narrowed down the problem - but a solution would be somewhat difficult.

The problem is with the `<` and `>` characters before and after the match.  It looks like the tokenizer is not splitting these characters out as separate characters.  Normally, to solve this we would just move the characters into the match and alt text (e.g. `<alt match="<(Ty Coon|Moe Ghoul)>" name="sample-signee">&lt;Ty Coon&gt;</alt>`).  However, the `<` and `>` characters are special characters in our original template language and can't be used in the match pattern.

I can think of two possible solutions - none of them easy:
1. Drop support for the legacy template and require all the tools folks to use the now supported XML format.
2. Change the license matcher to tokenize the `<` and `>` characters to match the original

I've added an issue to the license list publisher for 2: https://github.com/spdx/Spdx-Java-Library/issues/269

It may be a while before I get to working on 2 - likely the easier solution.

goneall:
This has been fixed in https://github.com/spdx/spdx-java-core/pull/31

I'm working on new releases of the library and publisher.  Once we update the versions including the license list publisher version used in the CI, this should resolve the test failure.

xsuchy:
Closing in favour of merged https://github.com/spdx/license-list-XML/pull/2954

szepeviktor:
Thank you.

PR #2636, opened 2024-12-17
Update license list publisher to version 3.0.0

goneall:
This version of the publisher generates SPDX spec version 3.0 compliant licenses.

See https://github.com/spdx/LicenseListPublisher/issues/183 for a detailed description of the changes for the SPDX 3 license list support.

### DIFF
--- a/.github/workflows/validate.yaml
+++ b/.github/workflows/validate.yaml
@@ -24,9 +24,10 @@ jobs:
     name: Validate canonical match
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/setup-java@v1
+      - uses: actions/setup-java@v3
         with:
-          java-version: '11'
+          distribution: 'temurin'
+          java-version: '17'
       - uses: actions/checkout@v4
       - name: Fetch changes for git diff
         id: fetchdiff
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -17,7 +17,7 @@ jobs:
       - uses: actions/setup-java@v3
         with:
           distribution: 'temurin'
-          java-version: '11'
+          java-version: '17'
       - uses: actions/checkout@v4
       - name: Fetch changes for git diff
         id: fetchdiff
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 # SPDX-License-Identifier: CC0-1.0
 
-TOOL_VERSION = 2.2.10
+TOOL_VERSION = 3.0.0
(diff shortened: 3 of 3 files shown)


.github/workflows/validate.yaml
```diff
@@ -24,9 +24,10 @@ jobs:
     name: Validate canonical match
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/setup-java@v1
+      - uses: actions/setup-java@v3
         with:
-          java-version: '11'
+          distribution: 'temurin'
```
xsuchy, line 29: Can you elaborate why the choice of `temurin`? I do not mind. V1 used zulu. And I am just curious if the other one is better or what?

xsuchy (approved):
+1

xsuchy:
Hmm, I think I was too fast. When I merged this one I get:
```
$ ./test-one-license BSD-3-Clause
Testing test/simpleTestForGenerator/BSD-3-Clause.txt
SLF4J: No SLF4J providers were found.
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See https://www.slf4j.org/codes.html#noProviders for further details.
Invalid license SPDX definition: Unknown spec version 3.0.1
```

.github/workflows/validate.yaml
goneall: Primarily because we are using termin for other SPDX project JDKs in the CI - just removing a variable for things that may fail when deploying the license list publisher.

A bit more background:

I switched over to termurin years ago for the other SPDX Java projects when there were some concerns about OpenJDK licensing and support - I don't think those concerns are still valid BTW - so I guess this is mostly a historical decision.  I haven't really researched zulu - I just did the update when I needed to move the Java support level up in the CI.

goneall:
I was able to duplicate the problem locally - definitely a problem with the `licenseListPublisher`.

I have a quick fix in progress which I could release within the hour, but I'm thinking of adding 2 additional fix / features:
- Add a unit test so we can find issues with testing a single license before release
- Add support for testing multiple files

@xsuchy - let me know if you're OK waiting until tomorrow at the latest for an update.  It looks like the errors only affect the test single license command.

xsuchy:
No problem, especially if it is broken just for me. Though I am afraid GH CI may fail too.

FELLIX000 (approved):

FELLIX000 (approved):

FELLIX000 (approved):

Deymon13 (approved):

PR #2645, opened 2024-12-28
MIT-open-group: Add missing name alt match

castholm:
>IN NO EVENT SHALL [THE OPEN GROUP] BE LIABLE [...]

"THE OPEN GROUP" was missing from the matching rules, but it is present in all of the linked licenses.

I also removed the link to the xvinfo repo since that repo is licensed under X11, and never appears to have been licensed under MIT-open-group.


xsuchy (approved):
+1

PR #2646, opened 2024-12-28
AFL-2.1: Add optional appendix

castholm:
[dbus](https://gitlab.freedesktop.org/dbus/dbus) is probably one of the most prolific projects licensed under AFL-2.1 and [their original AFL 2.1 license text](https://gitlab.freedesktop.org/dbus/dbus/-/blob/789d97ad53044f94351a958e785c2923643b346a/COPYING#L11-207) had an appendix (similar to the last part of GPL licenses) which ironically was lost when they made their repository REUSE-compliant and used REUSE's tooling to replace the license with the license text downloaded from this very repository!

I replaced the test license text with the license text from dbus, since it appears to be the most complete reproduction of the original AFL 2.1 license text. Apologies in advance to any maintainers for the change in text wrapping.

Additionally, both the dbus license text and the text in https://web.archive.org/web/20061013013632/http://opensource.linux-mirror.org/licenses/afl-2.1.txt has a space between the "v." and the "2.1". But because the license text from this repo has already proliferated, we can't correct the mistake outright and have to accept both a space and no space.

### DIFF
--- a/src/AFL-2.1.xml
+++ b/src/AFL-2.1.xml
@@ -9,7 +9,7 @@
     <text>
       <titleText>
          <p>The Academic Free License
-        <br/>v.2.1
+        <br/>v.<alt match=" ?" name="extraSpace"> </alt>2.1
       </p>
       </titleText>
       <optional>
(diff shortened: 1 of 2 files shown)


jlovejoy:
@castholm - sorry! This got a bit lost in the shuffle. Did you make an issue explaining this or just this PR?

I think this is ONLY adding the explanation part after the END OF LICENSE - this text was not in the original AFL-2.1, but was and is included in dbus, so you want to add it here as optional so the dbus variant will still match to AFL-2.1.- is that right?

I'm not sure this is necessary, as it's clearly not part of the license given it's after the END OF LICENSE text??

swinslow:
Discussed on 2026-05-14 legal team call. Agreed to add this section as `<optional>` language. Appears that the tests are passing so I'll go ahead and merge this, thank you @castholm and all!

PR #2648, opened 2024-12-30
Revert expected-warnings change from 166129d

### LINKED ISSUE #2468: Figure out why HPND-sell-variant* is matching incorrectly
In 3.24, HPND-sell-variant and HPND-sell-variant-MIT-disclaimer-rev were coming up as being duplicate licenses, though it didn't seem as though they should be matching.

See discussion at: https://github.com/spdx/license-list-XML/pull/2378

We are likely going ahead and merging #2378 once it's done in order to get it in for 3.24, but in 3.25 we should revisit this to see why these are coming up as matches. I suspect it's something to do with HPND-sell-variant's regexes, but I haven't had a chance to dig in yet.

swinslow:
In 166129d (as part of PR #2378), due to failing tests we added expected warnings to fix a message that HPND-sell-variant matched to HPND-sell-variant-MIT-disclaimer-rev.

These two licenses do not match, so this warning should not be appearing. This commit reverts the earlier change to reflect this.

Fixes #2468 (hopefully)


qoqnoos1:
ههع

Waham2000 (changes requested):
expected-warnings

Issue #2266, opened 2023-11-24
LicenseCompareHelper: BSD-3-Clause text not recognized depending on new lines and spaces


sdheh:
I found a BSD-3-Clause text where newlines make a difference to the results of LicenseCompareHelper.isTextStandardLicense:
Text 1:
```
Redistribution and use in source and binary forms, with or without
modification, are permitted provided that the following conditions
are met:

  - Redistributions of source code must retain the above copyright
    notice, this list of conditions and the following disclaimer.

  - Redistributions in binary form must reproduce the above copyright
    notice, this list of conditions and the following disclaimer in the
    documentation and/or other materials provided with the distribution.

  - Neither the name of the Eclipse Foundation, Inc. nor the names of its
    contributors may be used to endorse or promote products derived
    from this software without specific prior written permission.

THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS
IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO,
THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL THE COPYRIGHT OWNER OR
CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF
LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
```

Text 2:
```
Redistribution and use in source and binary forms, with or without
modification, are permitted provided that the following conditions
are met:

  - Redistributions of source code must retain the above copyright
    notice, this list of conditions and the following disclaimer.

  - Redistributions in binary form must reproduce the above copyright
[...]

Issue #1617, opened 2022-09-02
Matching guidelines: ignore length or number of underscores


jlovejoy:
Match the number of underscores to the test file for the form included in this license.

Note - we may want to update the license matching guidelines and the license matching software to cover this case.

(see #1594 )

jlovejoy:
@goneall - I think I created this due to one of the PRs you had. Is this solely a matter of updating the text of the matching guidelines or is there some tooling changes that need to go along with it?

goneall:
@jlovejoy If we update the matching guidelines to allow for any number of underscores, I will need to update the SPDX matching tools as well.  If/when we update the matching guidelines, we should add an issue to the [Spdx Java Library](https://github.com/spdx/spdx-java-library) to also allow any number of underscores.

goneall:
In terms of the exact guidelines text, I have a few considerations and recommendations:

1. Do we ignore the characters after the first, second or 3rd repeating characters? In other words is ` _ ` the same as `___`?
2. Do we include other characters in this guidelines (e.g. `=` and `-` are commonly used as separators - not to mention all the UTF variations on underscores)?
3. Do we want to allow for one type of line separator character to match others (e.g. `----` will match `____`)?

For 1. - I would recommend past the 3rd, but past the 2nd would be fine as well.  I wouldn't want to match one to more than one since the intent of the single underscore may not be a separator.
For 2. - As long as we do more than 2 characters, I think it is safe to include the following which I've seen used as line separators:

- [Dashes](https://en.wikipedia.org/wiki/Dash)
- [Underscores](https://en.wikipedia.org/wiki/Underscore)
- [Equals](https://en.wikipedia.org/wiki/Equals_sign)

For 3. - I would prefer we do not include matching of different line separator types - a bit more difficult to implement and I don't think it is that common - but I'm open to including it if others feel different

zvr:
I would suggest, that, similar to our existing matching guidelines:

> 5.1.2 Guideline:  Hyphens, Dashes  Any hyphen, dash, en dash, em dash, or other variation should be considered equivalent.

> 7.1.1 Guideline:  Where a line starts with a bullet, number, letter, or some form of a list item (determined where list item is followed by a space, then the text of the sentence), ignore the list item for matching purposes. Templates do not include markup for this guideline.

we should introduce a markup like `<separator>` (similar to `<bullet>`) that should be matched to a series of same characters like `-`, `_`, `=`, `*`, etc.

[...]

PR #2669, opened 2025-02-28
Match official Apache-1.1 text

goneall:
The official Apache 1.1 license text on the website has line breaks in different words.

This PR allows for line breaks on the matching regex for the alt text.

Reference https://github.com/spdx/Spdx-Java-Library/issues/230

### DIFF
--- a/src/Apache-1.1.xml
+++ b/src/Apache-1.1.xml
@@ -45,7 +45,7 @@
       </item>
          <item>
             <bullet>5.</bullet>
-        <alt match="(Products derived from this software may not be called\s.+nor may\s.+appear in their name, without prior written permission of\s.+.|Products may not include\s.+in their name, without prior written permission of\s.+.)" name="Clause5">Products derived from this software may not be called "Apache" 
+        <alt match="(Products\s+derived\s+from\s+this\s+software\s+may\s+not\s+be\s+called\s+.+nor may\s+.+appear\s+in\s+their\s+name,\s+without\s+prior\s+written\s+permission\s+of\s+.+.|Products\s+may\s+not\s+include\s+.+in\s+their\s+name,\s+without\s+prior\s+written\s+permission\s+of.+)" name="Clause5">Products derived from this software may not be called "Apache" 
 		nor may "Apache" [ex. the names] appear in their name, 
 		without prior written permission of the Apache Software Foundation.</alt>
       </item>


xsuchy:
Why? If LicenseCompareHelper has some issues then LicenseCompareHelper should be fixed. Because according
  https://spdx.github.io/spdx-spec/v3.0.1/annexes/license-matching-guidelines-and-templates/#whitespace
this should not be needed.

xsuchy:
Ah, it is in "match" attribute. But still - the tool should be fixed. Otherwise we should do the same on zillion other places too:
* src/FSFULLRWD.xml has "match="(as a special exception the author)|(...
* src/GFDL-1.1-or-later.xml has "match="the Back-Cover Texts being .+|...
* src/GFDL-1.2-only.xml has "match="the Invariant Sections...
* etc.

swinslow:
@goneall This is actually a question I've been meaning to ask you  :)  about whether the Matching Guidelines' "ignore whitespace" should apply to `match` attributes' regexes in `<alt>` tags as @xsuchy noted.

On the one hand, in an ideal world I'd like to be able to say "yes, the regexes should just disregard whitespace by default like everything else."

But, I suspect most existing regex engines are not capable of applying a rule like that by default. E.g., I know in [Python's `re` package](https://docs.python.org/3/library/re.html#regular-expression-syntax) that I can use a flag [`re.IGNORECASE`](https://docs.python.org/3/library/re.html#regular-expression-syntax) to perform case-insensitive matching. I don't think there's any equivalent flag for "treat all whitespace as a single whitespace character."

So the options would seem to be:

1. either require that `<alt>` expressions only match to _exact_ matches with the regex;
2. or sprinkle lots of `\s+` characters throughout the regexes on the license list.

This PR is proposing option 2. I've also seen this picked up in recent years in other long regexes, particularly in [`BSD-3-Clause`](https://github.com/spdx/license-list-XML/blob/97c98369206d21ea01284e66fd2ea06b64889d63/src/BSD-3-Clause.xml#L31) (see #2266 and the discussion I now realize you both had in that thread, so perhaps my whole comment here was unnecessary).  :)

In any case, I _think_ I'm +1 to merge this. But if so, then we should probably make a decision that it's worth converting other `<alt>` tags to use `\s+` for all spaces globally in the license list. I don't see that as a blocker to merging this one, but it looks like we may need to move in that direction overall.

goneall:
I may be reading the spec a bit too literally - I've interpreted the match to be a pure POSIX regex with a few documented exceptions.  In general, I don't think we should apply the license matching guidelines to the match regexes themselves - it would make for a very tricky transformation if we had to include things like punctuation, varietal word spellings, quotes, etc.

From the XML schema:

```
<xhtml:code>match</xhtml:code> is the <xhtml:a href="http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap09.html#tag_09_04">POSIX extended regular expression (ERE)</xhtml:a> defining allowed matches.  Where possible, alternates are listed in order of decreasing numbers of matched characters, which allows for regular expression engines using both POSIX's <xhtml:a href="http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap09.html#tag_09_01_02">longest-match</xhtml:a> semantics and those which prefer the leftmost branch (e.g. <xhtml:a href="https://golang.org/pkg/regexp/#Compile">Go</xhtml:a>, <xhtml:a href="https://www.ecma-international.org/publications/files/ECMA-ST/Ecma-262.pdf#sec-alternative.cE">JavaScript</xhtml:a>, <xhtml:a href="http://perldoc.perl.org/perlre.html#Combining-RE-Pieces">Perl</xhtml:a>, <xhtml:a href="https://docs.python.org/3.5/library/re.html#regular-expression-syntax">Python</xhtml:a>, etc.)
```

From the legacy match text:

```
match: a POSIX extended regular expression (ERE).
This field is required for a rule type: <<var; . . . >>
The [POSIX ERE](http://pubs.opengroup.org/onlinepubs/9699919799/) in the match field has the following restrictions and extensions:

Semicolons are escaped with \;
POSIX Bracket Extensions are not allowed
For example: <<var;name=organizationClause3;original=the copyright holder;match=.+>>
```

@xsuchy makes a very good point as to the practicality of manually fixing all the spaces in all the regexes.  I'm OK adding this one exception to the straightforward REGEX pattern as long as we document it in the schema and legacy match text.

I would also like to run a test with the change to make sure it doesn't break anything for the existing license XML patterns.

pmonks:
I noted what I think is a related issue in [this comment on #1617](https://github.com/spdx/license-list-XML/issues/1617#issuecomment-1297878419), albeit that was focused on punctuation rather than whitespace (to my mind they're two examples of the same underlying issue).  Perhaps looking at both whitespace and punctuation (and the examples I posted in that comment) will help identify a comprehensive solution?

goneall:
I ran some tests and this alt text tag fails if I implement the suggested change to the license matching software:

https://github.com/spdx/license-list-XML/blob/97c98369206d21ea01284e66fd2ea06b64889d63/src/Condor-1.1.xml#L61

I can think of fixes to the regex, but there may be more failures.

goneall:
I recommend we go ahead and merge this PR for the following reasons:
- The precedence for having `\s` in regexes has already been set in previous XMLs (MIT.xml, Autoconf-exception-generic.xml, BSD-3-Clause.xml).
- Making the change proposed above to automatically add the `\s+` will break at least one of the matches
- This PR will fix the issue where it won't match one of the "official" Apache 1.1 texts
- The change will be forward compatible if we do make the change to replace all spaces with `\s+` in regexes in the future

@xsuchy @swinslow @pmonks - let me know if you agree

goneall:
One thing this discussion illustrates is that the spec is not very clear on if the `alt` text regexes are to be applied to the original text or after some transformation to the regex and/or matching text to take into account the full license matching guidelines.  Perhaps we should open a separate issue to agree and document this.  As mentioned above, I feel very strongly that doing the transformations is not implementable - several have tried to create generic regexes to implement all the license matching guidelines and none have succeeded.  Having a limited set of documented exceptions to ease the development of the license XML may be reasonable, however.

xsuchy:
There are two occurrences of `\s+.+` . This is no needed. As previous `.+` or '\s.+` will match the same text.

> several have tried to create generic regexes to implement all the license matching guidelines and none have succeeded.

Sigh. That means I have to try too. :) I will try to find time for that.

> Making the change proposed above to automatically add the \s+ will break at least one of the matches

??? Do you have example?

> let me know if you agree

I agree that there is already precedence. With all my eyes closed, I can agree with this PR. But it is ugly, ugly, ugly.

goneall:
> I agree that there is already precedence. With all my eyes closed, I can agree with this PR. But it is ugly, ugly, ugly.

Thanks @xsuchy - we can always implement a better solution later - I'll go ahead and merge this so we have a solution in the near term.

> Sigh. That means I have to try too. :) I will try to find time for that.

A bit of history - we've had two summer of code student spend all summer trying to create such a solution.  My only requirement for the solution is that it completed in a reasonable amount of time (beware of runaway regexes) and that it pass the test for matching all SPDX licenses - neither met those requirements.

In addition, my first attempt to match licenses was using regexes - I gave up after a couple weeks and created a more sophisticated parser.

I was trying to find the repo on the last GSoC attempt, but I couldn't find it - I remember it was written in Python.  The repo must have been deleted.  They got pretty far, but couldn't pass all the tests.

I do use Regexes to narrow down the search for licenses - not good enough for solid matching, but it can give you a head-start.  The code is in Java here: https://github.com/spdx/Spdx-Java-Library/blob/2cd0d769a93640078542cf3b6f1045bb05cf1119/src/main/java/org/spdx/utility/compare/TemplateRegexMatcher.java#L225

PR #2678, opened 2025-03-06
add more license urls

joerg1985:
Here are some more URL mappings, please add them to your lists.

`https://www.eclipse.org/legal/epl-v20.html` used by `org.junit.jupiter:junit-jupiter-api:5.12.0`
`https://projects.eclipse.org/license/epl-2.0` used by `jakarta.json:jakarta.json-api:2.1.3`
`http://opensource.org/licenses/MIT` used by `org.checkerframework:checker-qual:3.49.1`

### DIFF
--- a/src/EPL-2.0.xml
+++ b/src/EPL-2.0.xml
@@ -5,6 +5,8 @@
 		<crossRefs>
 			<crossRef>https://www.eclipse.org/legal/epl-2.0</crossRef>
 			<crossRef>https://www.opensource.org/licenses/EPL-2.0</crossRef>
+			<crossRef>https://www.eclipse.org/legal/epl-v20.html</crossRef>
+			<crossRef>https://projects.eclipse.org/license/epl-2.0</crossRef>
 		</crossRefs>
     <notes>
       <p>Secondary Licenses declared via Exhibit A should be represented using the disjunctive OR operator (See: SPDX spec, section on SPDX License Expressions and https://www.eclipse.org/legal/epl-2.0/faq.php for more info).</p>
--- a/src/MIT.xml
+++ b/src/MIT.xml
@@ -3,6 +3,7 @@
    <license isOsiApproved="true" licenseId="MIT" name="MIT License">
       <crossRefs>
          <crossRef>https://opensource.org/license/mit/</crossRef>
+         <crossRef>http://opensource.org/licenses/MIT</crossRef>
       </crossRefs>
     <text>
       <titleText>


src/MIT.xml
```diff
@@ -3,6 +3,7 @@
    <license isOsiApproved="true" licenseId="MIT" name="MIT License">
       <crossRefs>
          <crossRef>https://opensource.org/license/mit/</crossRef>
+         <crossRef>http://opensource.org/licenses/MIT</crossRef>
```
xsuchy, line 6: they are the same url (the content is the same).

src/EPL-2.0.xml
```diff
@@ -5,6 +5,8 @@
 		<crossRefs>
 			<crossRef>https://www.eclipse.org/legal/epl-2.0</crossRef>
 			<crossRef>https://www.opensource.org/licenses/EPL-2.0</crossRef>
+			<crossRef>https://www.eclipse.org/legal/epl-v20.html</crossRef>
+			<crossRef>https://projects.eclipse.org/license/epl-2.0</crossRef>
```
xsuchy, line 9: This is just redirect to the url we have above.

xsuchy:
Closing as spam.

joerg1985:
@xsuchy some more context on this, i am trying to add them, to allow the `https://github.com/spdx/spdx-maven-plugin` to detect them. The plugin does not resolve the URLs and should not, as the plugin might run in a offline environment. 

@goneall is this the correct repo?

xsuchy:
> some more context on this, i am trying to add them, to allow the https://github.com/spdx/spdx-maven-plugin to detect them.

Can you reference the issue in spdx-maven-plugin so it is clear what you are trying to achieve? I still have no clue.

joerg1985:
@xsuchy sure this is the issue comment: https://github.com/spdx/spdx-maven-plugin/issues/180#issuecomment-2680231286
In an nutshell: The spdx-maven-plugin is used to generate a SBOM out of your maven POM, the mapping of licenses to SPDX identifiers is done using the URLs. So i am trying to add new URLs used in different POMs to make the mapping work.

xsuchy:
Ah. I see.
But the URL matching method is wrong. Very wrong. We cannot have full list of all URL where is the license used or even defined. Not even full, not even comprehensive. And I am afraid not even representative in some cases.

goneall:
A bit more history and context for the license URLs.

There's been quite a bit of discussion on the use of the URLs and in the past we've agreed it can be used to match licenses.  This is what led to the policy of not deleting URLs that are no longer live.

We've enhanced the license list publisher to check the status of the URLs.  It will check if the URL is live, if it is a wayback link, and if it actually contains text that matches the license per the license matching guidelines.  All of this information is recorded in the `crossRefs` field in the JSON (e.g. https://github.com/spdx/license-list-data/blob/d5f85740b2cd68f84672743987bc02c1408c48a0/json/details/EPL-2.0.json#L10)

When we render the HTML for spdx.org/licenses, we retain the order so the first license should always be the closest to canonical.  If the license is no longer live - we append the url with a "[not live]" note.

Since merging this will enhance the license data captured by the SPDX Maven plugin, I would like to merge this PR.

@xsuchy - let me know if you'd like to discuss prior to merging.

swinslow:
Discussed on 2025-03-13 legal team call. Agreed that we should add these, after Steve reviews. 

Will separately update documentation to better clarify that (1) we'll accept contributed PRs with accurate "other URLs" for licenses including where e.g. OSI has changed their website URLs; but (2) we -- the SPDX legal team participants -- won't be going out seeking to keep them up to date on our own.

swinslow (approved):

PR #2695, opened 2025-03-31
Add new cross-reference to Apache-2.0.xml

warting:
Add a new cross-reference to the Apache-2.0 license XML file.

* Add `<crossRef>https://opensource.org/license/apache-2-0</crossRef>` under the `<crossRefs>` section in `src/Apache-2.0.xml`

---

For more details, open the [Copilot Workspace session](https://copilot-workspace.githubnext.com/spdx/license-list-XML/pull/2695?shareId=8879add6-6cc2-446f-bb71-20f182b6b925).

### DIFF
--- a/src/Apache-2.0.xml
+++ b/src/Apache-2.0.xml
@@ -4,6 +4,7 @@
       <crossRefs>
          <crossRef>https://www.apache.org/licenses/LICENSE-2.0</crossRef>
          <crossRef>https://opensource.org/licenses/Apache-2.0</crossRef>
+         <crossRef>https://opensource.org/license/apache-2-0</crossRef>
       </crossRefs>
       <notes>This license was released January 2004</notes>
     <text>


jlovejoy:
ugh, so this is a problem created by the OSI changing URLs, instead of sticking to the policy they adopted in 2011 to use SPDX ids as their URLs. This is not really an SPDX problem, but we keep getting issues on this (hence we made a label).

Based on SPDX policies and processes, @warting - is correct to add a new URL (and not remove old ones that are no longer active). That policy was adopted for situations when a URL really did no longer work, but this (in my view) is a bit different since it's a URL for the OSI, which should not change as it creates problems for the many people who rely on it. I'd rather OSI fix this at the source, rather than us having to keep add new URLs b/c they have not been consistent to the original 2011 policy and adoption of SPDX ids.

@mccoysmith @jberkus @carlopiana - can someone at OSI help with the "non-sexy" work here?

xsuchy:
Thank you both for an explanation. I agree that OSI should make up a mind what URL they will use and stick to that, but at the same time I do not want to block merging and waiting for OSI as it can take ages.
So +1 from me.

PR #2680, opened 2025-03-07
Some-request

Rangawhenua:
The KAITIAKI License is a sovereignty-based licensing framework
            for Te Tūtohinga Matihiko. It ensures all data, technology, and
            digital taonga remain under Māori governance and cannot be used,
            modified, or distributed without adherence to tikanga.

PR #2696, opened 2025-04-03
validate

Rangawhenua:


xsuchy:
Still apply to what I said in
https://github.com/spdx/license-list-XML/pull/2681#issuecomment-2758745245
and 
https://github.com/spdx/license-list-XML/pull/2680

Reporting as spam.

PR #2707, opened 2025-04-17
Update matching guidelines to match 2 consecutive hyphens

### LINKED ISSUE #2704: Several Creative Commons licenses are not matching the "official text" on the website
This is due to the official text containing two dashes `--` whereas the license XML text contains only a single dash `-`.

The current license matching guidelines covers line separators of 3 or more dashes and single dashes, but does not specify if two dashes should be equivalent to a single dash.

One example is the CC-BY-4.0 official text:  https://creativecommons.org/licenses/by/4.0/legalcode.txt

The same problem occurs with the following:

* `CC-BY-SA-4.0`
* `CC-BY-NC-4.0`
* `CC-BY-NC-SA-4.0`
* `CC-BY-ND-4.0`
* `CC-BY-NC-ND-4.0`

Not that this may not be all the licenses with this issue.

goneall:
Fixes #2704

### DIFF
--- a/DOCS/license-matching-guidelines-and-templates.md
+++ b/DOCS/license-matching-guidelines-and-templates.md
@@ -109,7 +109,7 @@ Punctuation should be matched, unless otherwise stated in these guidelines or un
 
 ### 5.3 Guideline: hyphens, dashes <a name="5.3"></a>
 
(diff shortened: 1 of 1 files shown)


xsuchy:
I think this is equivalent to previous wording. But the new wording is definitely more clear. 
+1

Issue #2674, opened 2025-03-04
Markup change to cryptsetup-OpenSSL-exception


xsuchy:
During Fedora License review https://gitlab.com/fedora/legal/fedora-license-data/-/issues/570 we found license that has exception that is match to https://spdx.org/licenses/cryptsetup-OpenSSL-exception.html

  https://github.com/ocaml-omake/omake/blob/master/LICENSE.OMake#L20

There are minor changes. If you agree I will create PR with markup change to accommodate this variety.

![Image](https://github.com/user-attachments/assets/bb4330ca-113b-462e-927c-6f41d6e039ad)

jlovejoy:
yes @xsuchy - this is an easy add-markup case, thanks!

PR #2716, opened 2025-04-30
markup change for cryptsetup-OpenSSL-exception

xsuchy:
Fixes: #2674

### DIFF
--- a/src/exceptions/cryptsetup-OpenSSL-exception.xml
+++ b/src/exceptions/cryptsetup-OpenSSL-exception.xml
@@ -8,12 +8,13 @@
          <crossRef>https://github.com/nbs-system/naxsi/blob/951123ad456bdf5ac94e8d8819342fe3d49bc002/naxsi_src/naxsi_raw.c</crossRef>
          <crossRef>http://web.mit.edu/jgross/arch/amd64_deb60/bin/mosh</crossRef>
          <crossRef>https://sourceforge.net/p/linux-ima/ima-evm-utils/ci/master/tree/src/evmctl.c#l30</crossRef>
+         <crossRef>https://github.com/ocaml-omake/omake/blob/master/LICENSE.OMake#L20</crossRef>
       </crossRefs>
       <text>
          <p>
             <alt name="addition" match="In addition, as|As">In addition, as</alt> a special exception, the copyright
             holders give permission to link the code of portions
-            of this program with the OpenSSL library under certain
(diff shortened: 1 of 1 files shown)


xsuchy:
Checking the CI failure. The alt-match "OpenSSL library|Objective Caml runtime" does not match "OpenSSL\nlibrary". Is this bug in matching code? cc @goneall 
Because \n and " " should be treated identically. https://spdx.github.io/spdx-spec/v3.0.1/annexes/license-matching-guidelines-and-templates/#whitespace

src/exceptions/cryptsetup-OpenSSL-exception.xml
```diff
@@ -8,12 +8,13 @@
...
          <crossRef>http://web.mit.edu/jgross/arch/amd64_deb60/bin/mosh</crossRef>
          <crossRef>https://sourceforge.net/p/linux-ima/ima-evm-utils/ci/master/tree/src/evmctl.c#l30</crossRef>
+         <crossRef>https://github.com/ocaml-omake/omake/blob/master/LICENSE.OMake#L20</crossRef>
       </crossRefs>
       <text>
          <p>
             <alt name="addition" match="In addition, as|As">In addition, as</alt> a special exception, the copyright
             holders give permission to link the code of portions
-            of this program with the OpenSSL library under certain
+            of this program with the <alt name="library" match="OpenSSL library|Objective Caml runtime">OpenSSL library</alt> under certain
```
goneall, line 17:
```suggestion
            of this program with the <alt name="library" match="OpenSSL\s+library|Objective\s+Caml runtime">OpenSSL library</alt> under certain
```

goneall (changes requested):
The error is due to line breaks - using `\s+` instead of spaces should fix it.

xsuchy:
Ah, that's that issue. Nod.

I updated this PR with \s+

swinslow:
Thanks @goneall and @xsuchy.

@goneall I think I'm probably in favor of option 1 there. Primarily because I do think that most people would expect to be able to use the regexes as "standard" regexes as-is. So having them visible with the `\s+` markers in the license-list-XML repo is probably a good idea.

We're likely going to be pushing the next release in the next week (may be next weekend), so perhaps this would be good to look at incorporating after we get past that -- rather than adding it in just prior to the release. Happy to chat about specifics for how we might implement this.

@xsuchy this PR looks fine, and I'll go ahead and merge -- thank you.

swinslow (approved):

PR #2729, opened 2025-05-10
New License: libpng-1.6.35

### LINKED ISSUE #2591: libpng changes over the years
see https://gitlab.com/fedora/legal/fedora-license-data/-/issues/296

http://www.libpng.org/pub/png/src/libpng-LICENSE.txt is indeed a match to `libpng-2.0` (and the same URL is referenced in the SPDX entry, even)

as for the second part/license:
SPDX added https://spdx.org/licenses/Libpng.html in 2010 for version 1.4 of the SPDX License List (see https://lists.spdx.org/g/spdx/message/236?p=%2C%2C%2C20%2C0%2C0%2C0%3A%3Acreated%2C%2Clibpng%2C20%2C2%2C0%2C22079229 and discussion prior at https://lists.spdx.org/g/spdx/topic/22079227#msg235

SPDX added what existed on that same URL at that point in time. 
BUT over the years, various things were updated such as the list of names and references to version numbers (which would be replaceable by SPDX matching guidelines, but need markup added)
Notably, it was in 2016 that some additional text was added or changed more significantly: 
- https://github.com/pnggroup/libpng/commit/f78d1cd47ad2fb26f8435b82bedc2470dc6ae357#diff-c693279643b8cd5d248172d9c22cb7cf4ed163a3c98c8a3f69c2717edd3eacb7
- https://github.com/pnggroup/libpng/commit/08bd7654bc15ee53a6f3c6199aedfab636027588#diff-c693279643b8cd5d248172d9c22cb7cf4ed163a3c98c8a3f69c2717edd3eacb7
- https://github.com/pnggroup/libpng/commit/c36c09f00426a07fef5c6d6a7361c12a47e6439b

`libpng-2.0` was added in July 2018, which also altered some of the text order, see https://github.com/pnggroup/libpng/commit/e79085a1724ccc5526dc752f6a696b3a94ef8cfa

so... I'm going to open up an issue in SPDX regarding how to handle this part.

jlovejoy:
fixes #2591

### DIFF
--- a/src/libpng-1.6.35.xml
+++ b/src/libpng-1.6.35.xml
@@ -1,0 +1,143 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<SPDXLicenseCollection xmlns="http://www.spdx.org/license">
+   <license isOsiApproved="false" licenseId="libpng-1.6.35" name="PNG Reference
+   Library License v1 (for libpng 0.5 through 1.6.35" listVersionAdded="">
+      <crossRefs>
+         <crossRef>http://www.libpng.org/pub/png/src/libpng-LICENSE.txt</crossRef>
+      </crossRefs>
+      <notes>
+         This is similar but has some different text than libpng.
+      </notes>
+      <text>
+         <titleText>
+            PNG Reference Library License version 1
+            (for libpng 0.5 through 1.6.35)<br></br>
+            -----------------------------------------------------------------------
+         </titleText>
+         <p>
+            libpng versions 1.0.7, July 1, 2000, through 1.6.35, July 15, 2018
+            are Copyright (c) 2000-2002, 2004, 2006-2018 Glenn Randers-Pehrson,
+            are derived from libpng-1.0.6, and are distributed according
+            to the same disclaimer and license as libpng-1.0.6 with the
+            following individuals added to the list of Contributing Authors:
+         </p>
+         <p>
+            Simon-Pierre Cadieux<br></br>
+            Eric S. Raymond<br></br>
+            Mans Rullgard<br></br>
+            Cosmin Truta<br></br>
+            Gilles Vollant<br></br>
+            James Yu<br></br>
+            Mandar Sahastrabuddhe<br></br>
+            Google Inc.<br></br>
+            Vadim Barkov
+         </p>
+         <p>
+            and with the following additions to the disclaimer:
+         </p>
+         <p>
+            There is no warranty against interference with your
+            enjoyment of the library or against infringement. There is
+            no warranty that our efforts or the library will fulfill
+            any of your particular purposes or needs. This library is
+            provided with all faults, and the entire risk of satisfactory
+            quality, performance, accuracy, and effort is with the user.
+         </p>
+         <p>
+            Some files in the "contrib" directory and some configure-generated
+            files that are distributed with libpng have other copyright
+            owners, and are released under other open source licenses.
+         </p>
+         <p>
+            libpng versions 0.97, January 1998, through 1.0.6, March
+            20, 2000, are Copyright (c) 1998-2000 Glenn Randers-Pehrson,
+            are derived from libpng-0.96, and are distributed according
+            to the same disclaimer and license as libpng-0.96, with the
+            following individuals added to the list of Contributing Authors:
+         </p>
+         <p>
+            Tom Lane<br></br>
+            Glenn Randers-Pehrson<br></br>
+            Willem van Schaik
+         </p>
+         <p>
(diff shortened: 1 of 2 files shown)


src/libpng-1.6.35.xml
```diff
@@ -0,0 +1,143 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<SPDXLicenseCollection xmlns="http://www.spdx.org/license">
+   <license isOsiApproved="false" licenseId="libpng-1.6.35" name="PNG Reference
+   Library License v1 (for libpng 0.5 through 1.6.35" listVersionAdded="">
```
xsuchy, line 4: Please remove the line break from "name" attribute.
And add value to "listVersionAdded".

jlovejoy: thanks!

jlovejoy:
ugh, @xsuchy or @swinslow - help...

xsuchy:
It seems ok to me. But 
```
java -jar -DLocalFsfFreeJson=true -DlistedLicenseSchema="schema/ListedLicense.xsd" licenseListPublisher-3.1.1.jar LicenseRDFAGenerator 'src/libpng-1.6.35.xml' '/tmp' 1.0 2000-01-01 test/simpleTestForGenerator expected-warnings test/fullTestForGenerator
```
fails. This is what GH actions runs.

But
```
./test-one-license libpng-1.6.35
```
success.

@goneall ideas?

swinslow:
New theory -- I'm noting the following timeline:

* @jlovejoy created this PR on May 10th
* The LicenseListPublisher was updated to a newer version on May 12th -- see https://github.com/spdx/license-list-XML/commit/e4e3897cd58d44e0422f1f2feb553474de2c3c1c
* @jlovejoy made further edits to this PR since then

I see that the new publisher version (3.1.1) is running in the GitHub Actions check. But not sure whether something has gotten out of sync due to the publisher change since then.

I'm going to suggest re-opening a new PR with the same content of these XML and test text files, to see whether it passes then. I'll likely try that myself in a bit just to see if that resolves things.

goneall:
Note that there were some changes made to the crossref code in the license list publisher to improve performance.  From what is posted above, my guess is those changes may be causing the issue.

What is confusing me is I ran a full set of tests against the license list XML source before submitting the changes and the CI for the updated license list publisher should do the same.  Not sure why it didn't fail earlier.

goneall:
I'm able to duplicate the problem on my local machine.  If I get some more time today, I'll try to determine the cause.

In any case, there is definitely an issue with the license list publisher - specifically the license matching code in the SPDX Java Library - since an NPE should never be thrown.

goneall:
I think I found the issue.

The source of the issue is the license XML for the libpng-2.0 license.

The XML has a nested "rule" - an `<optional>` tag with an embedded `<alt>` tag ([here](https://github.com/spdx/license-list-XML/blob/1025da59b12980c2bdc329b74800ce6aeeda188f/src/libpng-2.0.xml#L11)). For the license list publisher, this generates the old template with nested rules which is explicitly disallowed ([reference](https://spdx.github.io/spdx-spec/v2.3/license-matching-guidelines-and-templates/#b153-legacy-text-template-format)).

I don't think this was introduced with the last license list publisher - this is likely showed up now since this license is close to the libpng-2.0 license and hit that piece of matching code.

I didn't see any restrictions on the license XML on embedded `<optional>` and `<var>` tags.  To be honest, I don't recall why we added that restriction - but we should at least look into the history and see if we need to add a similar restriction to the XML.

If we do want to carry forward the restriction, we should update the XML schema to reflect the restriction.  This will also catch this earlier with a better error message when submitting new licenses.

There is also an issue in the license list publisher in that it should be able to handle this situation more gracefully.

swinslow:
Thank you @goneall for taking the time to dig into this!

I note that that optional / alt tag in libpng-2.0 is for a line of hyphens. Now that we've changed the matching guidelines in recent years to make standalone hyphen lines like this ignoreable, we could probably just take that optional / alt tags out entirely.

I'll take a closer look again and then I'll submit a PR to remove those tags. Will also skim the rest of the license to see if any similar issues jump out.

goneall:
@swinslow - You'll probably need to combine the updated libpng-2.0 XML with this one since the license list publisher checks for duplicates against the published license list unless the license XML is included in the same PR.

swinslow:
Adding the extra "expected warning" for an LGPL variant in a789e74 appears to have fixed things. While this current PR doesn't touch on LGPL, it looks like the update to the publisher (or some other adjacent activity) may have triggered this error, as well as related failing builds in the main branch. I'm going to go ahead and merge this which will hopefully get us back on track here. Thanks all!

PR #2747, opened 2025-05-20
GPL-2.0 licenses: "Lesser|Library"

snazy:
The LGPL license was originally called the "GNU LIbrary General Public License" and some license tests refer to the "original" term "Library" instead of "Lesser".

This change is to let license matchers yield a positive match for the usage of "Library" for the "GPL-2.0*" licenses. This change does not add an `<alt>` to every occurence of "Lesser" in all licenses. It's rather there to support a special case.

See also [JDK-8355388](https://bugs.openjdk.org/browse/JDK-8355388).

### DIFF
--- a/src/GPL-2.0-or-later.xml
+++ b/src/GPL-2.0-or-later.xml
@@ -11,7 +11,7 @@
       choice to use code under GPL-2.0-or-later (i.e., GPL-2.0 or some later version),
       as distinguished from the use of code under GPL-2.0-only. The license notice (as seen
       in the Standard License Header field below) states which of these applies to
-      the code in the file. The example in the How to Apply These Terms appendix of the 
+      the code in the file. The example in the How to Apply These Terms appendix of the
       license shows the "or later" approach.
     </notes>
     <text>
@@ -40,7 +40,8 @@
       make sure the software is free for all its users. This General Public
       License applies to most of the Free Software Foundation's software
       and to any other program whose authors commit to using it. (Some other
-      Free Software Foundation software is covered by the GNU Lesser General
+      Free Software Foundation software is covered by the GNU
+      <alt match="Lesser|Library" name="lesserOrLibrary"/> General
       Public License instead.) You can apply it to your programs, too.
     </p>
     <p>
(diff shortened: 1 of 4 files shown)


snazy:
Reference also on [wikipedia](https://en.wikipedia.org/wiki/GNU_Lesser_General_Public_License#History)

src/GPL-2.0.xml
```diff
@@ -465,8 +466,9 @@
         This General Public License does not permit incorporating your program into
         proprietary programs.  If your program is a subroutine library, you may
         consider it more useful to permit linking proprietary applications with the
-        library.  If this is what you want to do, use the GNU Lesser General
-        Public License instead of this License.
+        library.  If this is what you want to do, use the GNU
+        <alt match="Lesser|Library" name="lesserOrLibrary">Lesser</a> General Public License
```
snazy, line 470:
```suggestion
        <alt match="Lesser|Library" name="lesserOrLibrary">Lesser</alt> General Public License
```

src/GPL-2.0.xml
```diff
@@ -57,7 +57,8 @@
       make sure the software is free for all its users. This General Public
       License applies to most of the Free Software Foundation's software
       and to any other program whose authors commit to using it. (Some other
-      Free Software Foundation software is covered by the GNU Lesser General
+      Free Software Foundation software is covered by the GNU
+      <alt match="Lesser|Library" name="lesserOrLibrary">Lesser</a> General
```
snazy, line 61:
```suggestion
      <alt match="Lesser|Library" name="lesserOrLibrary">Lesser</alt> General
```

src/GPL-2.0-or-later.xml
```diff
@@ -447,8 +449,9 @@
         This General Public License does not permit incorporating your program into
         proprietary programs.  If your program is a subroutine library, you may
         consider it more useful to permit linking proprietary applications with the
-        library.  If this is what you want to do, use the GNU Lesser General
-        Public License instead of this License.
+        library.  If this is what you want to do, use the GNU
+        <alt match="Lesser|Library" name="lesserOrLibrary">Lesser</a> General Public License
```
snazy:
```suggestion
        <alt match="Lesser|Library" name="lesserOrLibrary">Lesser</alt> General Public License
```

PR #2752, opened 2025-05-29
libpng-1.6.35: fix missing paren in name

swinslow:
Just fixing a missing right parenthesis in the license name field.

### DIFF
--- a/src/libpng-1.6.35.xml
+++ b/src/libpng-1.6.35.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <SPDXLicenseCollection xmlns="http://www.spdx.org/license">
-   <license isOsiApproved="false" licenseId="libpng-1.6.35" name="PNG Reference Library License v1 (for libpng 0.5 through 1.6.35" listVersionAdded="3.27.0">
+   <license isOsiApproved="false" licenseId="libpng-1.6.35" name="PNG Reference Library License v1 (for libpng 0.5 through 1.6.35)" listVersionAdded="3.27.0">
       <crossRefs>
          <crossRef>http://www.libpng.org/pub/png/src/libpng-LICENSE.txt</crossRef>
       </crossRefs>


xsuchy (approved):
+1

swinslow:
Noting for the record that this appears to be the same issue from https://github.com/spdx/license-list-XML/pull/2729. Though I think that the specific problem there (with libpng-2.0) was resolved in that thread, in order to get that PR to a working state to merge.

zvr:
@swinslow that is... tricky. I can confirm that it has nothing to do with the parenthesis you're adding.

I **think** the issue is, after the license generation, when the license list is being generated, a check for the title is happening...

After a _lot_ of testing, I have a minimal example that also fails:
```xml
   <license licenseId="myid" name="My License">
      <text>
         <titleText>
            PNG Reference Library License version 789
         </titleText>
         <p>
		 Some text.
         </p>
      </text>
   </license>
</SPDXLicenseCollection>
```

If you delete or change _any_ word in the `titleText` (besides changing the version number), no error is produced!!!

For example:

| `titleText` | Result |
|-------------|-----------|
| `PNG Reference Library License version 4` | FAIL |
| `PNG Reference Library License version 789` | FAIL |
| `Reference Library License version 789` | No problem |
| `PNG Reference Library License` | No problem |

I am completely baffled.

For future reference, among the false paths I went down to:
- `listVersionAdded` is too new (nope)
- the `crossRef` points to pure text (no matter)
- `titleText` should have text inside a `<p>` (some do; some don't)

I would need to check the LicenseListPublisher code (or the SPDX Java lib code, for comparison utilities) to dig deeper into what is happening -- but I'm not a Java programmer.

@goneall Does the above point to anything?

swinslow:
Thank you @zvr for digging into this!

An additional thing I'm noting, though not sure what to do with it yet:

Looking at [the checks](https://github.com/spdx/license-list-XML/pull/2752/checks), I see that the 'Validate canonical match' task failed for the "push" event for this PR; but it _passed_ for the "pull_request" event.

From the [YAML file for this GitHub Action](https://github.com/spdx/license-list-XML/blob/main/.github/workflows/validate.yaml), it looks like the two are handled differently in how they set BASE_REF, and then how they fetch the diff of what's changed.

I haven't sorted through the implications of this yet, but just putting it here so I don't forget as I'm continuing to look into this...

khanbilal732 (approved):

PR #2772, opened 2025-06-28
Change to JSON-formatted expected warnings

swinslow:
This is a draft PR for changing to a JSON-formatted version of the `expected-warnings` file. This corresponds to the related PR at https://github.com/spdx/LicenseListPublisher/pull/216.

~~This should _not_ be merged until after the 3.27.0 release is completed (which should be within the next 1-2 days, for real this time...)~~ 3.27.0 is now released, so let's see how we feel about moving this one forward!

The new `expected-warnings.json` file is formatted as an array of arrays of strings. Each array of strings is a list of license IDs that should together be treated as ignored warnings, if they are interpreted as being duplicates during the parsing and license list publishing process. (The specifics of how the publisher can work with this are described in the PR at https://github.com/spdx/LicenseListPublisher/pull/216.)

This should be much easier to maintain than the existing list of "Duplicates licenses:" strings. Open to thoughts on the format or any other considerations / concerns.


xsuchy (approved):
OMG, this is much better. Thank you.

swinslow:
Thanks @xsuchy! 

@goneall, when you have a minute to take a look at this as well as the related changes to the publisher at https://github.com/spdx/LicenseListPublisher/pull/216, I'd welcome your thoughts in particular.

expected-warnings.json
```diff
@@ -0,0 +1,19 @@
+[
```
goneall, line 1: Since the functionality of the "ignore warnings" is to ignore all warnings, not just duplicate licenses (even if that is the only warnings being generated currently), suggest we change this from a JSON array to a JSON object with 2 properties:

- `duplicates`
- `additionalWarnings`

both would be arrays.

e.g.:

```
{
   "duplicates" : [ ["AGPL ...], ...],
   "additionalWarnings": []
}
```

goneall (changes requested):
Suggested change to support possible future warnings in the license list publisher.

PR #2777, opened 2025-07-04
Replace LLVM Exception link with Wayback archive

tuliom:
The old link is now broken. Replace it with the last archive available from the Wayback Machine.
Notice there is also a redirection from http://llvm.org/foundation/ to https://foundation.llvm.org/.

### DIFF
--- a/src/exceptions/LLVM-exception.xml
+++ b/src/exceptions/LLVM-exception.xml
@@ -3,6 +3,7 @@
    <exception licenseId="LLVM-exception" name="LLVM Exception">
       <crossRefs>
          <crossRef>http://llvm.org/foundation/relicensing/LICENSE.txt</crossRef>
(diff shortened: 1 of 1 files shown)


goneall:
If we could add the wayback link and leave the original (dead) link it would support matching algorithms that use URL references.  Note that the published license list will indicate that the original link is dead.

goneall:
> I've just realized the Wayback Machine is also able to redirect the old link to the new one via https://web.archive.org/web/20240423023851/https://llvm.org/foundation/relicensing/LICENSE.txt
> 
> Which solution is preferred?
> 
> 1. Add an additional link pointing to https://web.archive.org/web/20240423023852/https://foundation.llvm.org/relicensing/LICENSE.txt and keep the old (dead) link.
> 2. Replace the previous link with https://web.archive.org/web/20240423023851/https://llvm.org/foundation/relicensing/LICENSE.txt

Prefer option 1

PR #2779, opened 2025-07-08
move content of test/simpleTestForGenerator to test/fullTestForGenerator

xsuchy:
Today, almost no one is using the directory  `fullTestForGenerator/(license-id)/license/good/` for tests and storing more variants of licenses. Including me, because I take some existing license as an example. This new directory has been supported by the publisher for some time. 
I suggest that we move all files from `test/simpleTestForGenerator/${@}.txt` to `test/fullTestForGenerator/${@}/license/good/${@}.txt`.
This is just a draft that moves MIT license and change `test-one-license` script (it will now pass for MIT but fail for anything else).

Let me know what do you think. If you would like it, I will draft a PR that moves all tests. But the final PR will be huge and hard to review. That is why I am making this draft first.


goneall:
@jlovejoy - Just saw it now

@xsuchy - I would prefer to keep the simpleTestForGenerator directory since I would need to change some of my release workflows for the LicenseListPublisher.  I agree it would be cleaner, but not enough benefit for the costs IMHO.

khanbilal732 (approved):

PR #2795, opened 2025-07-24
GNU Compiler Exception wording

### LINKED ISSUE #2785: Alternate wording in GNU Compiler Exception
The exception [GNU Compiler Exception](https://github.com/spdx/license-list-XML/blob/main/src/exceptions/GNU-compiler-exception.xml) uses the phrase "files compiled with a GNU compiler".

Some very old files (inside the Linux kernel) use the whole exception wording but with the change of that phrase to "files compiled with GCC".

I propose that alternate wording be introduced in the exception definition, so that both cases are covered by the same identifier.

zvr:
Adds alternative wording to GNU Compiler Exception, to handle both "GCC" and "a GNU compiler"

Closes #2785


xsuchy:
+1

xsuchy (approved):

Issue #2689, opened 2025-03-26
Request for New SPDX License Identifier or Modification of ANTLR-PD for PCCTS


nishanthsankaran:
I have a license for the Purdue Compiler Construction Tool Set (PCCTS) available at [PCCTS License](https://github.com/thaddeus002/pccts/blob/master/LICENSE). This license matches the SPDX license identifier [ANTLR-PD-fallback](https://spdx.org/licenses/ANTLR-PD-fallback.html) with the following differences:

1. **Fallback Copyright License**: The ANTLR license includes a fallback copyright license for countries where public domain status is not valid, while the PCCTS license does not mention this. (Could be blue omittable)
   > In countries where the Public Domain status of the work may not be valid, the author grants a copyright license to the general public to deal in the work without restriction and permission to sublicense derivatives under the terms of any (OSI approved) Open Source license.
2. **Name Replacement**: "ANTLR" is replaced with "Purdue Compiler Construction Tool Set (PCCTS)" or "PCCTS". (Could be in red so that it can be also used for PCCTS)
3. **Developer Information**: The PCCTS license includes specific mention of:
   > ANTLR 1.33
   > Terence Parr
   > Parr Research Corporation
   > with Purdue University and AHPCRC, University of Minnesota
   > 1989-1995

   (Could be blue as well)

Should these differences warrant creating a new SPDX license identifier, or should the existing ANTLR-PD-fallback license be modified with red and blue text to accommodate the PCCTS license?

karsten-klein:
**{metæffekt} Universe**
canonical name: [Public Domain](https://github.com/org-metaeffekt/metaeffekt-universe/blob/main/src/main/resources/ae-universe/[p]/README.md#user-content-Public-Domain:~:text=Public%20Domain)
short name: Public-Domain
category: Public Domain
ScanCode reference id: [public-domain](https://github.com/nexB/scancode-toolkit/blob/develop/src/licensedcode/data/licenses/public-domain.LICENSE)
OSI status: _none_

**ScanCode**
matched id: [antlr-pd](https://github.com/nexB/scancode-toolkit/blob/develop/src/licensedcode/data/licenses/antlr-pd.LICENSE)

**Comment**
Do you regard this a license text? For me this appears as a public domain dedication. 
The part

> An individual or
> company may do whatever they wish with source code distributed with
> PCCTS or the code generated by PCCTS, including the incorporation of
> PCCTS, or its output, into commerical software.

I would consider explanatory only.

[...]

PR #2801, opened 2025-08-02
Fix ANTLR-PD: Add alt/optional tags for PCCTS and text variants (#2689)

nishanthsankaran:
This PR addresses issue #2689 by adding `<alt>` and `<optional>` markup to the ANTLR-PD license to support variations found in the [PCCTS Software Rights text](https://github.com/thaddeus002/pccts/blob/master/LICENSE).

### Summary of changes:
- Used `<alt>` to match "ANTLR" and "PCCTS"
- Used `<alt>` to handle "legal" vs "LEGAL"
- Used `<alt>` for "header" vs "headers", and "this" vs "the"
- Used `<optional>` for clauses like "use ANTLR or", "fully", etc.

Validated that the file remains well-formed XML and is consistent with SPDX markup guidelines.

Let me know if further edits are needed.


swinslow (approved):
Looks good to me. If in the future there are additional projects using these variants, then the `ANTLR|PCCTS` regexes can be updated to `.*` or similar, but this is fine for now.

PR #2808, opened 2025-08-05
 BSD-3-Clause.xml - add markup to third clause

### LINKED ISSUE #2807: BSD-3-Clause variant (yet another)
This license has a slight different wording for the third clause, using "any contributors" instead of "its" or "their" contributors.

Has some declarations on top for US gov't, but I'd would not consider that part of the license text

jlovejoy:
fixes #2807


bsdimp (approved):
Seems reasonable to me. There's no other antecedent that 'their' could refer to.
But IANAL

PR #2820, opened 2025-08-19
Add ESA-PL-strong-copyleft-2.4

LuisMartinPoloESA:


lase3511 (changes requested):
Gffyuiiu

xsuchy:
This is wrong conentent. You pushed permisive license instead of the strong-copyleft. I think you confused branch names.

PR #2847, opened 2025-09-26
test-one-license: modernize bash script

quantenzitrone:
this makes the script runnable on nixos and other distros that don't install bash in /bin

### DIFF
--- a/test-one-license
+++ b/test-one-license
@@ -1,9 +1,14 @@
-#!/bin/bash
+#!/usr/bin/env bash
+#
+# Usage:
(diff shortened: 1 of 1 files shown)


test-one-license
```diff
@@ -1,9 +1,14 @@
-#!/bin/bash
+#!/usr/bin/env bash
```
xsuchy, line 1: +1 to /usr/bin
but -1 to env because of https://jmmv.dev/2016/09/env-considered-harmful.html
Can you just put there /usr/bin/bash ?

PR #2848, opened 2025-09-26
MIT: add markup for additional "on"

quantenzitrone:
found this license version in https://gitlab.freedesktop.org/xorg/xserver/-/blob/dd5c2595a42d3ff0c4f18d9b53d1f6c3fd934fd4/COPYING#L365-389

since the change is so minimal i thought it would be easier to just make a pr instead of an issue first

### DIFF
--- a/src/MIT.xml
+++ b/src/MIT.xml
@@ -14,32 +14,32 @@
          </p>
       </copyrightText>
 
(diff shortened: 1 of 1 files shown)


xsuchy:
+1 
but I propose to add the URL of this case to crossref at the header of XML.

xsuchy (approved):
+1

PR #2892, opened 2025-11-26
Update LiLiQ-P-1.1.xml - add URL

### LINKED ISSUE #2624: Update URL for LiLiQ-P-1.1
The description page for [LiLiQ-P-1.1 ](https://spdx.org/licenses/LiLiQ-P-1.1.html) shows that the official website is no longer live -- it actually is live, they just [changed the URL to remove the version number](https://forge.gouv.qc.ca/licence/liliq-p/). I'm not sure if this shows that they intend to no longer make changes, as the page itself also doesn't mention version 1.1 but is the same license text as 1.1.

I would do a pull request but I'm not confident I understand the processes here well enough :)

jlovejoy:
fixes #2624

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 # SPDX-License-Identifier: CC0-1.0
 
-TOOL_VERSION = 3.1.3
+TOOL_VERSION = 3.1.4
 TEST_DATA = test/simpleTestForGenerator
 FULL_TEST_DATA = test/fullTestForGenerator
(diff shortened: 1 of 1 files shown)


xsuchy:
The test fails with " WARNING: Duplicates licenses: LiLiQ-P-1.1, LiLiQ-P-1.1" but this change cannot trigger it. What do you think @goneall ?
+1 from me.

PR #2900, opened 2025-12-03
NIST-PD: Fix typo in reference to US Code

lfaraone:
There's no such thing as 15 USC § 105. 

Instead, ref https://www.law.cornell.edu/uscode/text/17/105


xsuchy:
-1 
This is a collection of licenses and their definitions. We are not the ones who define or modify them. If the author has specified "15 USC" in the license, we simply copy and document what the author has used. And this is how it is in the upstream license. The license specified in `crossRef` uses "15 USC". When a license with "17 USC" is used in a project, we will document it as a possible variant.

I recommend contacting the author of the license or the projects that use that license.

PR #2908, opened 2025-12-22
Remove duplicate text in alt node organizationClause4

mikini:
The current non-alt license text [as presented on the SPDX site][1] does not correspond to the original license's bullet 4.

This removes the superfluous and duplicated sentence part added in commit de07000d812e768727af9a1db0a566d7161deb27 (2022-09-14) to again align with the historical license text.

[1]: https://spdx.org/licenses/BSD-4-Clause.html


xsuchy:
I think it does not matter what is inside of `<alt>` as that is actually a placeholder. And the proposed change match what is in https://directory.fsf.org/wiki/License:BSD-4-Clause so I am leaning toward +1 from me.

xsuchy:
Thank you.

Issue #2913, opened 2025-12-28
New license request: CLL-1.0.0 [SPDX-Online-Tools]


Chaosyr:
**1.** License name: Cecil Libraries License (1.0.0)
**2.** Short identifier: CLL-1.0.0
**3.** License author or steward: Cecil Libraries Organization
**4.** Comments: It is a overarching license defining rights of all groups associated with the Product, and any limitations that each group may have. Including over associated Product IP. [Also going to state it in advance here, this is originally formatted in Markdown so there may need to be some structural modification.]
**5.** License request Url: https://tools.spdx.org/app/license_requests/491
**6.** URL(s): https://www.nuget.org/packages/Cecil.ANSI_Utils/1.0.1/License, https://github.com/Cecil-Libraries/Cecil-ANSI-Utils?tab=License-1-ov-file, https://github.com/Cecil-Libraries/Cecil-ANSI-Utils/wiki/Licensing-and-Attribution
**7.** OSI status: Not Submitted
**8.** Example projects: https://www.nuget.org/packages/Cecil.ANSI_Utils, https://github.com/Cecil-Libraries/Cecil-ANSI-Utils, https://github.com/Cecil-Libraries/ANSI-Utils-Tester

Chaosyr:
aight #2915 fixes the XML formatting. Sorry if its a bit complexly layed out trying to follow the original Markdown e.g.

### Cecil Libraries License (1.0.0)
Copyright YEAR ORGANIZATION and the Founder(s) FOUNDERS

In the case where it so falls that a Sub-License and this License conflicts, this License dictates. 

Also, any Licenses applied to this Product outside of this License are deemed as Sub-Licenses unless otherwise stated.

1. Key Terms

    This section likely self-explanatory goes over any Key Terms that may be unfamiliar to those looking at this License and seeing if their Project breaks it.
    1. Founder
        1. The person who Originally made this Product. OR under the condition the Product does not come soley from them Definition 2 applies.
        2. The group behind a given Product.
    2. Organization
        1. The GitHub Team that houses all the Products that this License affects.
    3. Contributors
        1. Anyone who has contributed to the Product in any way shape and or form.
    4. Developer
        1. The person in which is using this Product for purposes of developing their own projects.
    5. Mod-Maker
        1. A Developer who is using this Product for purposes of making mods.
    6. Game Developer
        1. A Developer who is using this Product for purposes of developing their own games.
    7. End-User
[...]

PR #2915, opened 2025-12-29
Fixes the CLL XML

Chaosyr:
* Reformats the CLL XML File to be properly valid, in hopes that it displays correctly.


swinslow:
#2913 is closed as not accepted, so I am closing this PR.

PR #2935, opened 2026-01-29
add new crossRef URL for LGPL-2.1 and LGPL-2.1-or-later

spiffcs:
👋 Over in the https://github.com/anchore/syft project we noticed we were cataloging some packages that would surface this URL as a reference for LGPL-2.1. We match on the URL to to send us to the correct SPDX ID in some cases where the ID is not available.

I noticed the discussion on https://github.com/spdx/license-list-XML/pull/2678 and thought this could be in the same vein as that change:

> (1) we'll accept contributed PRs with accurate "other URLs" for licenses including where e.g. OSI has changed their website URLs; but (2) we -- the SPDX legal team participants -- won't be going out seeking to keep them up to date on our own.

We had some users try and patch this into our consumption of the SPDX license list (generated code), but I thought it would be beneficial to update the list so other consumers could get the update who might use URL -> ID lookups.

Happy to make any changes needed here 😄

### DIFF
--- a/src/LGPL-2.1-or-later.xml
+++ b/src/LGPL-2.1-or-later.xml
@@ -4,6 +4,7 @@
   name="GNU Lesser General Public License v2.1 or later">
     <crossRefs>
       <crossRef>https://www.gnu.org/licenses/old-licenses/lgpl-2.1-standalone.html</crossRef>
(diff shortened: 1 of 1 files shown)


spiffcs:
Thanks @swinslow for the quick response - I'll also add it to the page for `LGPL-2.1-or-later`

Issue #2956, opened 2026-03-02
New version for ESA licenses


LuisMartinPoloESA:
Dear SPDX Team, 
I am opening this request to introduce a new version of the ESA licenses: 2.5.
Thank you in advance,
Luis

xsuchy:
Closing in favour of #2986

PR #2957, opened 2026-03-02
#2956 ESA licenses version 2.5

LuisMartinPoloESA:
Replaced version 2.4 by version 2.5 of ESA licenses.


xsuchy:
We do not replace old version (see the case of GPL 1, 2, 3).  Please create issue first that you want to add 2.5 and note that 2.4 already exist.

PR #2961, opened 2026-03-05
Add Brian Gladman 3-Clause License (no conversion)

### LINKED ISSUE #2684: Request to Add Brian-Gladman-3-Clause License with Omittable Clause
Please add the following license text to the Brian-Gladman-3-Clause license and mark it in blue as omittable:

> ALTERNATIVELY, provided that this notice is retained in full, this product may be distributed under the terms of the GNU General Public License (GPL), in which case the provisions of the GPL apply INSTEAD OF those given above.

We have a variation of the Brian-Gladman-3-Clause license without the above text present in the [aesopt.h file](https://github.com/krb5/krb5/blob/kfw-4.0.1-beta1/src/lib/crypto/builtin/aes/aesopt.h).

Requesting to mark the above text as omittable, so that the license text can be used as Brian-Gladman-3-Clause.

jlovejoy:
fixes #2684


xsuchy:
Oh, This is conflicting with 1ff5448a90588c9b173ca45f388125937812aa36 where you added just the test. Can you please rebase and resolve the conflict?

PR #2984, opened 2026-04-23
Fix notes for X11 License description

pmetzger:
Currently, the X11 license says that it is the same as MIT but without an advertising clause, but in fact, it is the same as MIT but with an advertising clause.


swinslow:
Thanks for flagging this! Discussed on legal team call 2026-04-23; made a slight change to clarify, and will merge after CI tests run.

PR #2987, opened 2026-04-23
Add Brian-Gladman-3-Clause-no-conversion.xml

### LINKED ISSUE #2684: Request to Add Brian-Gladman-3-Clause License with Omittable Clause

jlovejoy:
fixes https://github.com/spdx/license-list-XML/issues/2684


xsuchy (approved):
+1

PR #2960, opened 2026-03-04
Update all opensource.org/license links to use SPDX

nickvidal:
As per #2959, the OSI is now using SPDX identifiers for license URLs.

PR #2988, opened 2026-04-25
SimPL-2.0: Fix OSI URL and mark last section as optional

waldyrious:
The last section of SimPL-2.0 text is not strictly part of the license itself, so this PR marks it as optional, to avoid conveying misleading guidance to users.

Additionally, the OSI URL (modified in PR #2960) has incorrect capitalization: it is written as "SiMPL" instead of "SimPL". Although that URL does work, resolving to the correct page, the capitalization does not match the actual license's short name, and can be potentially misleading, so this PR also corrects the URL.

While modifying the file, I took the opportunity to fix some indentation inconsistencies in this file.

Finally, this PR corrects a punctuation inconsistency in the list of disclaimers: the last item in that list currently ends with a semicolon, which is a leftover from an [earlier draft](https://lists.opensource.org/pipermail/license-discuss_lists.opensource.org/2007-March/012630.html) in which an additional list item followed it. When that extra item was removed, the semicolon in the item before it should have been adjusted to a period, but that detail appears to have been missed at the time. This PR fixes the punctuation to ensure the consistency of the recorded text.

Each of the changes above are made in separate, atomic commits, so that they can be reviewed independently (I would suggest viewing the [diff in whitespace-insensitive mode](https://github.com/spdx/license-list-XML/pull/2988/changes?w=1)), and dropped if undesirable (or split to separate PRs if that's preferred).

### DIFF
--- a/src/SimPL-2.0.xml
+++ b/src/SimPL-2.0.xml
@@ -1,22 +1,22 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <SPDXLicenseCollection xmlns="http://www.spdx.org/license">
-   <license isOsiApproved="true" licenseId="SimPL-2.0" name="Simple Public License 2.0">
-      <crossRefs>
-         <crossRef>https://opensource.org/license/SiMPL-2.0</crossRef>
-      </crossRefs>
+  <license isOsiApproved="true" licenseId="SimPL-2.0" name="Simple Public License 2.0">
+    <crossRefs>
+      <crossRef>https://opensource.org/license/SiMPL-2.0</crossRef>
+    </crossRefs>
     <text>
       <titleText>
-         <p>Simple Public License (SimPL)</p>
+        <p>Simple Public License (SimPL)</p>
       </titleText>
       <optional>
-         <p>Preamble</p>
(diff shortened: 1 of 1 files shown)


waldyrious:
I should also point out that the page at https://spdx.org/licenses/SimPL-2.0.html still shows the old URL from before #2960 (https://opensource.org/licenses/SimPL-2.0), and includes the annotation "[no longer live]", despite that URL being functional (if outdated).

The first issue I assume is likely to be due to the recency of the change from #2960, and the second one should become moot once the URL change is deployed; nevertheless, I thought I'd point them out, just in case there's something missing or incorrect in the site publication and/or URL validation routines.

xsuchy:
LGTM for first three commits.
I am not sure about the last commit changing the semicolon to a full stop.
Conditional +1 from me.

waldyrious:
Thanks for the review, @xsuchy. Can you help me understand what process would be acceptable for updating that final semicolon to a period? Other than releasing a new version of the license I mean 😅

p.s.: I've rebased the branch.

xsuchy:
No special process is needed. I will just leave the final review to @swinslow or @jlovejoy

PR #3005, opened 2026-05-18
Updated the url of the ImageMagick license.

dlemstra:


swinslow:
Hi @dlemstra, thanks for this. Good with the general concept, but I think we'll want to make a couple of tweaks to align with SPDX practices for updates to existing URLs. I've started an issue at #3006 to discuss.

xsuchy:
As described in https://github.com/spdx/license-list-XML/issues/3006#issuecomment-4566089584 this cannot be merged in this form.
It can be merged if you add the new url as second crossRef instead of modyfing the first.
And the second add as `<alt match=....>`

dlemstra:
Thanks for the reminder, I have updated the pull request and made the requested changes.

src/ImageMagick.xml
```diff
@@ -2,6 +2,7 @@
 <SPDXLicenseCollection xmlns="http://www.spdx.org/license">
    <license isOsiApproved="false" licenseId="ImageMagick" name="ImageMagick License" listVersionAdded="1.20">
       <crossRefs>
+         <crossRef>https://imagemagick.org/license/</crossRef>
```
xsuchy, line 5: Hmm, This is interesting. The license on this URL actually differs from the canonical text in our db.

* DVD vs. digital media
*  in point 3 is additional "(except as stated in this section)"

I think we can add this as '<alt>' markup. But this will require review from more people as this is no longer pure technical change.

PR #3040, opened 2026-07-17
Update name for Unicode-TOU

### LINKED ISSUE #1852: Consider changes to Unicode-TOU
This is following from the discussion on the spdx-legal mailing list at https://lists.spdx.org/g/Spdx-legal/topic/97116566

[`Unicode-TOU`](https://spdx.org/licenses/Unicode-TOU.html) is not a versioned identifier. It points to the license text for the Unicode Terms of Use as they existed as of (I believe) 2014. But the terms of use posted on the Unicode website at https://www.unicode.org/copyright.html have evolved over time, and the mailing list thread discusses how this has caused some confusion with folks using `Unicode-TOU` to refer to the wrong set of terms.

A separate issue #1851 discusses whether to add a new, separate identifier with the current updated terms.

For the current `Unicode-TOU` identifier, we should consider either or both of the following options:
1. For the license's `Full name`, modify it to be `Unicode Terms of Use - 2014` or something similar, to help emphasize that it is the 2014 version (again, assuming that's the correct year - should be verified); and/or
2. Deprecate `Unicode-TOU` as an identifier, and add a new ID `Unicode-TOU-2014` with the same existing content.

I'm in favor of 1, and on the fence for 2. Interested in other community members' thoughts.

For clarity, in neither case would we remove `Unicode-TOU` as a valid identifier. It would remain valid (even if deprecated) and pointing to the existing license text that's currently associated with it.

jlovejoy:
fixes #1852

@swinslow - see issue... I think this should be enough to close that one?

### DIFF
--- a/src/Unicode-TOU.xml
+++ b/src/Unicode-TOU.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <SPDXLicenseCollection xmlns="http://www.spdx.org/license">
-   <license isOsiApproved="false" licenseId="Unicode-TOU" name="Unicode Terms of Use" listVersionAdded="1.20">
+   <license isOsiApproved="false" licenseId="Unicode-TOU" name="Unicode Terms of Use - 2014 version" listVersionAdded="1.20">
       <crossRefs>
          <crossRef>http://web.archive.org/web/20140704074106/http://www.unicode.org/copyright.html</crossRef>
          <crossRef>http://www.unicode.org/copyright.html</crossRef>


swinslow:
@jlovejoy I'm not merging this right now, because looking at the thread in #1852 I think we also were discussing whether to simultaneously add a note explaining why newer versions of the TOU were not added to the list. I'll move this to 3.30.0 and we can confirm + get it included there.

saitanallensantiago26-beep:
> fixes #1852
> 
> Signed-off-by: Jilayne Lovejoy
> 
> @swinslow - see issue... I think this should be enough to close that one?

PR #3041, opened 2026-07-17
Update faq.md - add question about SMLNJ

### LINKED ISSUE #2758: SMLNJ is literally just HPND
like look at this diff:
Image

> <copyright holder> makes no representations about the suitability of this software for any purpose. It is provided "as is" without express or implied warranty.

is omittable in HPND

the only difference at this point is

> and warranty disclaimer

if these 3 words are really worth an entire new license, i think it should be called HPND-warrany-disclaimer or something

jlovejoy:
fixes #2758 

### DIFF
--- a/DOCS/faq.md
+++ b/DOCS/faq.md
@@ -114,3 +114,10 @@ expression: `Apache-2.0 AND MIT AND GPL-2.0-or-later`
  
 ## How are license identifiers used in an SPDX Document?
 The SPDX specification includes fields to record license information at the [package](https://spdx.github.io/spdx-spec/v2.3/package-information/), [file](https://spdx.github.io/spdx-spec/v2.3/file-information/), and [snippet](https://spdx.github.io/spdx-spec/v2.3/snippet-information/) level. 
+
+## Questions about specific licenses
+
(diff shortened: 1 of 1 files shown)


swinslow:
@jlovejoy I'm generally fine with the content, but do we want this to be in the FAQ document? I feel like if we open the door to having the FAQ contain notes about specific licenses, then it's going to be unclear what should go there vs. in the notes part of the license list entry. Plus now it would require people to look in both the FAQ and the license list entry for various licenses.

Do we maybe just want to incorporate this text, or something similar, into the notes entry for the SMLNJ license?

mxmehl:
> Do we maybe just want to incorporate this text, or something similar, into the notes entry for the SMLNJ license?

FWIW, as a consumer of SPDX I agree to @swinslow's take. Please put such information as close to the licenses as possible.

Regarding the content and the decision not to alter the IDs, the proposition LGTM.

xsuchy:
> then it's going to be unclear what should go there

It is an FAQ. So **frequently asked quesitons** should go there. The frequency is the only metric here.

PR #3079, opened 2026-09-02
Update new-license-workflow.md

jlovejoy:
We had meant to remove this file sometime ago. I've starting deleting stuff that is outdated or elsewhere, but left some bits that are not otherwise documented, and/or if we want to "re-purpose" this file/filename otherwise

@swinslow - FYI


DOCS/new-license-workflow.md
```diff
@@ -1,148 +1,4 @@
...
-                </list>
-            </item>
-        </list>
-    </item>
-</list>
-```
-6. Save the file and [commit it](https://docs.github.com/en/github/committing-changes-to-your-project) to your clone of the repository.
-
-You're done! Yes, there may be more advanced cases where more work than this is required. If you don't add licenses often, it's unlikely the core team will have assigned you something like that, but if they did or you have any questions at all, please ask! They love to help!
+This file is outdated or in process of getting updated or moved!
```
xsuchy, line 1: Can we remove it only when we have the new doc available?

